PR #4, opened 2020-09-04
Kafka prefix added

psav:


kylape:
1. The prefix should be set on the CR by the user.  In the ephemeral case, it is the responsibility of the bonfire script to set the random string
2. If we were to set it ourselves, I think it should go in the CR status, not labels.

PR #33, opened 2020-09-11
Added annotate config

psav:


psav:
This does not do as intended - an annotation change does not cause a pod bounce

PR #42, opened 2020-09-16
Test refactor

jhjaggars:
The goal of this change is to refactor makers to be more testable.  Starting out with separating the object construction and manipulation from the network IO bits.


kylape:
Rebased locally and worked through all the conflicts.  Also pushed to master locally.

PR #49, opened 2020-09-22
Resource rename

psav:


j00bar:
Those are quite generic names for the Resource types - we really don't want to have any additional clarity about them?

In other popular operators, I'm seeing "VaultService" or "EtcdCluster" or "PgCluster" or such.

PR #50, opened 2020-09-23
Make Environment cluster scope

psav:


kylape:
Only a couple of things:

1. I kind of prefer the attribute name `TargetNamespace` as opposed to `Namespace`
2. We need to add some logic to check for the presence of the target namespace.  What if the requester shouldn't have access to that namespace?  If the namespace doesn't exist, should we create it? Or maybe we shouldn't allow a user to specify a target namespace that _does_ already exist??  This makes me wonder if we should actually generate the namespace name used as the target for env-wide resources

kylape:
That said, we could just merge this and address the target namespace stuff in a follow up PR

PR #54, opened 2020-09-25
Service dependency changed to build based on config

psav:
Service dependency changed to build based on config

* Previously the service dependency was set up to query k8s for the apps
  and their services in order to build the dependency list. This has now
  been modified so that we infer the names and port numbers from known
  configuration.
* Tests have been written to fulfil Single, Missing and Multiple
  dependencies.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -19,4 +19,4 @@ jobs:
           tar xzvf kubebuilder/kubebuilder_2.3.1_linux_amd64.tar.gz -C kubebuilder
           make
       - name: Run Tests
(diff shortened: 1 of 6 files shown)


kylape:
Actually I think that the code doesn't work both before or after your changes.  If we are adding multiple pods into one app, we need to fetch the app and examine each pod to know the hostname and port.

psav:
This is now passing well, I rejigged the way call of the functions to be able to write some better testing - requires no k8s plane to be running - so more unit testy. The downside is, we may have to rethink a little how we name/write service dependencies - Not sure how you want to move forward on this, but with multiple pods present under a `ClowdApp`, how do we target an inner pod service? We could use a dotted notation of `app_name.pod_name` and, at the moment I have assumed that service names are unique across ALL `ClowdApps` in a `ClowdEnvironment`. I am happy to change this before merging this PR.

PR #61, opened 2020-10-01
Refactor of secret handling

psav:


kylape:
Alright, I'm going to try and write down my thought process on this PR.

GetData is only called on the result of dataInit, and Populate is only called at the very end of the MakeOrGetSecret function.  In addition, the SecretMarshal parameter to MakeOrGetSecret is only used at the very end (to call Populate).  I don't see there being much reuse in the interface beyond these single calls to each function.  Thus here is my conterproposal:

```
func MakeOrGetSecret(ctx context.Content, env *crd.ClowdEnvironment, client client.Client, nn types.NamespacedName, dataInit InitFn)) (map[string]string, error)
```

The function only deals with string maps, and the calling function can handle populating the config object.  We can drop the GetData function from the interface since it's only ever used by the DataInit functions; those can just return map[string]string.  The DataInit functions should be moved to their respective provider since they could differ based on the provider (e.g. minio vs app-interface).

controllers/cloud.redhat.com/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (ojs *ObjectStoreConfig) Populate(data *map[string]string) {
+	port, _ := strconv.Atoi((*data)["port"])
+	ojs.Hostname = (*data)["hostname"]
+	ojs.AccessKey = (*data)["accessKey"]
+	ojs.SecretKey = (*data)["secretKey"]
+	ojs.Port = port
+}
+
+func MakeOrGetSecret(obj ConfigPopulater, env *crd.ClowdEnvironment, client client.Client, ctx context.Context, nn types.NamespacedName, dataInit func() map[string]string) error {
```
kylape, line 38: I'd still prefer for the method to return a `map[string]string` instead of take a `ConfigPopulater`.  Also - minor point - but I think it's best practice to make `ctx` the first argument of a function.

kylape:
Looks good besides my one comment

PR #65, opened 2020-10-05
Added resource defaults

psav:


controllers/cloud.redhat.com/makers/maker.go
```diff
@@ -306,17 +316,24 @@ func (m *Maker) makeDeployment(pod crd.PodSpec, hash string) error {
 
 	}
 
+	var resources core.ResourceRequirements
+	if pod.Resources.Limits == nil && pod.Resources.Requests == nil {
```
kylape, line 320: I wonder if we should check for each individual request and limit, and if one is not specified, we pull the respective default from the environment.

controllers/cloud.redhat.com/makers/makers_test.go
```diff
@@ -216,3 +220,173 @@ func TestMultiDependency(t *testing.T) {
...
+	}
+
+	d := &apps.Deployment{}
+
+	nn := types.NamespacedName{
+		Name:      app.Spec.Pods[0].Name,
+		Namespace: app.Namespace,
+	}
+
+	initDeployment(&app, &env, d, nn, app.Spec.Pods[0], "hihi")
```
kylape, line 327: I kind of feel like this function call should be in the calling function since this is the actual function we're testing.  Like, `setupResourcesForTest` just sets up the resources for running the test, where "the test" is `initDeployment`.

kylape (commented):
Overall much more readable!  Once you address my comments (not required to make changes), feel free to merge.

PR #66, opened 2020-10-07
Add app-interface providers for object store and logging

kylape:
I think the logging one was technically already working, but it's not officially split out into its own file.  I added unit tests for the app-interface object store, but I have no integration test yet.

### DIFF
--- a/controllers/cloud.redhat.com/providers/appinterface_objectstore.go
+++ b/controllers/cloud.redhat.com/providers/appinterface_objectstore.go
@@ -104,7 +104,7 @@ func genObjStoreConfig(secrets []core.Secret) (*config.ObjectStoreConfig, error)
 
 	if len(buckets) > 0 && objectStoreConfig.Hostname == "" {
 		err := errors.New("Could not find object store hostname from secrets")
-		return &objectStoreConfig, err
+		return nil, err
 	}
 
 	objectStoreConfig.Buckets = buckets
--- a/controllers/cloud.redhat.com/providers/appinterface_objectstore.go
+++ b/controllers/cloud.redhat.com/providers/appinterface_objectstore.go
@@ -1,8 +1,15 @@
 package providers
 
 import (
+	"fmt"
+	"reflect"
+	"strings"
+
 	crd "cloud.redhat.com/clowder/v2/apis/cloud.redhat.com/v1alpha1"
 	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/config"
+	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/errors"
+	core "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type AppInterfaceObjectstoreProvider struct {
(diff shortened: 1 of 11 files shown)


controllers/cloud.redhat.com/providers/appinterface_objectstore.go
```diff
@@ -0,0 +1,112 @@
...
+		endpoint, endpointOk := secret.Data["endpoint"]
+
+		if accessKeyOk && secretKeyOk && nameOk {
+			bucketConfig := config.ObjectStoreBucket{
+				AccessKey: strPtr(string(accessKey)),
+				SecretKey: strPtr(string(secretKey)),
+				Name:      string(name),
+			}
+			if endpointOk {
+				objectStoreConfig.Hostname = string(endpoint)
```
psav, line 99: This is going to update the `objectStoreConfig` for each secret, replacing the previous value. I'm assuming they are all going to be the same, but what if they're not?

controllers/cloud.redhat.com/providers/appinterface_objectstore.go
```diff
@@ -0,0 +1,112 @@
...
+
+	c.Buckets = buckets
+	return nil
+}
+
+func genObjStoreConfig(secrets []core.Secret) (*config.ObjectStoreConfig, error) {
+	buckets := []config.ObjectStoreBucket{}
+	objectStoreConfig := config.ObjectStoreConfig{Port: 443}
+
+	for _, secret := range secrets {
```
psav, line 86: This loops through ALL secrets in a namespace looking for AWS related ones. Is there a naming convention we can leverage here? Just feels kinda weird that we are pulling all of them.

controllers/cloud.redhat.com/providers/appinterface_objectstore.go
```diff
@@ -0,0 +1,112 @@
...
+			if endpointOk {
+				objectStoreConfig.Hostname = string(endpoint)
+			}
+			buckets = append(buckets, bucketConfig)
+		}
+	}
+
+	if len(buckets) > 0 && objectStoreConfig.Hostname == "" {
+		err := errors.New("Could not find object store hostname from secrets")
+		return &objectStoreConfig, err
```
psav, line 107: Doesn't convention dictate we should return `nil` here, instead of a pointer to the object. The calling code immediately interrogates the error and returns anyway. Just making sure I understand why this is different from the other examples across the codebase.

controllers/cloud.redhat.com/providers/appinterface_objectstore_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"testing"
+
+	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/config"
+	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/utils"
+	core "k8s.io/api/core/v1"
+)
+
+type TestSecrets struct {
+	ExactKeys   map[string]string
+	ExtraKeys   map[string]string
```
psav, line 14: These three (`ExtraKeys`, `NoKeys`, `MissingKeys`) are there to be able to create extra tests from?

psav (approved):
Overall, I like the direction of this, I was perturbed by dropping the config object in favour of a `map[string]string`, just because it allows more manipulation of the object to included data that can get lost easily, but I see why, and I'm happy with it.

psav:
This will need to be merged before I `makeComponent` on the minio. When this is merged I'll create and release new versions of the config clients.

controllers/cloud.redhat.com/providers/appinterface_objectstore.go
kylape: Yes, I admit that is a risk.  If they turn out to not be the same, this is a bug, but I don't know how we'd choose which hostname to use.

controllers/cloud.redhat.com/providers/appinterface_objectstore.go
kylape: Can you articulate what specifically feels weird about it?

controllers/cloud.redhat.com/providers/appinterface_objectstore.go
kylape: Yeah we can return `nil`.

controllers/cloud.redhat.com/providers/appinterface_objectstore_test.go
kylape: I did not finish my unit tests, but I ran out of time at the end of the day and figured I could come back later and add them.

controllers/cloud.redhat.com/providers/appinterface_objectstore.go
psav: I guess because having those keys doesn't MAKE it an AWS secret, kinda like a Go interface, though it satisfies, doesn't mean it IS of the type we want and I suppose, the endpoint is the most important piece, and we are saying if it doesn't exist we are just going to assume it is one of our secrets?

controllers/cloud.redhat.com/providers/appinterface_objectstore.go
psav: OK

controllers/cloud.redhat.com/providers/appinterface_objectstore.go
kylape: If you're concerned about the endpoint, we can construct that without any secret data as long as we know the region.  We could leverage a naming convention, but I just don't want to force teams to do more than we're already asking.  That said, I should go see what the secret names are just in case they all conform.

PR #67, opened 2020-10-07
makeComponentUsage for Redis

psav:
* Made test environment setups localized
* Moved makeFn, makeComponent and getNamespacedName to providers
* Reworked RedisTests
* Used makeComponent pattern for Redis
* Added liveness/readiness probes to Redis
* Made utils.UpdateAllOrErr ignore empty objects

### DIFF
--- a/controllers/cloud.redhat.com/utils/utils.go
+++ b/controllers/cloud.redhat.com/utils/utils.go
@@ -25,6 +25,7 @@ import (
 const rCharSet = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
 
 type ClowdObject interface {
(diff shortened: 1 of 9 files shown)


controllers/cloud.redhat.com/utils/utils.go
```diff
@@ -91,7 +91,9 @@ func UpdateAllOrErr(ctx context.Context, cl client.Client, nn types.NamespacedNa
 	updates := map[runtime.Object]Updater{}
 
 	for _, resource := range obj {
-
+		if reflect.ValueOf(obj).IsZero() {
```
kylape, line 94: wouldn't you want to do this check outside of the loop?

kylape (changes requested):
It seems like you are now creating a redis instance when the env is created.  I thought we were creating redis instances on a per-app basis.

PR #74, opened 2020-10-09
Fixed up Minio for makeComponent

psav:


controllers/cloud.redhat.com/providers/objectstore/provider.go
```diff
@@ -15,6 +15,7 @@ import (
 type ObjectStoreProvider interface {
 	p.Configurable
 	CreateBuckets(app *crd.ClowdApp) error
+	CreateObjectStore(env *crd.ClowdEnvironment) error
```
kylape, line 18: The addition of this function to the interface makes this provider interface out of step with all the others.  Jesse and I discussed when the creation of environment-wide resources, and we agreed they should happen within the `New*` functions because the object returned should be fully initialized, which includes the k8s resources required by the provider.

kylape (changes requested):
I think we can discuss whether or not we want to keep all the component creation happening within the `New*` function, but for now we should keep it there for consistency's sake.

Everything else looked good to me.

PR #79, opened 2020-10-14
Fixed up pod naming to avoid clashes

psav:


kylape (approved):
LGTM, minus the failing tests of course

PR #81, opened 2020-10-15
RFR: Update README for unit/functional testing

bsquizz:
Improve our documentation related to running the unit/functional tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,3 +95,6 @@ Run the tests:
 ```
 make test
 ```
+
+If you're just getting started with writing tests in Go, or getting started with Go in general, take
+a look at https://quii.gitbook.io/learn-go-with-tests/
--- a/README.md
+++ b/README.md
@@ -82,12 +82,14 @@ get initialized against a control pane just like it would against a real Openshi
 
 1. Download and install `kubebuilder`.
 [See install instructions here](https://book.kubebuilder.io/quick-start.html#installation).
+
 2. Make sure that the directory you installed kubebuilder in is present in your `PATH`. You should
 see the following executables in that directory:
-```
-etcd  kube-apiserver  kubebuilder  kubectl
-```
-It is recommended you append the directory to your `PATH` in your `.bashrc`, `.zshrc`, or similar.
+    ```
+    etcd  kube-apiserver  kubebuilder  kubectl
+    ```
+    It is recommended you append the directory to your `PATH` in your `.bashrc`, `.zshrc`, or similar.
+
 3. Download and install `kustomize`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,19 +77,24 @@ The diagram below describes how the application accesses the configuration.
...
-enables the operator to get initialized against a control pane just like it
-would against a real Openshift cluster.
+The tests rely on the test environment set up by controller-runtime.  This enables the operator to 
+get initialized against a control pane just like it would against a real Openshift cluster.
+
+1. Download and install `kubebuilder`.
+[See install instructions here](https://book.kubebuilder.io/quick-start.html#installation).
+2. Make sure that `kubebuilder` is present in your `PATH`. It is recommended you append it to your
+`PATH` in your `.bashrc`, `.zshrc`, or similar.
+3. Download and install `kustomize`.
```
psav, line 87: Kustomize isn't required for the current tests - is this required for the new unit tests you added?

bsquizz: I was getting some errors about kustomize not being in my PATH when I ran the tests initially :man_shrugging:

kylape: I think you need kustomize to do the initial code generation done by the SDK.  Worst case devs install kustomize when they don't have to, which doesn't seem that bad.

README.md
```diff
@@ -77,19 +77,24 @@ The diagram below describes how the application accesses the configuration.
...
 
-The tests rely on the test environment set up by controller-runtime.  This
-enables the operator to get initialized against a control pane just like it
-would against a real Openshift cluster.
+The tests rely on the test environment set up by controller-runtime.  This enables the operator to 
+get initialized against a control pane just like it would against a real Openshift cluster.
+
+1. Download and install `kubebuilder`.
+[See install instructions here](https://book.kubebuilder.io/quick-start.html#installation).
+2. Make sure that `kubebuilder` is present in your `PATH`. It is recommended you append it to your
```
kylape, line 85: Just putting the kubebuilder binary in my PATH was not sufficient.  You also need to have `etcd` and `kube-apiserver` binaries available, and afaict you have to specify where they are using the KUBEBUILDER_ASSETS env var.

kylape (changes requested):

README.md
bsquizz: True, when I installed kubebuilder, those two were included. Once they are on the PATH it appears setting the ASSETS is not needed for me.

bsquizz: Updated to mention the other executables

bsquizz: As long as kubebuilder is installed in the default location (/usr/local/kubebuilder/bin) you don't need KUBEBUILDER_ASSETS ... if you place those in a different dir, you do need that env var. Updated README.

PR #82, opened 2020-10-19
Add reconciliation for replica/partition

psav:


controllers/cloud.redhat.com/providers/kafka/provider.go
```diff
@@ -14,7 +14,7 @@ import (
 // KafkaProvider is the interface for apps to use to configure kafka topics
 type KafkaProvider interface {
 	p.Configurable
-	CreateTopic(nn types.NamespacedName, topic *strimzi.KafkaTopicSpec) error
+	CreateTopic(nn types.NamespacedName, topic *strimzi.KafkaTopicSpec, app *crd.ClowdApp) error
```
kylape, line 17: Doesn't `nn` represent the `NamespacedName` of the app that you're now passing in?

kylape: If you're going to pass the app in, I suggest you remove the other two arguments and change the name to `CreateTopics` and then iterate over the topics in each kafka mode.

kylape:
I know I said I'd merge unless there are glaring issues, but..... I can't help myself!

1. do we want to change the requeue behavior in this PR?  It seems unrelated
2. Can you remove all the printf statements (commented or otherwise?)
3. Micro optimization suggestion: maybe call client.List outside the for loop in the strimzi provider mode?

PR #87, opened 2020-10-21
Fix target namespace handling

psav:
* Used to rely on the targetNamespace field being set but now it's
  optional.
* The targetNamespace is set to cloudenv-[ClowdEnvironment.Name] which
  should be safe to be unique because the ClowdEnvironment name
  is unique in a single cluster anyway
* ObjectStore in the status was not being used as status seems to be
  invalid for cluster scoped resources


psav:
Revisit Status

psav:
Status is working - still needed to "create" namespace and decide what to do on removal of the env - should we delete namespace or leave it

kylape (dismissed):

PR #102, opened 2020-11-06
Creates app migration checklist

BlakeHolifield:
Feel free to suggest changes or additions

### DIFF
--- a/docs/migration/checklist.md
+++ b/docs/migration/checklist.md
@@ -1,39 +1,30 @@
 # Migration Checklist
 
+> Suggested order is top to bottom
+
+## App Changes - Settings, Config, and ClowdApp (if applicable)
+  - [ ] database (RDS) is connected via clowder config and is present in ClowdApp
+  - [ ] inMemoryDb (Redis) is connected via clowder config and is present in ClowdApp
+  - [ ] kafka is connected via clowder, topics are present in Clowdapp
+  - [ ] logging is sent to cloudwatch via clowder
+  - [ ] objectstores are connected via clowder, minio is used instead of boto
+  - [ ] webport is present in Clowdapp
+  - [ ] dependent services are present in Clowdapp
+  - [ ] CLOWDER_ENABLED can toggle backwards compatability in app config
+  - [ ] ENV_NAME is parameterized and set to required
+  - [ ] CLOWDER_ENABLED is parameterized and set to required
+  - [ ] MIN_REPLICAS is parameterized
+
 ## Build and Deploy
-  - [ ] Dockerfile
-  - [ ] build_deploy.sh
-  - [ ] pr_check.sh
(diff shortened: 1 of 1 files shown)


docs/migration/checklist.md
```diff
@@ -0,0 +1,26 @@
+# Migration Checklist
+
+### The Source Repo has 
+  - [ ] Dockerfile
+  - [ ] build_deploy.sh
+  - [ ] pr_check.sh
+  - [ ] e2e build disabled (after build pipeline is validated)
```
kylape, line 7: I think e2e should be its own section titled "backwards compatibility"

docs/migration/checklist.md
```diff
@@ -0,0 +1,26 @@
...
+### App config (check if not applicable)
+  - [ ] Kafka topics and bootstrap url
+  - [ ] Object store buckets (minio, s3)
+  - [ ] Databases (RDS, Redis)
```
kylape, line 11: This could be two separate items

docs/migration/checklist.md
```diff
@@ -0,0 +1,26 @@
...
+  - [ ] Logging via Cloudwatch
+  - [ ] Web path and port 
+  - [ ] Metrics path and port  
+  - [ ] Dependent services (rbac, etc) 
+### Clowdapp resource
```
kylape, line 16: We should add all the things not related to pod spec, e.g. kafka topics, web and in-memory db flags, etc.

docs/migration/checklist.md
```diff
@@ -0,0 +1,26 @@
+# Migration Checklist
+
+### The Source Repo has 
+  - [ ] Dockerfile
+  - [ ] build_deploy.sh
+  - [ ] pr_check.sh
+  - [ ] e2e build disabled (after build pipeline is validated)
+### App config (check if not applicable)
+  - [ ] Kafka topics and bootstrap url
+  - [ ] Object store buckets (minio, s3)
```
kylape, line 10: Should we say something about using the minio lib if possible?

kylape (changes requested):
This is a really great idea.  I think we should make as many checklist items linked to the migration doc as possible, though that of course doesn't have to be done before we merge this PR.

docs/migration/checklist.md
bsquizz: There's a few things to cover in e2e-deploy:
* You need to disable the 'quay copier' build config for your app (so that the app's image is not being pushed to quay from 2 places)
* You need to update your app's qa/smoke env YAML files so that your app pulls its image from quay -- if you completely get rid of all webhooks triggering the old BuildConfigs on OpenShift then you'd need to update the ci env YAML for this too.
* Disable the "prod" env option in your app's Jenkins deploy job

I can add more details for each of these to whatever doc we decide to house this info in ...

kylape (approved):

PR #105, opened 2020-11-09
Add AutoScaling via Keda.sh operator

lindgrenj6:
Adding autoscaling via (keda)[keda.sh] for Insights Hackathon fall 2020

\# TODO:
- [x] Define Autoscaling schema (under `podSpec`)
- [x] Create `ScaledObject` resource when creating ClowdApp CR
- [ ] Remove `ScaledObject` resource when deleting ClowdApp CR
- [ ] Add logic to **NOT** update replica count in deployment if autoscaling is present.
- [ ] tests (or maybe not since this is a hackathon...)


kylape (commented):
The implementation looks good to me.  My main feedback is regarding the design of the spec in ClowdApp.  We need to be able to support autoscaling for webservices as well.  Does that mean we create a list of AutoScalingSpecs?  Or should we create two subspecs under AutoScalingSpec, one for kafka and one for web?  Do we need to plan for even more types of autoscaling resources down the road?

Alternatively we could do away with a separate AutoScalingSpec and instead add autoscaling attributes to the WebSpec (we'd have to change `web` to be a struct) and `KafkaTopicSpec`. WDYT?

lindgrenj6:
@kylape Yeah I like the idea of having 2 different specs - one for web and one for kafka. That way we can still validate the fields to create the `ScaledObject`. The main problem is that for the Trigger array under the ScaledObjectSpec is just a `map[string]string` - would be a pain to not have validation on that and then have people going "why isn't my scaler working".

kylape:
I think we could be more restrictive than `map[string]string`.  This can piggyback off the work we have planned for specifying SLIs/SLOs in a ClowdApp and generating PrometheusRules from them.  We expose a few autoscaling params, pair it with SLI/SLO data, and then Clowder can write out the prometheus-based ScaledObjectSpecs.

bennyturns:
This so out of date we will need to rework a ton to get this merged.  We are DEF interested in this, lets open a new PR for when we want to get this merged.

PR #122, opened 2020-11-24
Added status checking

psav:


controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -142,9 +142,48 @@ func (r *ClowdAppReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	app.Status.ReadyDeployments = readyDeployments
+	err = proxyClient.Status().Update(ctx, &app)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	return ctrl.Result{Requeue: requeue}, nil
 }
 
+func DeploymentStatusChecker(deployment *apps.Deployment) (string, bool, error) {
```
psav, line 171: This came from kubectl

kylape:
The code looks good but there are some clear opportunities to apply the DRY principle.  I wonder if we could create a new file `status.go`, create a common struct/interface that contains the deployment status, and move the `GetOwnedDeployments` function in there.  Rename `GetOwnedDeployments` to `FilterOwnedDeployment` and pass in the UID instead of having it be a member of the struct.

wdyt?  I'm not convinced it's better, but it feels like it would DRY up the code a bit.

kylape:
Well I went ahead and tried out the refactor.  Lemme know if you think it's any better.

psav:
Yup - think it's much better

config/rbac/role.yaml
```diff
@@ -9,12 +9,12 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - configmaps
```
psav, line 12: This is a result of another change, but it's fine to leave in here too

kylape:
@psav looks like my refactor is gone?  was that intentional?

PR #126, opened 2020-11-25
Addition of SOP documentation

psav:


docs/sop.rst
```diff
@@ -0,0 +1,135 @@
+Operating Clowder
+=================
+
+Two primary aspects of operating Clowder: Operating the apps managed by Clowder and operating
+Clowder itself.
+
+Contrary to many other deployment models where apps configure themselves, an application that is
```
kylape, line 7: "configure themselves" sounds odd.  I wonder if saying something more like "where devs are burdened with the repsponsibility of wiring up all aspects of their applications in k8s", though that is much more wordy.

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+managed by Clowder will have configuration presented to it and is expected to use it. Clowder
```
kylape, line 8: apps are "expected" to use it because there is no other way to obtain this information.  I know this is an SOP and not marketing material, but I would like to use this paragraph to emphasize the core value of Clowder by explicitly stating how we are removing the configuration effort from devs.

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+governs many different aspects of an application's configuration, from defining the port it should
+listen on, to metrics, kafka topics, service hostnames, and more. In this way a common configuration
+format is presented to the application no matter the envrionment it is running in, enabling a far
+easier development experience.
```
kylape, line 12: ok yeah you get to the benefits here, but I wonder if this should be explained at the beginning of the paragraph rather than the end?

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+
+Operating Apps Managed by Clowder
+---------------------------------
+
+ClowdEnvironment
+++++++++++++++++
+
+**abbreviated to ``env`` in k8s**
```
kylape, line 20: ugh I wish the back ticks worked in bold text, but alas...

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+
+The ``ClowdEnvironment`` CRD is responsible for configuring the environment that the Clowder enabled
```
kylape, line 22: what is an environment comprised of?  From Clowder's perspective, an environment is made up of the configuration and deployments needed to operate a complete clouddot environment.  I guess just saying "environment" here is kind of redundant.

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+apps will interact with. It is a *cluster scoped* CRD and thus must have a unique name inside the
+k8s cluster. For production environments it is usual to have only one ``ClowdEnvironment``, whereas
+in other scenarios &mdash; such as ephemeral testing &mdash; Clowder enables the management of
```
kylape, line 25: I don't know how to make em-dashes in RST

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+multiple environments that operate completely independently from each other.
+
+Providers
+^^^^^^^^^
+
+Clowder provides a number of providers which govern the creation of services, e.g. Kafka topics,
```
kylape, line 31:
```suggestion
An environment's specification is broken into **providers**, which govern the creation of services, e.g. Kafka topics,
```

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+object storage, etc, that applications may depend on. The ``ClowdEnvironment`` CRD configures these
+providers principally by making use of a provider's mode.
```
kylape, line 33:
```suggestion
providers principally by making use of a provider's **mode**.
```

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+
+Modes
+^^^^^
+
+Providers often operate in different modes. As an example the Kafka provider can operate in three
```
kylape, line 38: This first sentence seems redundant with the last sentence of the previous paragraph.  Maybe remove that last sentence??

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+- `Metrics <https://redhatinsights.github.io/clowder/api_reference.html#k8s-api-cloud-redhat-com-clowder-v2-apis-cloud-redhat-com-v1alpha1-metricsconfig>`_
+
+Target Namespace
+^^^^^^^^^^^^^^^^
+
+Clowder will provision two types of resources during operation, *environmental* and *application*
+specific. Application resources will be deployed into the same namespace the ClowdApp resides in.
+Environmental resources, such as the Kafka/Zookeeper from the exmaple in the *Modes* section, will
+be placed in the ``ClowdEnvironment``'s target namespace. This is configured by setting the
+``targetNamespace`` attribute of the ``ClowdEnvironment`` and should always be set for production
```
kylape, line 63: maybe remove the assertion that it should be set in prod? Seems unnecessary.

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+- `Object storage <https://redhatinsights.github.io/clowder/api_reference.html#k8s-api-cloud-redhat-com-clowder-v2-apis-cloud-redhat-com-v1alpha1-objectstoreconfig>`_
+- `In-memory DB <https://redhatinsights.github.io/clowder/api_reference.html#k8s-api-cloud-redhat-com-clowder-v2-apis-cloud-redhat-com-v1alpha1-inmemorydbconfig>`_
+- `Relational DB <https://redhatinsights.github.io/clowder/api_reference.html#k8s-api-cloud-redhat-com-clowder-v2-apis-cloud-redhat-com-v1alpha1-databaseconfig>`_
+- `Web <https://redhatinsights.github.io/clowder/api_reference.html#k8s-api-cloud-redhat-com-clowder-v2-apis-cloud-redhat-com-v1alpha1-webconfig>`_
+- `Metrics <https://redhatinsights.github.io/clowder/api_reference.html#k8s-api-cloud-redhat-com-clowder-v2-apis-cloud-redhat-com-v1alpha1-metricsconfig>`_
+
+Target Namespace
+^^^^^^^^^^^^^^^^
+
+Clowder will provision two types of resources during operation, *environmental* and *application*
```
kylape, line 59: I think we can omit the app-specific info here since we're still under the ClowdEnv section.

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+
+**abbreviated to ``app`` in k8s**
+
+The ``ClowdApp`` CRD is responsible for configuring an application and is namespace scoped. Any
+resources Clowder creates on behalf of the application will reside in the same namespace that the
+``ClowdApp`` resources is applied to. As such the ``ClowdApp`` name must be unique within a
+particular namespace.  An ``ClowdApp`` does not have to be placed in the ``ClowdEnvironment``'s
+target namespace.
+
+A ``ClowdApp`` may define multiple services inside it. These services, though defined by a
```
kylape, line 78: You refer to the `pod` specs as "services", causing an inconsistency.  Is that what we want to call them?

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+specification that is very similar to the k8s pod specification, will be deployed as individual
+deployment resources.  Functionally, defining multiple applications in the same ``ClowdApp``
+specification allows the sharing of some infrastructure dependencies such as databases.
+Applications in different ClowdApp's should not expect to be able to share databases.
```
kylape, line 82: hmm I don't know if this is true anymore.  I think it may be hard to hold this assertion as more apps onboard

docs/sop.rst
```diff
@@ -0,0 +1,135 @@
...
+respectively. By default these are set to 8000 for the web service port and 9000 for the metrics
+port.
+
+Clowder will also set certain fields in the pod spec, inline with best practice, such as pull
+policy, and anti-affinity.
+
+Clowder will also create a ``Secret`` resource which will contain the generated configuration
+for that app. This secret will be mounted at ``/cdappconfig.json`` and will be consumed by the app
+to configure itself on startup.
+
```
kylape, line 126: What about local redis and postgresql DBs?

docs/sop.rst
psav: Agreed

docs/sop.rst
psav: I know it's sad :(

docs/sop.rst
psav: Sure - but if a section is navigated to having a brief context isn't a bad idea.

docs/sop.rst
psav: What do you mean?

docs/sop.rst
psav: I'm not sure, should we do a spec change to bring the terminology inline before this is used outside of the platform?

kylape: Yeah but we should make sure we choose a good name, which is a challenge.

psav: Agreed......

PR #131, opened 2020-12-01
Added redis config map for ephemeral env

psav:
* Redis requires a mode to be set for it to operate as we require in ephemeral env


controllers/cloud.redhat.com/providers/inmemorydb/redis.go
```diff
@@ -71,16 +93,35 @@ func makeLocalRedis(o obj.ClowdObject, dd *apps.Deployment, svc *core.Service, p
 		TimeoutSeconds:      2,
 	}
 
+	dd.Spec.Template.Spec.Volumes = []core.Volume{{
+		Name: nn.Name,
+		VolumeSource: core.VolumeSource{
+			ConfigMap: &core.ConfigMapVolumeSource{
+				LocalObjectReference: core.LocalObjectReference{
+					Name: fmt.Sprintf("%v", nn.Name),
```
kylape, line 101: Is there a reason you can't just do `Name: nn.Name,`?

PR #134, opened 2020-12-02
Add check for commit message length

kylape:
This should also catch the formatting mistake where the second line is
not left blank.

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -2,6 +2,15 @@
 
 echo "$MINIKUBE_SSH_KEY" > minikube-ssh-ident
 
+while read line; do
+    if [ ${#line} -ge 100 ]; then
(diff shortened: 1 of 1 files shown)


kylape:
based on https://stackoverflow.com/a/34664075

psav:
How does this catch the second line not being blank?

psav (commented):
I'm not sure I see how this protects the second line not being blank

kylape:
`git log --pretty=format:%s` prints the subject line from each commit message.  If there is no blank line, it will concatenate all the lines together into one lone line.

psav:
So you're banking on the fact that the summary and second line combined will break the line length?

kylape:
Yes.

kylape:
It's admittedly a "cheap" solution.

PR #136, opened 2020-12-03
Add Kafka Topic creation in local mode

psav:
* Kafka will now create topics in local mode
* An additional port-forward is required to connec to kafka


kylape:
```
build command-line-arguments: cannot find module for path github.com/confluentinc/confluent-kafka-go/kafka/librdkafka
```

kylape:
That was much easier than I expected

controllers/cloud.redhat.com/providers/kafka/localkafka.go
```diff
@@ -43,6 +46,12 @@ func (k *localKafka) CreateTopics(app *crd.ClowdApp) error {
 				RequestedName: topic.TopicName,
 			},
 		)
+
+		conn, err := kafka.DialLeader(k.Ctx, "tcp", host, topicName, 0)
```
bsquizz, line 50: I am guessing this results in a single partition on the topic, right? Do we need to go with the second approach here so we can create the topic with the explicit number of partitions specified in our `ClowdEnvironment KafkaTopicSpec`?

https://github.com/segmentio/kafka-go#to-create-topics

PR #138, opened 2020-12-04
Add Field to Override Database Image in ClowdApp

BlakeHolifield:
* Adds an Image field
* local db provider defaults to ClowdEnv image if override is not present


psav (changes requested):
There are a few items to address here

- I'm not sure that we want to give people the opportunity for custom DBs, I'd far prefer to get hll into the base image so that everyone can benefit. The more custom stuff we allow, the more diversity we could get.
- There are some big changes to the linewraps in the CRDs, are you using a different version of the operator-sdk?
- The `go.sum` file has a bunch of libraries in there that I'm not even sure we use.

BlakeHolifield:
Closing to implement a different fix.

PR #142, opened 2020-12-08
App status for ClowdEnvs

BlakeHolifield:
Closes https://issues.redhat.com/browse/RHCLOUD-11038

### DIFF
--- a/apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
+++ b/apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
@@ -19,7 +19,6 @@ import (
 	"strings"
 
 	"cloud.redhat.com/clowder/v2/apis/cloud.redhat.com/v1alpha1/common"
-	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/config"
 	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/utils"
 	core "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
@@ -240,7 +239,12 @@ type ClowdEnvironmentStatus struct {
 	// Important: Run "make" to regenerate code after modifying this file
 	TargetNamespace string                  `json:"targetNamespace"`
 	Deployments     common.DeploymentStatus `json:"deployments"`
-	AppEndpoints    []config.AppEndpoint    `json:"appEndpoints"`
+	AppEndpoints    []AppEndpoint           `json:"appEndpoints"`
+}
+
+type AppEndpoint struct {
+	Name     string `json:"name"`
+	Hostname string `json:"hostname"`
 }
 
 // +kubebuilder:object:root=true
--- a/apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
+++ b/apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
@@ -239,15 +239,15 @@ type ClowdEnvironmentStatus struct {
 	// Important: Run "make" to regenerate code after modifying this file
 	TargetNamespace string                  `json:"targetNamespace"`
 	Deployments     common.DeploymentStatus `json:"deployments"`
-	Apps            []AppStatus             `json:"apps,omitempty"`
+	Apps            []AppInfo               `json:"apps,omitempty"`
 }
 
-type AppStatus struct {
-	Name        string             `json:"name"`
-	Deployments []DeploymentStatus `json:"deployments"`
+type AppInfo struct {
+	Name        string           `json:"name"`
+	Deployments []DeploymentInfo `json:"deployments"`
 }
 
-type DeploymentStatus struct {
+type DeploymentInfo struct {
 	Name     string `json:"name"`
 	Hostname string `json:"hostname,omitempty"`
 }
--- a/apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
+++ b/apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 
 	"cloud.redhat.com/clowder/v2/apis/cloud.redhat.com/v1alpha1/common"
+	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/config"
 	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/utils"
 	core "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
@@ -239,6 +240,7 @@ type ClowdEnvironmentStatus struct {
 	// Important: Run "make" to regenerate code after modifying this file
 	TargetNamespace string                  `json:"targetNamespace"`
 	Deployments     common.DeploymentStatus `json:"deployments"`
+	AppEndpoints    []config.AppEndpoint    `json:"appEndpoints"`
 }
 
 // +kubebuilder:object:root=true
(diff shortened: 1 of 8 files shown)


controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -177,12 +183,86 @@ func runProvidersForEnv(provider providers.Provider) error {
 func (r *ClowdEnvironmentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	r.Recorder = mgr.GetEventRecorderFor("env")
 	return ctrl.NewControllerManagedBy(mgr).
-		Owns(&apps.Deployment{}).
-		Owns(&core.Service{}).
```
kylape, line 181: why were these two lines removed?

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -157,7 +157,7 @@ func (r *ClowdAppReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	return ctrl.Result{Requeue: requeue}, nil
 }
 
-// SetupWithManager sets up wi
+// SetupWithManager sets up with Manager
```
kylape, line 160: haha

controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -177,12 +183,86 @@ func runProvidersForEnv(provider providers.Provider) error {
...
+		if app.Spec.Pods != nil {
+			app.ConvertToNewShim()
+		}
+
+		appstatus := crd.AppStatus{
+			Name:        app.Name,
+			Deployments: []crd.DeploymentStatus{},
+		}
+
+		if app.Spec.EnvName == p.Env.Name {
```
kylape, line 242: imho it would be better to change this to 

```suggestion
		if app.Spec.EnvName != p.Env.Name {
		    continue
		}
```

and move it up to the top of the loop

apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
```diff
@@ -239,6 +239,17 @@ type ClowdEnvironmentStatus struct {
 	// Important: Run "make" to regenerate code after modifying this file
 	TargetNamespace string                  `json:"targetNamespace"`
 	Deployments     common.DeploymentStatus `json:"deployments"`
+	Apps            []AppStatus             `json:"apps,omitempty"`
+}
+
+type AppStatus struct {
```
kylape, line 245: The name `AppStatus` is a bit odd to me -- you're not giving the status of an app :slightly_smiling_face: .  Maybe something like `AppInfo` or `AppListing`?

apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
```diff
@@ -239,6 +239,17 @@ type ClowdEnvironmentStatus struct {
...
+	Name        string             `json:"name"`
+	Deployments []DeploymentStatus `json:"deployments"`
+}
+
+type DeploymentStatus struct {
```
kylape, line 250: Some comment here -- `DeploymentStatus` is odd because you're not giving the status.

controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -177,12 +183,86 @@ func runProvidersForEnv(provider providers.Provider) error {
...
+				if pod.Web {
+					deploymentStatus.Hostname = fmt.Sprintf("%s.%s.svc", deploymentStatus.Name, app.Namespace)
+				}
+				appstatus.Deployments = append(appstatus.Deployments, deploymentStatus)
+			}
+			apps = append(apps, appstatus)
+		}
+	}
+
+	fmt.Printf("\n%v\n", apps)
```
kylape, line 256: probably should remove this debug statement

controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -177,12 +183,86 @@ func runProvidersForEnv(provider providers.Provider) error {
...
+}
+
+func (r *ClowdEnvironmentReconciler) generateAppStatus(p providers.Provider) error {
+	// Get all the ClowdApp resources
+	appList := crd.ClowdAppList{}
+	r.Client.List(p.Ctx, &appList)
+	apps := []crd.AppStatus{}
+
+	// Populate
+	for _, app := range appList.Items {
```
kylape, line 231: should check to see if `app.GetDeletionTimestamp() != nil` skip the app if true.

controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -96,8 +99,11 @@ func (r *ClowdEnvironmentReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
 		Env:    &env,
 	}
 
+	err = r.generateAppStatus(provider)
```
kylape, line 102: There is a call to `SetDeploymentStatus` later on in this function that populates another part of the status section.  Can you move this call next to that one?  They should probably share the same parameter set as well.  Either way works, just so they're consistent.  Names could be consistent, too -- one starts with `Generate` and the other with `Set`.

controllers/cloud.redhat.com/clowdenvironment_controller.go
kylape: The reasons are 1) fewer indented lines, and 2) you don't create structs that get thrown away.  pretty minor stuff

controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -177,12 +183,86 @@ func runProvidersForEnv(provider providers.Provider) error {
...
+				}
+				appstatus.Deployments = append(appstatus.Deployments, deploymentStatus)
+			}
+			apps = append(apps, appstatus)
+		}
+	}
+
+	fmt.Printf("\n%v\n", apps)
+	p.Env.Status.Apps = apps
+	err := r.Client.Status().Update(p.Ctx, p.Env)
```
kylape, line 258: This line is unnecessary because the client is called to update the status in the calling function.

kylape (changes requested):

controllers/cloud.redhat.com/clowdenvironment_controller.go
BlakeHolifield: I thought these were the cause of the loop and removed them. It was elsewhere.

kylape: I tried running the e2e tests locally with those two lines added back in, and there was no infinite reconciliation loop.  Is there a particular scenario I need to set up by hand to trigger the loop?

BlakeHolifield: No that should be totally resolved. It was a misunderstanding on my end. The loop was caused by the constant stream of appends we mentioned in the call this morning.

config/crd/bases/cloud.redhat.com_clowdenvironments.yaml
```diff
@@ -279,6 +300,7 @@ spec:
                 this file'
               type: string
           required:
+          - apps
```
psav, line 303: I don't think it should be required, but I'm not sure where this comes from - it shouldn't be required because if there are no apps, this will not get populated. Will it just get initialized to an empty array?

controllers/cloud.redhat.com/clowdapp_controller.go
psav: Yeh I keep meaning to fix that one..........

PR #146, opened 2020-12-09
Migrate Tests to new Deployments structure

psav:
This should be merged after we have converted everyone over to the new structure for the ClowdApps


psav:
All running apps in stage are converted - this can now be merged

PR #162, opened 2020-12-16
Updated ranges

psav:


kylape:
I had just pushed a bunch of changes to the dashboard, so I went and made your changes again and applied them in a separate commit.

PR #163, opened 2020-12-17
Create CyndiPipeline resources in kafka providers

bsquizz:
* Add `connectNamespace` to "kafka" provider in `ClowdEnvironment` spec which indicates the namespace that the "Kafka Connect" cluster resides in for the given ClowdEnvironment. If unset, we default to be the same namespace as 'kafka'
* Add param "cyndi: <boolean>" to ClowdApp spec

If kafka provider mode is "app-interface", and ClowdApp has "cyndi: true", Clowder will simply verify that a `CyndiPipeline` resource exists in the `connectNamespace`. At the moment these are being defined in app-interface here: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/xjoin/deploy-cyndi.yml

If provider mode is "operator", and ClowdApp has "cyndi: true", Clowder will drop a `CyndiPipeline` resource into the `connectNamespace` that looks something like:

```
apiVersion: cyndi.cloud.redhat.com/v1alpha1
kind: CyndiPipeline
metadata:
  name: {ClowdApp name}
spec:
  appName: {ClowdApp name}
```

If provider mode is 'local', we will drop the `CyndiPipeline` resource into the target env namespace. For the scope of this PR, we are assuming kafka-connect is already provisioned in that namespace (Clowder is not yet creating it)

In 'app-interface' mode, it is assumed that all the configuration that cyndi-operator needs (correct CyndiPipeline content, 'cyndi' ConfigMap, and secrets for accessing respective app databases) are already present in the kafka connect namespace (because they've been provisioned via definitions in app-interface). We'll add more validations in a future PR.


apis/cyndi-operator/v1alpha1/cyndipipeline_types.go
```diff
@@ -0,0 +1,115 @@
+/*
```
psav, line 1: Is this and the next few files all copied over from the cyndipipeline operator? and if so, do we expect them to change and how should we keep them in sync? Is this because cyndi itself is private?

controllers/cloud.redhat.com/providers/kafka/localkafka.go
```diff
@@ -28,6 +28,15 @@ type localKafka struct {
 }
 
 func (k *localKafka) Provide(app *crd.ClowdApp, c *config.AppConfig) error {
+	if app.Spec.Cyndi {
+		// for now we're assuming the kafka connect cluster is already present in the namespace
+		err := createCyndiPipeline(
+			k.Ctx, k.Client, app.Name, getConnectNamespace(k.Env), getConnectClusterName(k.Env))
```
psav, line 34: Stylistically, it would be nice to have the ending `)` on the line below

controllers/cloud.redhat.com/providers/kafka/provider.go
```diff
@@ -22,6 +23,28 @@ func GetKafka(c *p.Provider) (p.ClowderProvider, error) {
 	}
 }
 
+func getKafkaNamespace(e *crd.ClowdEnvironment) string {
```
psav, line 26: I'm a little reticent about putting mode specific knowledge in the top level package

go.mod
```diff
@@ -3,15 +3,16 @@ module cloud.redhat.com/clowder/v2
 go 1.13
 
 require (
-	github.com/go-logr/logr v0.1.0
+	github.com/go-logr/logr v0.2.1
```
psav, line 6: I'd prefer not to update all these mods in this PR, same as the one below, unless there is a good reason for this?

psav (changes requested):

psav (changes requested):
The docs will also need to be regenerated for this as we are adding something to the spec, details are in the README, however if you get stuck, shout!

go.mod
bsquizz: Yes, I had a separate commit that explained the reason but it got rolled into a rebase: I had to update k8s packages to 0.19.3 to pick up `metav1.Condition`

https://github.com/RedHatInsights/cyndi-operator/blob/master/api/v1alpha1/cyndipipeline_types.go#L79

apis/cyndi-operator/v1alpha1/cyndipipeline_types.go
bsquizz: Yes it is copied over from here: https://github.com/RedHatInsights/cyndi-operator/blob/master/api/v1alpha1/

With one change, I removed [this line](https://github.com/RedHatInsights/cyndi-operator/blob/master/api/v1alpha1/cyndipipeline_types.go#L32) because when I tried to 'make install' my k8s setup did not seem to approve of an optional bool with a default (which should have basically no consequence, since the empty value of a bool is `false` anyway)

bsquizz: In future I'd like to *not* copy this, and hopefully we could just `go get` the cyndi operator instead. But yes, since the repo is private, I just copied this in.

controllers/cloud.redhat.com/providers/kafka/provider.go
bsquizz: I'll change this by having the provider implementation tell `getConnectNamespace` which default value to use

apis/cyndi-operator/v1alpha1/cyndipipeline_types.go
psav: Sure - we should make sure that @we have some way of knowing when it's changed so we can sync it

bundle/tests/scorecard/kuttl/test-cyndi-local-kafka/01-install.yaml
```diff
@@ -0,0 +1,48 @@
...
+kind: ClowdApp
+metadata:
+  name: myapp
+  namespace: test-cyndi-local-kafka
+spec:
+  cyndi: true
+  envName: test-cyndi-local-kafka
+  pods:
+  - name: processor
+    image: quay.io/psav/clowder-hello
```
psav, line 48: newline here pls

psav:
Do we merge this now - or add the extra stuff @BlakeHolifield is working on into it?

bsquizz:
Well, the plan was to add in full cyndi support over the course of several PRs. What he's working on should be able to go in as a separate piece, and even once his work goes in, there are still a couple more PRs to go after that. This PR was just the first step at adding cyndi support

BlakeHolifield:
Yeah I'd like to get this in and add my changes in a new PR

kylape (changes requested):
I think we should switch the data type of the cyndi attribute

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -189,6 +189,9 @@ type ClowdAppSpec struct {
 	// A list of optional dependencies in the form of the name of the ClowdApps that are
 	// will be added to the configuration when present.
 	OptionalDependencies []string `json:"optionalDependencies,omitempty"`
+
+	// If true, Clowder configures 'cyndi' database syndication for this app
+	Cyndi bool `json:"cyndi,omitempty"`
```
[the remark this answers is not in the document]
bsquizz: yeah I have thought about how we may want to customize the options we pass along to the `CyndiPipeline`

OK let's make it a struct..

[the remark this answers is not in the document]
bsquizz: Done.. `cyndi` is now a struct that will look like this in YAML:
```
cyndi:
  enabled: bool
  appName: string (optional, default is ClowdApp's name)
  insightsOnly: bool (optional, default false)
```

The presence of the entire struct is optional.

PR #164, opened 2020-12-17
Start of refactor

psav:
Refactor for dynamic-ish providers

* Created a providersRegistration type to deal with the registration and
  ordering of provider resolution, paving the way for resources passing
  between providers.
* Instantiated a singleton called ProvidersRegistration which is called from
  each provider to register its SetupProvider function.
* Controllers now loop through the registered functions in order to create
  resources.
* The Create* functions inside each provider modules have been renamed to
  Provide.
* The Configure function inside each of the provider modules has been removed
  and the configuration step performed as part of the Provide function for each
  provider.
* Individual provider types have been removed as the interface has now been
  standardized and has been replaced with a new type called ClowderProvider.
* The RunAppProvider and RunEnvProvider functions inside each provider module
  have been removed as they are now superfluous.
* The collapsing into a single interface for the providers allowed a singular
  way of invoking the providers action. This further allowed a looping mechanism
  to invoke each provider in order, which is defined by an integer in the
  registration call.

### DIFF
--- a/controllers/cloud.redhat.com/clowdenvironment_controller.go
+++ b/controllers/cloud.redhat.com/clowdenvironment_controller.go
@@ -31,14 +31,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	// Import the providers to initialize them
+	_ "cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers/database"
+	_ "cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers/inmemorydb"
+	_ "cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers/kafka"
+	_ "cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers/logging"
+	_ "cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers/objectstore"
+
 	crd "cloud.redhat.com/clowder/v2/apis/cloud.redhat.com/v1alpha1"
 	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/errors"
 	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers"
-	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers/database"
-	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers/inmemorydb"
-	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers/kafka"
-	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers/logging"
-	"cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers/objectstore"
 	k8serr "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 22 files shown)


kylape (changes requested):
Returning a provider implementation from RunEnvProvider exposes an internal struct that is not intended to be used directly by a client of the provider interface. I'm not sure off the top of my head how to work around that, but I think we need to go back to the drawing board a bit to avoid having to return a provider from RunEnvProvider

psav:
> Returning a provider implementation from RunEnvProvider exposes an internal struct that is not intended to be used directly by a client of the provider interface. I'm not sure off the top of my head how to work around that, but I think we need to go back to the drawing board a bit to avoid having to return a provider from RunEnvProvider

True, but it's the exact same object that is passed back from the ``Get*`` functions

kylape:
Yes, it is, but I think we've recognized the challenge with developing one single interface is that half of the interface is currently implemented in each provider package and the other half on the instantiated provider implementation.  This is because the method on the provider package actually creates the provider implementation.  So one way to solve this is to hide the implementation struct behind the package's interface somehow.

psav:
I totally agree with that - this PR was more about removing some of the extra steps we identified to make it clearer to plan that next step.....i.e...removing the ``Configure`` .....and making the Env take the place of the other function

kylape (approved):
Fair enough.  I'm good with the refactor in this PR given it's just the first step in the bigger refactor.

controllers/cloud.redhat.com/makers/maker.go
```diff
@@ -433,20 +434,19 @@ func (m *Maker) runProviders() (*config.AppConfig, error) {
...
-		return &c, errors.Wrap("setupenv: inmemorydb", err)
-	}
-	if err := database.RunAppProvider(provider, &c, m.App); err != nil {
-		return &c, errors.Wrap("setupenv: database", err)
+	sort.Sort(providers.ProvidersRegistration)
+
+	for _, provAcc := range providers.ProvidersRegistration {
+		if prov, err := provAcc.ProviderGet(&provider); err != nil {
+			return &c, errors.Wrap(fmt.Sprintf("getprov: %s", provAcc.Name), err)
+		} else if err == nil {
```
kylape, line 442: why add the `if`? isn't `err == nil` the only other possibility if the original `if` checks `err != nil`?

controllers/cloud.redhat.com/providers/providers.go
```diff
@@ -16,6 +16,27 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type ProviderAccessor struct {
+	ProviderGet  func(c *Provider) (ClowderProvider, error)
+	Order        int
+	Name         string
+	SetupRequeue bool
```
kylape, line 23: maybe rename to `RequeueOnFail`?

controllers/cloud.redhat.com/providers/providers.go
```diff
@@ -24,11 +45,8 @@ type Provider struct {
 	Env    *crd.ClowdEnvironment
 }
 
-// Configurable is responsible for applying the respective section of
-// AppConfig.  It should be called in the app reconciler after all
-// provider-specific API calls (e.g. CreateBuckets) have been made.
-type Configurable interface {
-	Configure(c *config.AppConfig)
+type ClowderProvider interface {
+	SetupProvider(app *crd.ClowdApp, c *config.AppConfig) error
```
kylape, line 49: what if this func is just called `Provide`?

controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -31,14 +31,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	// Import the providers to initialize them
```
kylape, line 34: Could these imports be pushed to `providers.go`? you could perhaps then do the sort in there, too.

controllers/cloud.redhat.com/providers/providers.go
```diff
@@ -16,6 +16,27 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type ProviderAccessor struct {
+	ProviderGet  func(c *Provider) (ClowderProvider, error)
```
kylape, line 20: I feel like `SetupProvider` would be a better name here because it implies some more action beyond a simple fetch.

controllers/cloud.redhat.com/clowdenvironment_controller.go
psav: No, there is a cyclical issue. I wanted the definition of the ProviderAccessor to be inside the ``provider`` file for the provider. In doing this it means that we need to import them where we use them, we can't import them inside the ProviderRegistration, because they themselves manipulate the provider registration. Thinking of another way to do it.

PR #178, opened 2021-01-11
Shared DB functionality

psav:


bundle/tests/scorecard/kuttl/test-multi-app-interface-db/01-pods.yaml
```diff
@@ -0,0 +1,82 @@
...
+      mode: operator
+      path: "/metrics"
+    kafka:
+      namespace: kafka
+      clusterName: my-cluster
+      mode: none
+    db:
+      imageList: 
+        - version: 12
+          image: "registry.redhat.io/rhel8/postgresql-12:1-36"
```
BlakeHolifield, line 23: Do we want to swap these to the quay/postgresql-rds images?

psav: yeh probably makes sense
Can you give me the image name and I'll replace it

BlakeHolifield: `quay.io/cloudservices/postgresql-rds:12-1`
`quay.io/cloudservices/postgresql-rds:10-1`

controllers/cloud.redhat.com/providers/database/appinterface.go
```diff
@@ -69,15 +69,58 @@ func NewAppInterfaceDBProvider(p *p.Provider) (providers.ClowderProvider, error)
...
+		if !found {
+			return errors.New("The requested app was not found in the dependencies")
+		}
+
+		appList := crd.ClowdAppList{}
+		var refApp crd.ClowdApp
+		err := a.Client.List(a.Ctx, &appList)
+
+		if err != nil {
+			return errors.New("Cloud not get app list")
```
kylape, line 100: This is an excellent typo

controllers/cloud.redhat.com/providers/database/appinterface.go
```diff
@@ -69,15 +69,58 @@ func NewAppInterfaceDBProvider(p *p.Provider) (providers.ClowderProvider, error)
...
+		found := false
+		for _, appName := range app.Spec.Dependencies {
+			if app.Spec.Database.SharedDBAppName == appName {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			return errors.New("The requested app was not found in the dependencies")
```
kylape, line 92:
```suggestion
			return errors.New("The requested DB's app was not found in the dependencies")
```

controllers/cloud.redhat.com/providers/database/appinterface.go
```diff
@@ -69,15 +69,58 @@ func NewAppInterfaceDBProvider(p *p.Provider) (providers.ClowderProvider, error)
...
+
+		appList := crd.ClowdAppList{}
+		var refApp crd.ClowdApp
+		err := a.Client.List(a.Ctx, &appList)
+
+		if err != nil {
+			return errors.New("Cloud not get app list")
+		}
+
+		found = false
```
kylape, line 103: Do we need to check this boolean?

controllers/cloud.redhat.com/providers/database/appinterface.go
```diff
@@ -69,15 +69,58 @@ func NewAppInterfaceDBProvider(p *p.Provider) (providers.ClowderProvider, error)
...
 		return nil
 	}
 
+	var dbSpec crd.DatabaseSpec
+	var namespace string
+
+	if app.Spec.Database.Name != "" {
+		dbSpec = app.Spec.Database
+		namespace = app.Namespace
+	} else if app.Spec.Database.SharedDBAppName != "" {
```
kylape, line 82: It might be nice to put this else block in a separate function.

controllers/cloud.redhat.com/providers/database/localdb.go
```diff
@@ -28,7 +28,71 @@ func NewLocalDBProvider(p *p.Provider) (providers.ClowderProvider, error) {
 // CreateDatabase ensures a database is created for the given app.  The
 // namespaced name passed in must be the actual name of the db resources
 func (db *localDbProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error {
-	if app.Spec.Database.Name == "" {
+	if app.Spec.Database.Name == "" && app.Spec.Database.SharedDBAppName == "" {
+		return nil
+	}
+
+	if app.Spec.Database.SharedDBAppName != "" {
```
kylape, line 35: I think this code block would also look nicer in a separate function

controllers/cloud.redhat.com/providers/database/localdb.go
```diff
@@ -28,7 +28,71 @@ func NewLocalDBProvider(p *p.Provider) (providers.ClowderProvider, error) {
...
+
+		found := false
```
kylape, line 37: This chunk of code seems to be copypasta, but I don't know if it's enough to justify a common function.

kylape (changes requested):
Few minor observations.

kylape:
Can you explain why the `Name` attribute in the DB spec is insufficient?  What if two apps want to share a DB and they both use the same DB name?  Wouldn't that effectively be sharing a DB?

bundle/tests/scorecard/kuttl/test-cron-job/01-pods.yaml
```diff
@@ -18,7 +18,11 @@ spec:
       clusterName: my-cluster
       mode: none
     db:
-      image: "registry.redhat.io/rhel8/postgresql-12:1-36"
+      imageList:
```
kylape, line 21: are all these imageList changes related to the shared db stuff?

psav: No - it was just an opportunistic update

psav:
> Can you explain why the `Name` attribute in the DB spec is insufficient? What if two apps want to share a DB and they both use the same DB name? Wouldn't that effectively be sharing a DB?

Yes - but two apps can use the same DB name and not want to be shared and how would I know if I was supposed to be sharing or not, the idea here is that we are instead saying, tell me the App, that you want to copy it from, then if the app does change their db name, the app is insulated.

mkanoor:
Can we add something in the dependency to say I am dependent on the other app and I also needs its database parameters.

kylape:
Digging in my my counterproposal for a little bit.  I'm trying to implement DB sharing by simply setting the same `spec.database.name` in two different ClowdApps.  

I have to not only change the name of the DB in the code, but I also have to remove the app label from the DB as well as _deal with the owner reference_ for both the DB deployment and secret.  I believe you can have more than one owner, but of course we don't handle that in clowder today.

That said, I still believe my change will be significantly less code and more intuitive.  And the semantics of ownership I think fit a bit better with the multiple owners model, contingent on testing the ownership behavior.  E.g. you could deploy advisor-service without having to deploy advisor-api, which may be nice for local development (though it should be easy to just deploy advisor-api regardless).

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -378,3 +384,43 @@ func Labels(labels map[string]string) omfunc {
 		o.SetLabels(labels)
 	}
 }
+
+func GetAppInSameEnv(client client.Client, ctx context.Context, app *ClowdApp, appList *ClowdAppList) error {
+	err := client.List(ctx, appList)
```
kylape, line 389: could we use a `client.MatchingFields` here to remove the need for the `for` loop below?

https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/client#MatchingFields

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -378,3 +384,43 @@ func Labels(labels map[string]string) omfunc {
...
+		appList.Items = append(appList.Items, iapp)
+	}
+	return nil
+}
+
+func GetAppForDBInSameEnv(client client.Client, ctx context.Context, app *ClowdApp) (*ClowdApp, error) {
+	appList := &ClowdAppList{}
+	var refApp ClowdApp
+
+	err := GetAppInSameEnv(client, ctx, app, appList)
```
kylape, line 413: could you use `client.MatchingFields` here to combine both the `EnvName` and `Name` checks?

controllers/cloud.redhat.com/providers/database/appinterface.go
```diff
@@ -69,15 +69,38 @@ func NewAppInterfaceDBProvider(p *p.Provider) (providers.ClowderProvider, error)
 }
 
 func (a *appInterface) Provide(app *crd.ClowdApp, c *config.AppConfig) error {
-	if app.Spec.Database.Name == "" {
+	if app.Spec.Database.Name == "" && app.Spec.Database.SharedDBAppName == "" {
```
kylape, line 72: what if neither `Name` nor `SharedDBAppName` are empty?

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -50,6 +53,9 @@ type DatabaseSpec struct {
 	// name of the logical database inside the database server in (*_local_*) mode
 	// and the name of the secret to be used for Database configuration in (*_app-interface_*) mode.
 	Name string `json:"name,omitempty"`
+
+	// Defines the Name of the app to share a database from
+	SharedDBAppName string `json:"sharedDbAppName,omitempty"`
```
kylape, line 58: Have you considered defining `shared bool` instead of `SharedDBAppName string`?  We'd then rely on `Name` for both the shared and non-shared DB name.

psav: I did, but this would be overloading the term as we look for the "app" name to share from, this insulates the dependent app from a dbname change, as they ask for the app, not the db.

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
psav: I fixed it for one, unfortunately, with the cache, you can only match on a single field, and I had to add a custom index for it as it is :)

kylape (approved):

PR #183, opened 2021-01-15
Added ops metrics

psav:


BlakeHolifield:
It seems like most of this is also in #182. Are you thinking you'll keep the metric changes in there or split them out here?

controllers/cloud.redhat.com/proxy_client.go
```diff
@@ -21,21 +22,34 @@ type ProxyClient struct {
 	client.Client
 }
 
+func (p *ProxyClient) Get(ctx context.Context, key types.NamespacedName, obj runtime.Object) error {
+	clientOpsMetric.With(prometheus.Labels{"operation": "get"}).Inc()
```
kylape, line 26: Are you okay knowing that this get will often return locally cached data?

psav: Yes

psav: It's just to track these things

PR #190, opened 2021-01-21
Removed deprecated Pods support

psav:


psav:
Waiting till all production systems no longer use pods support

psav:
It's probably going to be better to get rid of this through the mutating webhooks

PR #191, opened 2021-01-26
Add detailed dev environment setup steps to README

bsquizz:
Also to make the build/testing process a little more efficient: use the unauthenticated RH registry in the Dockerfile, and use `set -e` in e2e-test shell scripts.

Also removed `selfLink` from minio/strimzi kuttl tests -- I don't think our test needs to care about the presence of this attribute -- on my local k8s v1.20.2 env the tests were failing with:
```
    case.go:233: resource KafkaTopic:test-kafka-strimzi-kafka/topicthree-test-kafka-strimzi-test-kafka-strimzi: .metadata.selfLink: key is missing from map
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM registry.redhat.io/ubi8/go-toolset:1.13.15 as builder
+FROM registry.access.redhat.com/ubi8/go-toolset:1.13.15 as builder
 
 USER 0
 WORKDIR /workspace
--- a/README.md
+++ b/README.md
@@ -95,13 +95,47 @@ There are new features coming up as well:
 
 [Design docs](https://github.com/RedHatInsights/clowder/tree/master/docs/)
 
-## Dependencies
+## Dev Environment Setup
 
-- [Operator SDK](https://github.com/operator-framework/operator-sdk/releases)
-- [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder/releases)
-- [kustomize](https://github.com/kubernetes-sigs/kustomize/releases)
-- Either Codeready Containers or a remote cluster where you have access to
-  create CRDs.
+* Install [operator-sdk](https://sdk.operatorframework.io/docs/installation/#install-from-github-release)
+* Install the latest [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+* Install [krew](https://krew.sigs.k8s.io/docs/user-guide/setup/install/)
+* Install kuttl:
+  ```
+  kubectl krew install kuttl
+  ```
+* Install [kubebuilder](https://book.kubebuilder.io/quick-start.html#installation)
+
+  * Normally kubebuilder is installed to `/usr/local/kubebuilder/bin`. You should see the following
+executables in that directory:
+    ```
+    etcd  kube-apiserver  kubebuilder  kubectl
+    ```
+    You may want to append this directory to your `PATH` in your `.bashrc`, `.zshrc`, or similar.
+
+    **NOTE**: If you choose to place the kubebuilder executables in a different path, make sure to
+    use the `KUBEBUILDER_ASSETS` env var when running tests (mentioned in `Unit Tests` section below)
+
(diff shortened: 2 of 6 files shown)


psav:
A few thoughts - 
1) A part of me feels weird about changing the default image location, to be honest, I think it would perhaps be better if we put an ignore on it, because it's oing to change every time someone runs a test.
2) We perhaps should include a note telling people that they should install the quay secret to get the cloud apps working
3) The unauthenticated container you altered, should we also change our postgres images to point to the same url to avoid people having to use a rh_pull_secret altogether?

bsquizz:
> 1. A part of me feels weird about changing the default image location, to be honest, I think it would perhaps be better if we put an ignore on it, because it's oing to change every time someone runs a test.

Are you referring to [this](https://github.com/RedHatInsights/clowder/blob/master/Makefile#L13-L18) in the Makefile? What do you mean by 'put an ignore on it' though?

> 2. We perhaps should include a note telling people that they should install the quay secret to get the cloud apps working

Yes, true. Possible idea: We could host the secret for `quay-cloudservices-pull` (or create a new robot account that has similar read-only access) in our Vault instance, and write a quick script that pulls the secret out of vault and `kubectl apply`'s it... the script would require you to input your vault token. 

This brings up a good point about how we're going to manage secrets in local deployments. In the ephemeral cluster we rely on qontract-reconcile to insert secrets into namespaces, but obviously we can't do that on the local deployments.

> 3. The unauthenticated container you altered, should we also change our postgres images to point to the same url to avoid people having to use a rh_pull_secret altogether?

I don't think we can. Those images don't appear to be available on unauthenticated registries. See https://catalog.redhat.com/software/containers/search?q=postgresql&p=1

Should we just use @BlakeHolifield's custom built images via `quay.io/cloudservices/postgresql-rds:10-1` or `quay.io/cloudservices/postgresql-rds:12-1` everywhere?

bundle.Dockerfile
```diff
@@ -5,13 +5,11 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=clowder
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.bundle.channel.default.v1=
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.0.1
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.3.0
```
psav, line 8: We didn't update our SDK version yet, can we revert this change?

bundle/metadata/annotations.yaml
```diff
@@ -1,11 +1,10 @@
 annotations:
-  operators.operatorframework.io.bundle.channel.default.v1: ""
   operators.operatorframework.io.bundle.channels.v1: alpha
   operators.operatorframework.io.bundle.manifests.v1: manifests/
   operators.operatorframework.io.bundle.mediatype.v1: registry+v1
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: clowder
-  operators.operatorframework.io.metrics.builder: operator-sdk-v1.0.1
+  operators.operatorframework.io.metrics.builder: operator-sdk-v1.3.0
```
psav, line 7: Same as above

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/cloudservices/clowder
-  newTag: e3dd531
+  newName: 127.0.0.1:5000/clowder
+  newTag: 2c280b5
```
psav, line 8: I was referring to this :)

psav:
We should probably specify a version of minikube that we support

bundle.Dockerfile
bsquizz: Bah, this was accidentally pulled in.

PR #196, opened 2021-01-26
Moved all the build artifacts

psav:


kylape (changes requested):
I think the only file you moved that needs to stay at the root is `build_catalog.sh`.

PR #203, opened 2021-01-29
Add minikube setup script and move to codegen'd strimzi package

bsquizz:
* Add minikube_setup.sh with strimzi installation
* Use a [separate package](https://github.com/bsquizz/strimzi-client-go) for kafka -- that allows us to use a [code generator](https://github.com/bsquizz/crd-codegen) to create kafka types.
* Use the real strimzi CRDs in our unit tests (good way to validate that the JSON schema code generator has created compatible types)

This lays the foundation for having the operator create Kafka/KafkaConnect resources (coming up in next PR)


apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -203,6 +203,20 @@ type CyndiSpec struct {
 	InsightsOnly bool `json:"insightsOnly,omitempty"`
 }
 
+// KafkaTopicSpec described the desired configuration of kafka topics
```
psav, line 206: Is there a reason we are creating a new object here, when we could use the strimzi one. I realise that previously we were writing a spec and putting it in the `strimzi` package, but the idea was that it "closely" matched the REAL object so that we could ensure that we got the right yaml/json generated.

config/crd/bases/cloud.redhat.com_clowdapps.yaml
```diff
@@ -4491,27 +4491,23 @@ spec:
...
                 properties:
                   config:
                     additionalProperties:
                       type: string
-                    description: A key/value pair describing the configuration of
-                      a particular topic.
+                    description: The topic configuration.
                     type: object
                   partitions:
-                    description: The requested number of partitions for this topic.
```
psav, line 4503: Why remove the descriptions?

controllers/cloud.redhat.com/providers/kafka/appinterface.go
```diff
@@ -73,6 +73,8 @@ func validateKafkaTopic(ctx context.Context, cl client.Client, nn types.Namespac
 	t := strimzi.KafkaTopic{}
 	err := cl.Get(ctx, nn, &t)
 
+	// TODO: validate partitions and replicas?
```
psav, line 76: Validate in what way?

psav (changes requested):
Some comments

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
bsquizz: Yeah, see the comment just below inside the type def. "topicName" is actually an optional attribute in the real KafkaTopicSpec (if topicName is under, strimzi just uses the metadata.name of the KafkaTopic). We want name to be required, so I'm redefining the type here basically to represent our "clowdapp topic config".

If we want to make partitions and replicas optional here, we could set that in this type as well. The real KafkaTopic sets those as required params, so Clowder would be in charge of defaulting them to `1`

config/crd/bases/cloud.redhat.com_clowdapps.yaml
bsquizz: Will re-add them

controllers/cloud.redhat.com/providers/kafka/appinterface.go
bsquizz: If we think it's something worth checking, check that partitions/replicas on the KafkaTopic match what the user is requesting in their ClowdApp. It appears right now we only ensure that a topic with matching name exists.

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
psav: Hmm I see - I know it's more for us to look into - but if we have now gone to the effort to get a full KafkaTopic spec, we could use a hook to validate and reject if it isn't set. - Thoughts?

psav: Or- I mean - we can just validate it in the reconciler

controllers/cloud.redhat.com/providers/kafka/appinterface.go
psav: This is outside the scope for clowder - I think we assume that the thing in app-interface is right - I get that it means we will potentially "ignore" settings in the clowdapp - but I think this is OK for app-interface mode

bsquizz: I'll drop the TODO

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
bsquizz: I just updated topics/partitions to be optional, we will default to 3 if not present.

So our ClowdApp has:
replicas optional
partitions optional
topicName required

Whereas the [strimzi spec](https://github.com/bsquizz/strimzi-client-go/blob/main/apis/kafka.strimzi.io/v1beta1/kafkatopic_types.go#L50-L58) has:
replicas required
partitions required
topicName optional

So that's why I have a separate type defined here -- to correctly define what is optional/required to set on a ClowdApp

bsquizz:
I'm reading that the kube recommendation is to use 'int32', how do we feel about the auto-generated types being `int`? :grimacing: I think I'll edit the Strimzi types and convert int to int32.

bsquizz:
May want to start using something like this too: https://github.com/rung/go-safecast

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -5,13 +5,14 @@ import (
 	"strconv"
 
 	crd "cloud.redhat.com/clowder/v2/apis/cloud.redhat.com/v1alpha1"
-	strimzi "cloud.redhat.com/clowder/v2/apis/kafka.strimzi.io/v1beta1"
+	strimzi "github.com/bsquizz/strimzi-client-go/apis/kafka.strimzi.io/v1beta1"
```
kylape, line 8: can we have this moved to the redhatinsights org?

build/pr_check_inner.sh
```diff
@@ -73,4 +73,5 @@ $KUBEBUILDER_ASSETS/kubectl apply -f build/skuttl-namespace.yaml
 $KUBEBUILDER_ASSETS/kubectl apply -f build/skuttl-perms.yaml
 
 IMG=$IMAGE_NAME:$IMAGE_TAG make deploy
-$KUBEBUILDER_ASSETS/kubectl kuttl test --config bundle/tests/scorecard/kuttl/kuttl-test.yaml --crd-dir config/crd/bases/ bundle/tests/scorecard/kuttl/
\ No newline at end of file
+
+source build/run_kuttl.sh
```
kylape, line 77: is there a reason we `source` here and `bash` in other places?

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -188,37 +190,44 @@ func (s *strimziProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error
...
 				return errors.New(fmt.Sprintf("could not compute max for %v", replicaValList))
 			}
-			maxReplicaString, err := strconv.Atoi(maxReplicas)
+			maxReplicasInt, err := safecast.Atoi32(maxReplicas)
 			if err != nil {
-				return errors.New(fmt.Sprintf("could not convert to string %v", maxReplicas))
+				return errors.New(fmt.Sprintf("could not convert string to int32 for %v", maxReplicas))
+			}
+			k.Spec.Replicas = maxReplicasInt
+			if k.Spec.Replicas < int32(1) {
```
kylape, line 209: If someone sets replicas to `-1`, this would set it to `3`?  If that's the case, I might prefer to return an error instead and only default to `3` when `Replicas` is `0`.

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -188,37 +190,44 @@ func (s *strimziProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error
...
 				return errors.New(fmt.Sprintf("could not compute max for %v", partitionValList))
 			}
-			maxPartitionString, err := strconv.Atoi(maxPartitions)
+			maxPartitionsInt, err := safecast.Atoi32(maxPartitions)
 			if err != nil {
-				return errors.New(fmt.Sprintf("could not convert to string %v", maxPartitions))
+				return errors.New(fmt.Sprintf("could not convert to string to int32 for %v", maxPartitions))
+			}
+			k.Spec.Partitions = maxPartitionsInt
+			if k.Spec.Partitions < int32(1) {
```
kylape, line 225: Same comment regarding negative values.

kylape (changes requested):

build/pr_check_inner.sh
bsquizz: I did this because PATH is set a few lines above and I wanted to preserve that

controllers/cloud.redhat.com/providers/kafka/strimzi.go
bsquizz: How about using kubebuilder minimum/maximum annotations on our Type definition to prevent them from setting "-1" on a ClowdApp's config?

kylape: Yes, 💯

controllers/cloud.redhat.com/providers/kafka/strimzi.go
bsquizz: Done

controllers/cloud.redhat.com/providers/kafka/strimzi.go
bsquizz: Done

kylape (approved):

kylape:
Took a look at the go-safecast project, and its lack of maturity gives me concern.  Perhaps we should consider using a different type casting library?

kylape (changes requested):
Took a look at the go-safecast project, and its lack of maturity gives me concern. Perhaps we should consider using a different type casting library?

kylape (approved):

PR #207, opened 2021-01-29
Fix strimzi complexity

psav:


BlakeHolifield (approved):
Refactor looks great!

PR #210, opened 2021-02-04
Allow any real provider to be omitted

psav:


bsquizz:
Hey @psav  -- we'll also need to set `omitempty` on the rest of the provider types here, no? 

https://github.com/RedHatInsights/clowder/blob/master/apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go#L236

PR #214, opened 2021-02-05
Create Kafka and KafkaConnect clusters when using strimzi provider

bsquizz:
The `providers.kafka` section of a `ClowdEnvironment` now has been reworked with this format:
```
kafka:
  mode: operator
  pvc: true
  cluster:
    namespace: kafka-ns
    name: my-kafka-cluster
    replicas: 1
  connect:
    namespace: connect-ns
    name: my-kafka-connect-cluster
    replicas: 1
```

* When in operator mode, Clowder now takes care of dropping a Strimzi `Kafka` and `KafkaConnect` resource into the namespaces defined in your kafka provider config. In `operator` mode, other options can be tweaked to control behavior of the provisioner as well... see the `KafkaClusterConfig` and `KafkaConnectClusterConfig` ClowdEnvironment types.
* In app-interface mode, only the `name` and `namespace` are used.
* The `connect` and `cluster` sections are entirely ignored in `local` or `none` mode.


bsquizz:
[test]

build/e2e-test.sh
```diff
@@ -3,9 +3,7 @@ set -e
 
 export IMAGE_TAG=`git rev-parse --short HEAD`
 
-kubectl apply -f build/skuttl-namespace.yaml
-kubectl apply -f build/skuttl-perms.yaml
-kubectl apply -f build/prommie-operator-bundle.yaml
+bash build/setup_minikube.sh
```
psav, line 6: This script is incorrect - it should be `minikube_setup.sh`

build/pr_check_inner.sh
```diff
@@ -62,16 +60,41 @@ export PATH="$KUBEBUILDER_ASSETS:$PATH"
...
+mv kuttl-test.xml artifacts/junit-kuttl.xml
+
+CLOWDER_PODS=$(kubectl get pod -n clowder-system -o jsonpath='{.items[*].metadata.name}')
+for pod in $CLOWDER_PODS; do
+    kubectl logs $pod -n clowder-system > artifacts/$pod.log
+done
+
+STRIMZI_PODS=$(kubectl get pod -n strimzi -o jsonpath='{.items[*].metadata.name}')
+for pod in $STRIMZI_PODS; do
+    kubectl logs $pod -n clowder-system > artifacts/$pod.log
```
psav, line 90: The namespace here is wrong - should be `strimzi`

bundle/tests/scorecard/kuttl/test-cyndi-strimzi/00-install.yaml
```diff
@@ -13,3 +13,28 @@ metadata:
...
+        tls: false
+    storage:
+      type: ephemeral
+  zookeeper:
+    replicas: 1
+    storage:
+      type: ephemeral
+  entityOperator:
+    topicOperator: {}
+    userOperator: {}
```
psav, line 40: New lines at end of file please :)

bundle/tests/scorecard/kuttl/test-kafka-strimzi-kafka-provisioner/00-install.yaml
```diff
@@ -2,15 +2,15 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: test-kafka-strimzi
+  name: test-kafka-strimzi-cluster-provisioner-kafka
```
psav, line 5: The name of the test and the name of the _root_ namespace do not match.

bundle/tests/scorecard/kuttl/test-kafka-strimzi-topic/00-install.yaml
```diff
@@ -0,0 +1,41 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: test-kafka-strimzi
```
psav, line 5: The namespace here should match the test-name, or visa versa

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -27,27 +28,191 @@ type strimziProvider struct {
...
+			Version:  &version,
+			Replicas: replicas,
+			Listeners: []strimzi.KafkaSpecKafkaListenersElem{
+				strimzi.KafkaSpecKafkaListenersElem{Name: "tcp", Type: "internal", Tls: false, Port: 9092},
+				strimzi.KafkaSpecKafkaListenersElem{Name: "tls", Type: "internal", Tls: true, Port: 9093},
+			},
+			Storage: strimzi.KafkaSpecKafkaStorage{
+				Type:        strimzi.KafkaSpecKafkaStorageTypePersistentClaim,
+				Size:        &storageSize,
+				DeleteClaim: &deleteClaim,
```
psav, line 85: We already have the `pvc` option for the providers, can we piggy back off this? Is it likely that you'll want zookeeper and kafka to have different PVC policies?

psav (changes requested):
@BlakeHolifield and @psav took a look through - some changes need making here - there is also a bigger question about the API change - but that will have to wait till tomorrow.

build/minikube_setup.sh
```diff
@@ -16,41 +16,61 @@ if ! command -v yq; then
...
 
 echo "*** Installing Strimzi resources ..."
 kubectl apply -f . -n $OPERATOR_NS
+cd -
+
+PROM_VERSION=0.45.0
+PROM_TARFILE="prometheus-operator-${PROM_VERSION}.tar.gz"
+if ! test -f ${PROM_TARFILE}; then
+    echo "*** Downloading ${PROM_TARFILE} ..."
+    curl -Ls -o ${PROM_TARFILE} https://github.com/prometheus-operator/prometheus-operator/archive/v${PROM_VERSION}.tar.gz
```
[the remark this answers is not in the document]
bsquizz: Are you ok if we download to a consistent folder name that is in `.gitignore`? The goal was to not have to re-download the .tar.gz on my local box every time minikube_setup.sh is run.

bsquizz:
Thanks for the review @psav @BlakeHolifield  -- see https://github.com/RedHatInsights/clowder/pull/214/commits/6261743ad18068cf2b48a122a795feea3632d7f9 for all the incorporated changes. Also added one new commit to tweak minikube_setup.sh to only install operators if they are not present on the cluster already.

pr_check.sh
```diff
@@ -28,4 +28,28 @@ export KUBEBUILDER_ASSETS=$PWD/kubebuilder_2.3.1_linux_amd64/bin
 IMG=$IMAGE_NAME:$IMAGE_TAG make docker-build
 IMG=$IMAGE_NAME:$IMAGE_TAG make docker-push
 
-docker run -i -v $PWD:/srcroot -e IMAGE_NAME=$IMAGE_NAME -e IMAGE_TAG=$IMAGE_TAG -e QUAY_USER=$QUAY_USER -e QUAY_TOKEN=$QUAY_TOKEN -e MINIKUBE_HOST=$MINIKUBE_HOST -e MINIKUBE_ROOTDIR=$MINIKUBE_ROOTDIR -e MINIKUBE_USER=$MINIKUBE_USER quay.io/psav/clowder_pr_check:blarg24 /srcroot/build/pr_check_inner.sh
+CONTAINER_NAME="clowder-pr-check-$ghprbPullId"
```
BlakeHolifield, line 31: `ghprbPullId` is a Jenkins var for the PR number, correct? So this would be `clowder-pr-check-214` ?

go.mod
```diff
@@ -3,7 +3,7 @@ module cloud.redhat.com/clowder/v2
 go 1.13
 
 require (
-	github.com/RedHatInsights/strimzi-client-go v0.21.1-2
+	github.com/RedHatInsights/strimzi-client-go v0.21.1-3
```
BlakeHolifield, line 6: Do we need this minor update or did this sneak. Go has some sneaky libraries :upside_down_face:

go.sum
```diff
@@ -33,8 +33,8 @@ github.com/PuerkitoBio/purell v1.1.0/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbt
 github.com/PuerkitoBio/purell v1.1.1/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbtSwDGJws/X0=
 github.com/PuerkitoBio/urlesc v0.0.0-20160726150825-5bd2802263f2/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
-github.com/RedHatInsights/strimzi-client-go v0.21.1-2 h1:2FV/RhXGbIqHzgWqTSmlw5jjSnTzR37Zi/yBX7KrYZg=
-github.com/RedHatInsights/strimzi-client-go v0.21.1-2/go.mod h1:DQ0Lv64xIvOwDNd+WQF7jWiF7G4pci2pKZsgDFk0k78=
+github.com/RedHatInsights/strimzi-client-go v0.21.1-3 h1:9McbENXUSe+r1bfFxjdsyAaJghwHvoZMYn1vn7HHm3g=
```
BlakeHolifield, line 36: Same question about sneaky stuff as above

build/minikube_setup.sh
```diff
@@ -15,42 +15,90 @@ if ! command -v yq; then
...
+fi
 
 echo "*** Waiting for Strimzi operator to come up ..."
-kubectl rollout status deployment/strimzi-cluster-operator -n $OPERATOR_NS
+kubectl rollout status deployment/strimzi-cluster-operator -n $STRIMZI_OPERATOR_NS
+
+echo "*** Waiting for Prometheus operator to come up ..."
+kubectl rollout status deployment/prometheus-operator -n $PROM_OPERATOR_NS
+
+cd -
```
BlakeHolifield, line 104: it _looks_ like this is jumping you back to the prommie bundle instead of the root.

BlakeHolifield (commented):
Bit more feedback. Once the PR check is happy I think we're good to go. We can see if Kyle/Pete want to discuss the API stuff first or merge and then refactor after.

pr_check.sh
bsquizz: Correct- they use the github pull request builder which provides this env var.

If you were to run this script locally, your container name would just end up being `clowder-pr-check-` which isn't great, but not terrible either.

go.mod
bsquizz: Needed for KafkaConnect

I am keeping the version name of his library (0.21.1) to match the version of Strimzi operator. I am just bumping the `-N` if I make changes to the package if its still based on Strimzi v0.21.1

build/minikube_setup.sh
bsquizz: Yeah... ok, changed `cd $DOWNLOAD_DIR` and `cd -` to be only called in the `install` functions now so all the dir changes are "contained" within the function calls.

pr_check.sh
BlakeHolifield: Okay gotcha. This was mostly for my education on what that was doing. Ty!

build/minikube_setup.sh
bsquizz: annnnd bit by another 'cd' issue... fixed again. Just going to call `pwd` at the start of the script and keep cd'ing back to that dir at the end of those functions.

build/minikube_setup.sh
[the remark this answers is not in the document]
kylape: sounds reasonable to me.  maybe have it be a dot folder

bsquizz:
It's in `build/downloads` now -- which was added to .gitignore

build/minikube_setup.sh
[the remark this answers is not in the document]
bsquizz: I opted for `build/downloads` -- which was added to .gitignore -- chose not to use a dot folder so it's not too sneaky.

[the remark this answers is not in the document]
bsquizz: Correction -- `build/operator_bundles`

build/minikube_setup.sh
```diff
@@ -15,42 +15,91 @@ if ! command -v yq; then
...
+fi
+
+if [ -z "$PROM_DEPLOYMENT" ]; then
+    echo "*** Prometheus operator deployment not found, installing ..."
+    install_prometheus_operator
+else
+    echo "*** Prometheus operator deployment found, skipping install ..."
+fi
 
 echo "*** Waiting for Strimzi operator to come up ..."
```
kylape, line 100: could you background the `rollout status` commands in the respective `if` block and then call `wait` at the bottom instead?

build/minikube_setup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
```
kylape, line 1: the name of this script is fine :smile:

kylape (changes requested):
I think the big question in my mind regarding this PR is: Do we want to continue to support using already-existing Kafka clusters in operator mode?  If not, then we can simplify a lot of the spec and implementation. If yes, then I might consider creating a two separate modes: operator and operator-no-cluster (or something like that).  This way we can leverage the mode option to simplify the other options in the spec.

build/minikube_setup.sh
bsquizz: :tada:

bsquizz:
> Do we want to continue to support using already-existing Kafka clusters in operator mode?

I think that mainly depends on if we want to eventually have Clowder creating KafkaTopic resources in stage and prod or if we intend to always control that via app-interface.

I'm not that satisfied with the way the spec looks either but I didn't want to break backward compatibility

Right now if you want "operator mode with no cluster provisioning" you have:

```
kafka:
  mode: operator
  namespace: kafka-ns
  clusterName: my-kafka-cluster
  connectNamespace: connect-ns
  connectClusterName: my-kafka-connect-cluster
```

If you want "operator mode with cluster provisioning" you have:
```
kafka:
  mode: operator
  namespace: kafka-ns
  clusterName: my-kafka-cluster
  connectNamespace: connect-ns
  connectClusterName: my-kafka-connect-cluster
  pvc: true
  kafkaClusterOptions:
    provisionCluster: true
    replicas: 1
  kafkaConnectClusterOptions:
    provisionCluster: true
    replicas: 1
```

The ideal "look and feel" of the spec, IMO, would be something like:
```
kafka:
  mode: operator
  kafka:
    namespace: kafka-ns
    name: my-kafka-cluster
    clusterProvisioner:
      enabled: true
      pvc: true
      replicas: 1
  kafkaConnect:
    namespace: connect-ns
    name: my-kafka-connect-cluster
    clusterProvisioner:
      enabled: true
      pvc: true
      replicas: 1
```

I think that spec is usable by all 3 modes (operator, local, and app-interface). We'd just leave out the "clusterProvisioner" entirely when not interested in having Clowder spin up the cluster. But obviously that will be a breaking change for any existing ClowdEnvironments we have floating around out there and especially for stage/prod we'd have to make sure we don't blow anything up

bsquizz:
FYI we pulled the script enhancements out of this PR into https://github.com/RedHatInsights/clowder/pull/233

kylape:
I like the idea spec much more, of course, and I think we should try to move to that if possible.  Would it be reasonable to provide backwards compat while moving to the idea spec?

Using the ideal spec as a starting point, I have a few questions/suggestions.

1. Do we need to have individual configuration for both kafka and kafka connect?  e.g. would you ever want PVCs for kafka but not kafka connect?
2. Can we assume that kafka connect and kafka will always live in the same namespace?  I know they technically can live in a different namespace, but are there ever situations we encounter where they actually do live in separate namespaces?
3. If we are going to deal with a breaking spec change here, should we try to future-proof a bit and consider how the spec will change when we add managed kafka as a mode?

bsquizz:
OK, new spec implemented in https://github.com/RedHatInsights/clowder/pull/214/commits/29143e4a23c1a43db430de5b525bd8cd04df9f8e

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -72,43 +269,48 @@ func NewStrimzi(p *p.Provider) (providers.ClowderProvider, error) {
 		},
 	}
 
+	// NOTE: we will not create ownerReferences on generated strimzi objects, since it is possible
+	// for them to reside a namespace that does not match the ClowdEnvironment's 'targetNamespace'
+	//
+	// Therefore, if a ClowdEnvironment is removed, we do not remove created strimzi objects
```
kylape, line 275: How do they get removed if not deleted along with an env?

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -72,43 +269,48 @@ func NewStrimzi(p *p.Provider) (providers.ClowderProvider, error) {
 		},
 	}
 
+	// NOTE: we will not create ownerReferences on generated strimzi objects, since it is possible
+	// for them to reside a namespace that does not match the ClowdEnvironment's 'targetNamespace'
```
kylape, line 273: I didn't think this would be a reason to not set the owner ref

bsquizz: Set the owner ref as the ClowdEnvironment in https://github.com/RedHatInsights/clowder/pull/214/commits/ac9699b8ab0bc1459e395810731bf4e44d80a17f

bundle/tests/scorecard/kuttl/test-cyndi-strimzi/00-install.yaml
```diff
@@ -13,3 +13,28 @@ metadata:
 spec:
   finalizers:
   - kubernetes
+---
```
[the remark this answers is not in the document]
bsquizz: Good point. Yes, I should change these.

bundle/tests/scorecard/kuttl/test-kafka-strimzi-topic/00-install.yaml
```diff
@@ -0,0 +1,41 @@
...
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: test-kafka-strimzi-topic-kafka
+spec:
+  finalizers:
+  - kubernetes
+---
+# Install a kafka cluster so that the environment reconciler successfully finds 'listeners'
```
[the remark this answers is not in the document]
bsquizz: A relic of the past when operator mode had optional kafka provisioning..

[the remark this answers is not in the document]
bsquizz: (no longer needed)

psav (approved):

PR #215, opened 2021-02-08
Add appProtocol to service port definitions

kylape:
kiali/istioctl can report traffic as http as long as a port is named
"http" or starts with "http-".


psav:
Want to add a quick addition to the tests to test for the presence?

kylape:
Yes, that would be a good idea.

psav (approved):

PR #220, opened 2021-02-10
Rename minkube_setup.sh to strimzi_install.sh

kylape:
afiact this has nothing to do with minikube


psav:
@bsquizz do we need to update another script here?

bsquizz:
Technically this script is going to become the script that "prepares your local k8s cluster" -- whatever that might involve.

In #214 the script is now installing both strimzi and prometheus operator

bsquizz:
I'd close this and just address the rename in #214 -- because I have other scripts there calling this one now.

PR #225, opened 2021-02-11
Added feature list

psav:


kylape:
I suggest taking off “provisioning of” on the bullets that have it. I think the last three bullets are not really features, at least in the way that I think should go on this list.

PR #238, opened 2021-02-23
Added context deadline to Kafka topic create in local mode

psav:
* In local mode, Kafka create tries to create a topic. In the case
  that a kafka instance suffers an issue, this can sometimes block for
  a long period of time.
* Adding a deadline to the context, forces it to end early in this case.


psav:
Note, this is a consequence of not using PVC for kafka instances

bsquizz (approved):

kylape (approved):

PR #241, opened 2021-02-23
Job Invocation CRD

BlakeHolifield:
* Adds CRD for JobInvocation
* Allows an `oc apply` of CRD to trigger a job from the jobs list in the corresponding ClowdApp


apis/cloud.redhat.com/v1alpha1/jobinvocation_types.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// Jobs is the set of jobs to be run by the invocation
+	Jobs []string `json:"jobs"`
+}
+
+type JobResult struct {
+	Name      string `json:"name"`
+	Completed bool   `json:"completed"`
+	Attempts  int    `json:"attempts"`
+	StdOut    string `json:"stdout"`
```
psav, line 36: Is there a `StdErr` in this too? If so, perhaps it would be better to call it "output" - I didn't look further down yet - but we probably want to be careful how much data we put into this. I realise we will want to give people an output so they can debug - but we may need to think about this a little more.

apis/cloud.redhat.com/v1alpha1/jobinvocation_types.go
```diff
@@ -0,0 +1,67 @@
...
+	AppName string `json:"appName"`
+
+	// Jobs is the set of jobs to be run by the invocation
+	Jobs []string `json:"jobs"`
+}
+
+type JobResult struct {
+	Name      string `json:"name"`
+	Completed bool   `json:"completed"`
+	Attempts  int    `json:"attempts"`
```
psav, line 35: Do we have a limit on the number of attempts to make?

apis/cloud.redhat.com/v1alpha1/jobinvocation_types.go
```diff
@@ -0,0 +1,67 @@
...
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// JobInvocationSpec defines the desired state of JobInvocation
+type JobInvocationSpec struct {
```
psav, line 24: Now that I think about it - sorry @BlakeHolifield - I wonder if we should call this a `ClowdJobInvocation` to align it with the others?

config/samples/cloud.redhat.com_v1alpha1_clowdenvironment.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: cloud.redhat.com.cloud.redhat.com/v1alpha1
+apiVersion: cloud.redhat.com/v1alpha1
```
psav, line 1: We should do this for `clowdapp` file too if we are doing it elsewhere.

controllers/cloud.redhat.com/jobinvocation_controller.go
```diff
@@ -0,0 +1,240 @@
...
+// +kubebuilder:rbac:groups="",resources=configmaps;services;persistentvolumeclaims;secrets;events;namespaces,verbs=get;list;watch
+// +kubebuilder:rbac:groups=batch,resources=cronjobs,verbs=get;list;create;update;watch;patch;delete
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch
+
+func (r *JobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	qualifiedName := fmt.Sprintf("%s:%s", req.Namespace, req.Name)
+	log := r.Log.WithValues("jobinvocation", qualifiedName)
+	ctx := context.WithValue(context.Background(), errors.ClowdKey("log"), &log)
+	ctx = context.WithValue(ctx, errors.ClowdKey("recorder"), &r.Recorder)
+	//proxyClient := ProxyClient{Ctx: ctx, Client: r.Client}
```
psav, line 60: We "probably" don't need the proxyClient here as it's likely we are not going to be creating any other resources than jobs, and jobs won't be removed ever, at least - not by Clowder. I don't think.

controllers/cloud.redhat.com/jobinvocation_controller.go
```diff
@@ -0,0 +1,240 @@
...
+	// Get the ClowdApp
+	app := crd.ClowdApp{}
+	err = r.Client.Get(ctx, types.NamespacedName{
+		Name:      jinv.Spec.AppName,
+		Namespace: req.Namespace,
+	}, &app)
+
+	// Determine if the ClowdApp containing the Job exists
+	if err != nil {
+		r.Recorder.Eventf(&jinv, "Warning", "ClowdAppMissing", "ClowdApp [%s] is missing; Job cannot be jinved", jinv.Spec.AppName)
```
psav, line 85: Perhaps `jinved` should be replaced `invoked`

controllers/cloud.redhat.com/jobinvocation_controller.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (r *JobInvocationReconciler) InvokeJob(job crd.Job, app *crd.ClowdApp, env *crd.ClowdEnvironment, ctx context.Context) error {
+	now := time.Now()
+	nn := types.NamespacedName{
+		Name:      fmt.Sprintf("%v-%v-%v", app.Name, job.Name, now.Unix()),
+		Namespace: app.Namespace,
+	}
+	if !job.OnRequest {
+		r.Recorder.Eventf(app, "Warning", "Only OnRequest Jobs are supported at this time", "%s is not set to onRequest", job)
```
psav, line 126: OK a few things here. We should instruct the user on what they should change; saying that the app name is not "set to", can be ambiguous, perhaps instead we should say that the onrequest flag is False. (Though see my note above on the enum, perhaps then it's even easier.) Secondly, let's make the Info state the same error message.

controllers/cloud.redhat.com/jobinvocation_controller.go
```diff
@@ -0,0 +1,240 @@
...
+		r.Log.Info("Unsupported Job Type", "jobinvocation", job.Name, "namespace", app.Namespace)
+	} else {
+		j := batchv1.Job{}
+		createJob(app, env, nn, job, &j)
```
psav, line 130: If this is successful, perhaps we should emit an event that the job was successfully invoked?

controllers/cloud.redhat.com/jobinvocation_controller.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func createJob(app *crd.ClowdApp, env *crd.ClowdEnvironment, nn types.NamespacedName, job crd.Job, j *batchv1.Job) error {
+	labels := app.GetLabels()
+	labels["pod"] = nn.Name
+	app.SetObjectMeta(j, crd.Name(nn.Name), crd.Labels(labels))
+
+	pod := job.PodSpec
+
+	j.ObjectMeta.Labels = labels
```
psav, line 145: Did my previous comment make sense where we can perhaps reuse this same logic from the previous CronJob, code. Though the two specs for `Job` and `CronJob` are different, I think the spec for the Job is the same, just at a nested level in `Job`.

controllers/cloud.redhat.com/jobinvocation_controller.go
```diff
@@ -0,0 +1,240 @@
...
+	maker.ApplyPodAntiAffinity(&j.Spec.Template)
+
+	annotations := make(map[string]string)
+	//annotations["configHash"] = hash
+	j.Spec.Template.SetAnnotations(annotations)
+
+	return nil
+}
+
+func matchAndReturnJob(jobName string, app *crd.ClowdApp) (crd.Job, error) {
```
psav, line 225: Let's go with the more standard way of using k8s resources in "getters", don't return the object itself, as that will likely result in a copy, let's create a blank object in the calling function, pass the pointer to it, and then we only need to return the error which can be nil.

controllers/cloud.redhat.com/jobinvocation_controller.go
```diff
@@ -0,0 +1,240 @@
...
+func createJob(app *crd.ClowdApp, env *crd.ClowdEnvironment, nn types.NamespacedName, job crd.Job, j *batchv1.Job) error {
+	labels := app.GetLabels()
+	labels["pod"] = nn.Name
+	app.SetObjectMeta(j, crd.Name(nn.Name), crd.Labels(labels))
+
+	pod := job.PodSpec
+
+	j.ObjectMeta.Labels = labels
+	j.Spec.Template.ObjectMeta.Labels = labels
+
```
psav, line 147: Concurrency Policy?

controllers/cloud.redhat.com/makers/maker.go
```diff
@@ -323,14 +313,69 @@ func initJob(app *crd.ClowdApp, env *crd.ClowdEnvironment, cj *batch.CronJob, nn
...
+
+	ApplyPodAntiAffinity(&cj.Spec.JobTemplate.Spec.Template)
 
 	annotations := make(map[string]string)
 	annotations["configHash"] = hash
 	cj.Spec.JobTemplate.Spec.Template.SetAnnotations(annotations)
 }
 
-func processResources(pod *crd.PodSpec, env *crd.ClowdEnvironment) core.ResourceRequirements {
+func applyOneShotJob(app *crd.ClowdApp, env *crd.ClowdEnvironment, j *batchv1.Job, pt *core.PodTemplateSpec, nn types.NamespacedName, job crd.Job, hash string) {
```
psav, line 356: This function is currently unused?

docs/api_reference.rst
```diff
@@ -365,6 +365,7 @@ Appears In:
    "``restartPolicy`` (`RestartPolicy <https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#restartpolicy-v1-core>`_)", "Defines the restart policy for the CronJob, defaults to never"
    "``concurrencyPolicy`` (`ConcurrencyPolicy <https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#concurrencypolicy-v1beta1-batch>`_)", "Defines the concurrency policy for the CronJob, defaults to Allow"
    "``startingDeadlineSeconds`` (integer)", "Defines the StartingDeadlineSeconds for the CronJob"
+=======
```
psav, line 368: I feel there are some docs missing here - like for the entire JobInvocation piece....

psav (changes requested):
A few comments to look through :)

apis/cloud.redhat.com/v1alpha1/jobinvocation_types.go
BlakeHolifield: This was some noodling that I dropped in so we could discuss what we needed in here. I agree that it's currently an overload

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -68,7 +68,13 @@ type Job struct {
 	Name string `json:"name"`
 
 	// Defines the schedule for the job to run
-	Schedule string `json:"schedule"`
+	Schedule string `json:"schedule,omitempty"`
+
+	// A job that is a OneShot is run on apply
+	OneShot bool `json:"oneshot,omitempty"`
```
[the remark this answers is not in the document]
BlakeHolifield: >  if we are never going to have both oneshot and onrequest set together,
Hmm. I'm not sure this will be the case. Are there instances where a job that runs on deploy would want to be run again in the future?

controllers/cloud.redhat.com/jobinvocation_controller.go
BlakeHolifield: `%s/jinved/oops`

controllers/cloud.redhat.com/jobinvocation_controller.go
BlakeHolifield: My understand of concurrency policy was that it is only available for CronJobs. The k8s controller for cronjobs will use concurrency to determine if multiple instances of the same cronjob can run at the same time. Here, we are just triggering them as a normal job regardless of schedule.

psav:
This will need rebasing before tests will pass

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -0,0 +1,330 @@
...
+
+	case "request":
+		createJobResource(app, env, nn, job, &j)
+		if err := r.Client.Create(ctx, &j); err != nil {
+			return err
+		}
+		jinv.Status.PodNames = append(jinv.Status.PodNames, j.ObjectMeta.Labels["pod"])
+		r.Log.Info("Job Invoked Successfully", "jobinvocation", job.Name, "namespace", app.Namespace)
+
+	case "deploy":
```
BlakeHolifield, line 178: I think there's a conversation to be had here about types and if "deploy" jobs are valid usecases. Is a "deploy" job really just an initContainer?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	return errors.New(fmt.Sprintf("No such job %s", jobName))
+
+}
+
+// SetupWithManager registers the CJI with the main manager process
+func (r *ClowdJobInvocationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.Recorder = mgr.GetEventRecorderFor("jobinvocation")
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&crd.ClowdJobInvocation{}).
```
BlakeHolifield, line 298: Needs a watcher here to avoid redundant requeues

BlakeHolifield: Done!

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
kylape: I think you could argue that there are not a lot of use cases for deploy jobs, but they have several key differences from init containers:

- They have their own pod template
- They can run parallel with deployment pods
- They can be re-invoked via JobInvocation

So I don't think it hurts to have them, even if they aren't used very often.

psav: Here's my counter to it,
1) The ClowdApp is supposed to describe the app + operational maintenance jobs etc, I personally think that having a job that is invoked once and never ever again, essentially once per environment seems odd, especially if that job definition now just sits there and is never reinvoked.
2) How do we track that it is done and will never be done again? If the Job did get deleted, what would prevent a retriggering, some kind of annotation?
3) Your point about them being re-invoked makes me think, why not just make it on request and put in a job invocation when you deploy that app. There is the ability to "wait" until an app is fully deployed before running in a oneshot, and so we could just extend this functionality as a flag on the spec saying, don't run this job until the app is ready.

bundle/tests/scorecard/kuttl/test-clowder-jobs/02-json-asserts.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: kuttl.dev/v1alpha1
+kind: TestStep
+commands:
+- script: sleep 90
```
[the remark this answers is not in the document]
kylape: ah I see, gotta let the job run to completion

apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// GetLabels returns a base set of labels relating to the ClowdJobInvocation.
+func (i *ClowdJobInvocation) GetLabels() map[string]string {
+	if i.Labels == nil {
+		i.Labels = map[string]string{}
+	}
+
+	if _, ok := i.Labels["cji"]; !ok {
+		i.Labels["cji"] = i.ObjectMeta.Name
```
psav, line 78: we are soon going to move to a `clowd` prefix on the other labels - think we should do that here too before people start using it

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/cloudservices/clowder
-  newTag: e3dd531
+  newName: 127.0.0.1:5000/clowder
```
psav, line 7: Please revert

config/rbac/role.yaml
```diff
@@ -24,6 +24,23 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
```
psav, line 29: I wonder if this is a rebase gone wonky - when we added service accounts, this is a duplicate section afaict

config/samples/cloud.redhat.com_v1alpha1_clowdjobinvocation.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: cloud.redhat.com.cloud.redhat.com/v1alpha1
+kind: ClowdJobInvocation
+metadata:
+  name: clowdjobinvocation-sample
+spec:
```
psav, line 5: Can we give an example here - we do have them for app/env

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -64,7 +64,8 @@ type ClowdAppReconciler struct {
 // +kubebuilder:rbac:groups=cloud.redhat.com,resources=clowdapps,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=cloud.redhat.com,resources=clowdapps/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups="",resources=serviceaccounts;configmaps;services;persistentvolumeclaims;secrets;events;namespaces,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=batch,resources=cronjobs,verbs=get;list;create;update;watch;patch;delete
+// +kubebuilder:rbac:groups="",resources=configmaps;services;persistentvolumeclaims;secrets;events;namespaces,verbs=get;list;watch;create;update;patch;delete
```
psav, line 67: Wonky rebase again

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -178,6 +179,12 @@ func (r *ClowdAppReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	}
 
 	SetDeploymentStatus(ctx, &proxyClient, &app)
+
+	app.Status.Ready = false
+	if app.Status.Deployments.ManagedDeployments == app.Status.Deployments.ReadyDeployments {
```
psav, line 184: This "functionality" was recently moved into a function on the type for the env, care to just quickly do the same here, 5min job

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -0,0 +1,374 @@
...
+			return ctrl.Result{}, nil
+		}
+		r.Log.Error(err, "CJI not found ")
+		return ctrl.Result{}, err
+	}
+
+	// Set the initial status to an empty list of pods and a Completed
+	// status of false. If a job has been invoked, but hasn't finished,
+	// setting the status after requeue will ensure it won't be double invoked
+	r.setCompletedStatus(ctx, &cji)
```
psav, line 81: `r.setCompletedStatus` can return an error, but we don't handle it

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -0,0 +1,374 @@
...
+		fullJobName := fmt.Sprintf("%v-%v-%v", app.Name, job.Name, cji.Name)
+		if contains(cji.Status.Jobs, fullJobName) {
+			continue
+		}
+
+		// We have a match that isn't running and can invoke the job
+		r.Log.Info("Invoking job", "jobinvocation", jobName, "namespace", app.Namespace)
+
+		if err := r.InvokeJob(ctx, &job, &app, &env, &cji); err != nil {
+			r.Log.Info("Job Invocation Failed", "jobinvocation", jobName, "namespace", app.Namespace)
```
psav, line 153: I wonder if we should emit a few events here too when a job is started/invoked?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -0,0 +1,374 @@
...
+			return ctrl.Result{Requeue: true}, err
+		}
+	}
+
+	// Short running jobs may be done by the time the loop is ranged,
+	// so we update again before the reconcile ends
+	r.setCompletedStatus(ctx, &cji)
```
psav, line 160: The error from this function isn't handled

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -0,0 +1,374 @@
...
+// the status of that job
+func (r *ClowdJobInvocationReconciler) InvokeJob(ctx context.Context, job *crd.Job, app *crd.ClowdApp, env *crd.ClowdEnvironment, cji *crd.ClowdJobInvocation) error {
+	nn := types.NamespacedName{
+		Name:      fmt.Sprintf("%v-%v-%v", app.Name, job.Name, cji.Name),
+		Namespace: app.Namespace,
+	}
+	j := batchv1.Job{}
+
+	createJobResource(cji, env, nn, job, &j)
+	if err := r.Client.Create(ctx, &j); err != nil {
```
psav, line 179: `ctx` doesn't need to be passed as `r` already has it

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -0,0 +1,374 @@
...
+// InvokeJob is responsible for applying the Job. It also updates and reports
+// the status of that job
+func (r *ClowdJobInvocationReconciler) InvokeJob(ctx context.Context, job *crd.Job, app *crd.ClowdApp, env *crd.ClowdEnvironment, cji *crd.ClowdJobInvocation) error {
+	nn := types.NamespacedName{
+		Name:      fmt.Sprintf("%v-%v-%v", app.Name, job.Name, cji.Name),
+		Namespace: app.Namespace,
+	}
+	j := batchv1.Job{}
+
+	createJobResource(cji, env, nn, job, &j)
```
psav, line 178: `createJobResource` returns an error that isn't handled

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+	cji.Status.Jobs = append(cji.Status.Jobs, j.ObjectMeta.Name)
+	r.Log.Info("Job Invoked Successfully", "jobinvocation", job.Name, "namespace", app.Namespace)
+
+	return nil
+}
+
+// applyJob build the k8s job resource and applies it from the Job config
+// defined in the ClowdApp
+func createJobResource(cji *crd.ClowdJobInvocation, env *crd.ClowdEnvironment, nn types.NamespacedName, job *crd.Job, j *batchv1.Job) error {
```
psav, line 190: It looks like we don't need to return an error here at all

psav (changes requested):

psav (approved):

PR #245, opened 2021-02-24
add `minikube config` settings to readme

lindgrenj6:
Just adding some docs around persisting the flags that way they don't need to be remembered/referred. Can easily be tweaked as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,14 @@ initiated with the following options:
   minikube start --cpus 4 --disk-size 36GB --memory 8000MB
   ```
 
+  To persist these changes for every minikube invocation, run the following:
+  ```shell
+  minikube config set cpus 4
+  minikube config set memory 8000
+  minikube config set disk-size 36GB
+  minikube config set vm-driver kvm2
+  ```
+
 Clowder can then be installed by visiting the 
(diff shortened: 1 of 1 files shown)


bsquizz:
Any idea if you can persist `--addons=registry` via the minikube config as well? That's the only other thing I'd suggest adding.

lindgrenj6:
@bsquizz I haven't found a way to do that unfortunately, it's not possible to enable certain addons by default with `minikube config` yet. There are a few WIP PR's and issues adding the functionality though it looks like.

One option for that would be to run a registry somewhere, then config minikube with `minikube config set insecure-registry ip:port`, that basically configured the docker daemon to use an insecure registry. But that involves running the registry elsewhere which doesn't gain us much.

bsquizz:
Yeah no worries, was just curious. I agree, the separate registry isn't worth it. `minikube --addons=registry` has been working well for me... and I don't have to ever check if some other VM/container/app is running so I like that ;)

lindgrenj6:
100% same, having everything integrated is super nice.

psav:
It's not passing I presume because of a lack of app-interface GPG? - plus it will need rebasing due to the kuttl update

PR #257, opened 2021-03-03
Add docs building to PR check

psav:


pr_check.sh
```diff
@@ -13,6 +13,18 @@ done <<< "$(git log --pretty=format:%s $(git merge-base master HEAD)..HEAD)"
...
+source .venv/bin/activate
+pip3 install sphinx-rtd-theme
+
+cd docs/
+make clean
+make html
+
+deactivate
+
+cd -
```
BlakeHolifield, line 26: These are great as one use, but I've seen em cause some mighty headaches in the past. Just a heads up if this is modified in the future or moved to a function.

BlakeHolifield (approved):

pr_check.sh
bsquizz: Maybe just keep it all on one line. `cd docs && make clean && make html && cd -`

then call `deactivate` on the next line?

psav: I like that

PR #258, opened 2021-03-04
Add SSL Mode

psav:


BlakeHolifield (approved):
Kuttl tests cover both modes so I think we're good here.

PR #259, opened 2021-03-05
Beef up minikube instance on pr_check

bsquizz:
Moved to using a new VM w/ 8cores/32gig mem -- we'll start minikube w/ 6 cores and 16gig mem for now.

### DIFF
--- a/build/pr_check_inner.sh
+++ b/build/pr_check_inner.sh
@@ -24,7 +24,7 @@ ls -la /dev/tty
 chmod 600 minikube-ssh-ident
 
 ssh -o StrictHostKeyChecking=no $MINIKUBE_USER@$MINIKUBE_HOST -i minikube-ssh-ident "minikube delete"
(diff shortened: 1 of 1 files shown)


build/pr_check_inner.sh
```diff
@@ -24,7 +24,7 @@ ls -la /dev/tty
 chmod 600 minikube-ssh-ident
 
 ssh -o StrictHostKeyChecking=no $MINIKUBE_USER@$MINIKUBE_HOST -i minikube-ssh-ident "minikube delete"
-ssh -o StrictHostKeyChecking=no $MINIKUBE_USER@$MINIKUBE_HOST -i minikube-ssh-ident "minikube start"
+ssh -o StrictHostKeyChecking=no $MINIKUBE_USER@$MINIKUBE_HOST -i minikube-ssh-ident "minikube start --cpus 6 --disk-size 36GB --memory 16000MB"
```
kylape, line 27: I actually don't think the disk is 32G :grimacing: I believe I chose 20G

bsquizz: I happen to be awake :laughing: Downsized it to 10GB

PR #260, opened 2021-03-08
Add profiling capability

psav:


main.go
```diff
@@ -28,8 +31,10 @@ import (
 )
 
 var setupLog = ctrl.Log.WithName("setup")
+var cpuprofile = flag.String("cpuprofile", "", "write cpu profile to file")
```
BlakeHolifield, line 34: A nitpick, but do we want to camelCase this as `cpuProfile`?

BlakeHolifield (approved):
One tiny thought. Otherwise, neat!

main.go
psav: Not as an arg to command line.....

PR #266, opened 2021-03-10
Overwrite port number for liveness/readiness probes

kylape:
Devs currently have to manually copy the port number from the
environment spec to their liveness/readiness probes in their app spec.
This would automatically update the probes with the correct port number
given an app uses the HTTPGet handler.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: 127.0.0.1:5000/clowder
-  newTag: 95777a1
+  newTag: 7248ca5
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: 127.0.0.1:5000/clowder
-  newTag: 7248ca5
+  newTag: 95777a1
--- a/config/manifests/bases/clowder.clusterserviceversion.yaml
+++ b/config/manifests/bases/clowder.clusterserviceversion.yaml
@@ -4,7 +4,7 @@ metadata:
   annotations:
     alm-examples: '[]'
     capabilities: Basic Install
-    operators.operatorframework.io/builder: operator-sdk-v1.0.0
+    operators.operatorframework.io/builder: operator-sdk-v1.0.1
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v2
   name: clowder.vX.Y.Z
   namespace: placeholder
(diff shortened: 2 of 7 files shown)


psav:
This needs a little rethinking due to people using nonstandard ports for their checks, ie checks against metrics, should perhaps be gated behind a check for web port presence

kylape:
Refined this PR.  Please refer to commit message for details.

config/manager/kustomization.yaml
```diff
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: 127.0.0.1:5000/clowder
-  newTag: 7248ca5
+  newTag: 95777a1
```
psav, line 8: This needs removing

config/manifests/bases/clowder.clusterserviceversion.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   annotations:
     alm-examples: '[]'
     capabilities: Basic Install
-    operators.operatorframework.io/builder: operator-sdk-v1.0.1
+    operators.operatorframework.io/builder: operator-sdk-v1.0.0
```
psav, line 7: Same as above

controllers/cloud.redhat.com/makers/maker.go
```diff
@@ -554,9 +620,16 @@ func initDeployment(app *crd.ClowdApp, env *crd.ClowdEnvironment, d *apps.Deploy
...
-	if deployment.Web {
+	if deployment.WebServices.Private.Enabled {
 		c.Ports = append(c.Ports, core.ContainerPort{
-			Name:          "web",
+			Name:          "private",
+			ContainerPort: env.Spec.Providers.Web.Port,
+		})
+	}
+
+	if deployment.WebServices.Public.Enabled {
```
psav, line 630: We need an or in here to keep backwards compat with `.Web` until it's gone

psav (changes requested):

jameswnl:
is this addressed already? do we still keep the port hardcoded in clowder  app template?

bennyturns:
Yes, this is addressed in a different PR.

bennyturns:
Actually let me check with pete and kyle, I am not 100% sure now

BlakeHolifield:
Yeah @jameswnl this code is in master here: https://github.com/RedHatInsights/clowder/blob/master/controllers/cloud.redhat.com/providers/web/impl.go#L50

We currently don't set that value in the ephemeral or stage environments, so it will be default 10000. You can override and test on a local environment if you'd like.

PR #271, opened 2021-03-18
Changed pod label

psav:


controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -326,7 +327,7 @@ func (r *ClowdJobInvocationReconciler) cjiToEnqueueUponJobUpdate(a handler.MapOb
 	for _, cji := range cjiList.Items {
 		// job event triggered a reconcile, check our jobs and match
 		// to enable a requeue
-		if contains(cji.Status.Jobs, job.ObjectMeta.Labels["pod"]) {
+		if contains(cji.Status.Jobs, job.ObjectMeta.Labels["clowdapp-deployment"]) {
```
BlakeHolifield, line 330: I think this is a goof on my part. We are no longer using this label. It should just be `job.ObjectMeta.Name` now

PR #274, opened 2021-03-19
Mutating Webhook

psav:
* Currently there are several deprecated features of the spec these are dealt
  with by using certain shims in different places.
* This commit moves all of the 'shim' functionality to a mutating webhook
  so that specs are mutated on creation/edit and Clowder then only deals
  with the latest version of the three specs.
* In time, each shim deprecation will be removed from the webhook, but it is
  less costly to keep it than if it is polluting the codebase.
* kubesetup.sh has been modified to download and install the cert-manager
  required for the creation and security of the webhook endpoint
* Certain YAMLs have been tweaked to provide the mutating webhook endpoint
* All current mutations/shims have been removed from the controller itself.
* Tests have been added to assert the mutation.
* Some tests had a database image removed as it is no longer required.

### DIFF
--- a/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
+++ b/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
@@ -315,6 +315,7 @@ type ClowdAppStatus struct {
 // +kubebuilder:printcolumn:name="Managed",type="integer",JSONPath=".status.deployments.managedDeployments"
 // +kubebuilder:printcolumn:name="EnvName",type="string",JSONPath=".spec.envName"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:webhook:path=/mutate-app,mutating=true,failurePolicy=fail,groups=cloud.redhat.com,resources=clowdapps,verbs=create;update,versions=v1alpha1,name=clowdapp.cloud.redhat.com
 
 // ClowdApp is the Schema for the clowdapps API
 type ClowdApp struct {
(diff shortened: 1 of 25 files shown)


controllers/cloud.redhat.com/mutant.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (a *mutantAnnotatorApp) Handle(ctx context.Context, req admission.Request) admission.Response {
+	app := &crd.ClowdApp{}
+	err := a.decoder.Decode(req, app)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	if app.Spec.Pods != nil {
+		// TODO find out why events are not submitted
```
BlakeHolifield, line 28: Is this a stretch or a TODO before merging?

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -419,31 +421,11 @@ func (i *ClowdApp) GetClowdSAName() string {
...
-				ReadinessProbe: pod.ReadinessProbe,
-				Volumes:        pod.Volumes,
-				VolumeMounts:   pod.VolumeMounts,
-			},
-		}
-		deps = append(deps, dep)
-	}
-	i.Spec.Deployments = deps
-}
+// func (r *ClowdApp) SetupWebhookWithManager(mgr ctrl.Manager) error {
```
BlakeHolifield, line 424: Is this not required? It looks to be the boilerplate from the `kubebuilder webhook create`. Does this get covered by registering both webhooks in `run.go`?

BlakeHolifield (commented):
Few quick clarifications. Admission webhooks are really cool and this looks sweet.

controllers/cloud.redhat.com/mutant.go
psav: Stretch - it's not _super_ important but would be nice

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
psav: It seems like it does. I tried removing it and re-running - and everything worked out fine.

BlakeHolifield (approved):

kylape:
Everything looks good to go to me.  I tested it on my local minikube, and it seemed to work.

I remembered one last thing: We need a NetworkPolicy installed in OSD to allow the k8s API to make calls to our webhook service.  Should we rely on app-interface to install this, or should we have Clowder create it?  I think my preference is to have Clowder create it. wdyt?

psav:
I agree with you - I don't know the format of this NetworkPolicy - do you have an example of it?

kylape (approved):

PR #276, opened 2021-03-24
Added trigger debug

psav:
* Currently there is log message to determine which resource triggered a reconciliation.
* A log message is now emitted detailing which resource in k8s triggered the reconciliation.
* This commit also adds several new debugging tools to Clowder which are controlled by
  environment variables.
* A new struct was added and a configmap can be created called clowder-config in the same
  namespace as clowder.

```
    debugOptions:
      cache:
        create: true
        update: true
        apply: true
      trigger:
        diff: true
```

* For "update" reconciliations, where a resource was modified, a diff can be logged by setting
  debugOptions->trigger->diff to true
* When the resource cache CREATEs an item, a message containing the content of this resource can be
  shown by setting debugoptions->cache->create to true
* When the resource cache UPDATEs an item, a message containing a diff of the change can be
  shown by setting debugOptions->cache->update to true
* When the resource cache APPLYs an item, a message containing the content of this resource can be
  shown by setting debugOptions->cache->apply to true
* In all cases, if the resource is a kind Secret, the output will be hidden to protect sensitive
  information.

### DIFF
--- a/bundle/olm-bundle.yml
+++ b/bundle/olm-bundle.yml
@@ -9,6 +9,14 @@ parameters:
   value: quay.io/cloudservices/clowder-index
 - name: TARGET_NAMESPACE
   value: clowder
+- name: DEBUG_TRIGGERS
+  value: "false"
+- name: DEBUG_CACHE_CREATE
+  value: "false"
+- name: DEBUG_CACHE_UPDATE
+  value: "false"
+- name: DEBUG_CACHE_APPLY
+  value: "false"
 objects:
 - apiVersion: operators.coreos.com/v1alpha1
   kind: CatalogSource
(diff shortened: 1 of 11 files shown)


kylape:
The difflib library is a bit concerning since it's no longer maintained.

I don't know how to configure environment variables on the clowder deployment via OLM.  I propose we mount an optional configmap into the container and load it in a new config package.

kylape (changes requested):
The difflib library is a bit concerning since it's no longer maintained.

I don't know how to configure environment variables on the clowder deployment via OLM. I propose we mount an optional configmap into the container and load it in a new config package.

psav:
OK sure - that's not a problem to me, the difflib is no longer maintained, it is in use by kuttl too at the moment. I'm not sure what other options we have. There is a newer fork of the project I think. If you know of anything better, please let me know.

kylape (changes requested):
Update the deployment template to include the ConfigMap

psav:
> Update the deployment template to include the ConfigMap

I thought I had already done that ....

kylape:
I went ahead and updated the deployment template and added a basic unit test for the clowder_config package.

psav:
This looks great to me! - Thanks for the additions

kylape (approved):

PR #278, opened 2021-03-24
Change 25 to 25% for rollouts

bennyturns:
Having int 25 instead of string 25% caused rollouts to not wait for the new pods to fully init before killing the old pods.  Changing it to a string 25% will give us the desired behavior.


psav:
Screenshot from 2021-03-25 11-06-44
This is what appears in the deployment from current Clowder, at least on my local setup, do we have an instance where what lands in the deployment yaml is different?

psav (approved):

PR #281, opened 2021-03-25
Refactor updates for tech debt

psav:


controllers/cloud.redhat.com/providers/serviceaccount/provider.go
```diff
@@ -20,5 +20,5 @@ func GetServiceAccount(c *p.Provider) (p.ClowderProvider, error) {
 }
 
 func init() {
-	p.ProvidersRegistration.Register(GetServiceAccount, 1, ProvName)
+	p.ProvidersRegistration.Register(GetServiceAccount, 97, ProvName)
```
BlakeHolifield, line 23: What's the reason for registering SA's towards the end?

controllers/cloud.redhat.com/providers/deployment/impl.go
```diff
@@ -99,17 +97,13 @@ func initDeployment(app *crd.ClowdApp, env *crd.ClowdEnvironment, d *apps.Deploy
...
 
 	c := core.Container{
-		Name:         nn.Name,
-		Image:        pod.Image,
-		Command:      pod.Command,
-		Args:         pod.Args,
-		Env:          envvar,
-		Resources:    ProcessResources(&pod, env),
-		VolumeMounts: pod.VolumeMounts,
-		Ports: []core.ContainerPort{{
```
BlakeHolifield, line 109: This is now factored out into the metrics provider, correct?

BlakeHolifield (approved):
Just a few clarifications.

controllers/cloud.redhat.com/providers/serviceaccount/provider.go
psav: Only because we need to add it to most deployments, and so we want to wait until as many providers have created their resources as possible,

controllers/cloud.redhat.com/providers/deployment/impl.go
psav: Correct - you can see it there right?

psav: This was left over from the other half of the refactor I think

BlakeHolifield: After double-checking, it is indeed there.

PR #285, opened 2021-03-31
Adds sorting to the db secret search in app-interface mode

psav:
* Owing to the fact that the order of the result from the search is not
  deterministic, we now sort these to prevent the issue of two different
  secrets causing a pod to flap due to config changes.


kylape:
Shoudln't the sort be done after the error check for the list API call?

kylape (changes requested):
Shouldn't the sort be done after the error check for the list API call?

psav:
Fixed sort order + added a test

PR #286, opened 2021-03-31
Added user operator to strimzi

psav:


controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -297,18 +328,88 @@ func (s *strimziProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error
...
+
 	// update s.Config.Topics
-	if err := s.processTopics(app); err != nil {
+	if err := s.processTopics(app, ku); err != nil {
 		return err
 	}
 
+	s.Cache.Update(KafkaUser, ku)
+
+	for _, topic := range s.Config.Brokers {
```
kylape, line 361: I'm sorry but I cannot understand how iterating over `Brokers` gives you a set of `topic` objects.  This makes it impossible to understand what this block of code is doing.

kylape (changes requested):

controllers/cloud.redhat.com/providers/kafka/strimzi.go
psav: You are absolutely right - I can't understand how I did that...

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -296,18 +335,88 @@ func (s *strimziProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error
 		}
 	}
 
+	ku := &strimzi.KafkaUser{}
```
bsquizz, line 338: For readability, can we move this logic into a private func named `createKafkaUser` that returns a `*strimzi.KafkaUserSpec`?

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -296,18 +335,88 @@ func (s *strimziProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error
...
+		if *broker.Authtype == config.BrokerConfigAuthtypeSasl {
+			ku := &strimzi.KafkaUser{}
+			nn := types.NamespacedName{
+				Name:      getKafkaUsername(s.Env, app),
+				Namespace: getKafkaNamespace(s.Env),
+			}
+
+			err := s.Client.Get(s.Ctx, nn, ku)
+			if err != nil {
+				return err
```
bsquizz, line 381: Do you need to use `errors.Wrap` here to trigger requeue?

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -296,18 +335,88 @@ func (s *strimziProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error
...
+			secnn := types.NamespacedName{
+				Name:      *ku.Status.Secret,
+				Namespace: getKafkaNamespace(s.Env),
+			}
+
+			kafkaSecret := &core.Secret{}
+
+			err = s.Client.Get(s.Ctx, secnn, kafkaSecret)
+			if err != nil {
+				return err
```
bsquizz, line 402: Same question here re `errors.Wrap`

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -296,18 +335,88 @@ func (s *strimziProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error
...
+
 	// update s.Config.Topics
-	if err := s.processTopics(app); err != nil {
+	if err := s.processTopics(app, ku); err != nil {
 		return err
 	}
 
+	s.Cache.Update(KafkaUser, ku)
+
+	for _, broker := range s.Config.BrokersTLS {
```
bsquizz, line 368: For readability, can we move this logic into another func named something like `setBrokerCredentials` (I think that is the spirit of what's being done here?)

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -210,46 +227,68 @@ func (s *strimziProvider) configureKafkaConnectCluster() error {
...
-	// }
-
-	// if len(s.Config.Brokers) < 1 {
-	// 	return fmt.Errorf(
-	// 		"Kafka cluster '%s' in ns '%s' has no listeners", clusterNN.Name, clusterNN.Namespace,
-	// 	)
-	// }
+	for _, listener := range kafkaResource.Status.Listeners {
+		if listener.Type != nil && *listener.Type == "tls" {
+			authType := config.BrokerConfigAuthtypeSasl
```
bsquizz, line 260: Can we move this to a func called `buildTlsBrokerConfig` which returns a `config.BrokerConfig` ?

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -210,46 +227,68 @@ func (s *strimziProvider) configureKafkaConnectCluster() error {
...
+			}
+			port := listener.Addresses[0].Port
+			if port != nil {
+				p := int(*port)
+				bc.Port = &p
+			}
+			s.Config.BrokersTLS = append(s.Config.BrokersTLS, bc)
+
+		} else if listener.Type != nil && (*listener.Type == "plain" || *listener.Type == "tcp") {
+			bc := config.BrokerConfig{
```
bsquizz, line 275: Can we move this to a func called `buildTcpBrokerConfig` which returns a `config.BrokerConfig` ?

bsquizz (changes requested):
Looks good, mainly I just see some opportunities to move long code blocks into functions that describe what the code block is doing for easier readability -- and a couple places where I'm wondering if we need to use `errors.Wrap` instead of returning a pure `err` -- IF it is a section of code we'd want a re-queue to occur on if it fails (since in my prior experience, the pure `err` does not cause a reconciler re-queue)

Do we have tasks backlogged to get PRs into app-common-* to add the new `brokersTLS`?

controllers/cloud.redhat.com/providers/kafka/strimzi.go
psav: I see no reason why not

controllers/cloud.redhat.com/providers/kafka/strimzi.go
psav: We shouldn't need to, because any error created by a provider will be cause by the provider and wrapped

controllers/cloud.redhat.com/providers/kafka/strimzi.go
psav: as above

controllers/cloud.redhat.com/providers/kafka/strimzi.go
psav: We can

controllers/cloud.redhat.com/providers/kafka/strimzi.go
psav: we can

controllers/cloud.redhat.com/providers/kafka/strimzi.go
psav: we can

controllers/cloud.redhat.com/providers/kafka/strimzi.go
psav: Actually this doesn't need to return the ku at all

psav:
@bsquizz Fixed up the last of the changes you requested

bsquizz (approved):

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -77,18 +81,24 @@ func (s *strimziProvider) configureKafkaCluster() error {
 
 	k.Spec = &strimzi.KafkaSpec{
 		Kafka: strimzi.KafkaSpecKafka{
+			Authorization: &strimzi.KafkaSpecKafkaAuthorization{
```
kylape, line 84: Does this mean that apps must use auth to connect to kafka when using strimzi mode?

bsquizz: My understanding of the code here was the app config will have both `brokers` and `brokersTLS` so if you want TLS you'd have to choose one of the TLS brokers.

kylape: auth is orthogonal to tls

kylape: unless there's an implicit lack of auth on non-tls connections

bsquizz: So yeah, I see what you mean, if this config here sets the auth type globally for both tcp and TLS that's probably not ideal. We'll need to double check the strimzi docs...

bsquizz: My hope is, this sets the auth type and applies only for a listener that has `Authentication` config on it

bsquizz: https://strimzi.io/docs/operators/in-development/using.html#con-securing-kafka-authorization-str

> You can configure authorization for Kafka brokers using the authorization property in the Kafka.spec.kafka resource. If the authorization property is missing, no authorization is enabled and clients have no restrictions. When enabled, authorization is applied to **all enabled listeners**. The authorization method is defined in the type field.

Unless we intend to always use auth, it seems like we might need a flag in the kafka ClowdEnvironment config that allows us to turn auth completely off for a cluster?

If we always use auth, we'll have to modify various iqe plugins so they connect to kafka using users as well... we might want to coordinate on "flipping that switch"

psav: The Non TLS endpoint is not auth
The TLS endpoint is

psav: oh - perhaps it IS applied to all then

psav: I'll check this out - I hoped you had run it up @bsquizz as you have more kafka experience than I - I'll do this later

bsquizz: Hah, yeah, I made the assumption auth wouldn't apply to a tcp listener... thinking "who would pass auth over a plain connection"? Based on the strimzi doc, bad assumption perhaps. There's also the chance the strimzi doc isn't 100% clear on it. Maybe we just need to spin it up and try it to confirm the behavior.

psav: yeh I can do that

psav: It is right - auth on means auth for all

kylape (approved):

PR #288, opened 2021-04-06
Generation check added to ClowdApp reconciliation 

psav:
* Currently app + env reconciliations can often get kicked off together
  and result in one reconciliation failing as it tries to modify a resource
  that either already exists now, or has been modified by another
  reconciliation. This is unnecessary.
* The generation check ensures that a ClowdEnv has been reconcilied before
  the ClowdApp is allowed to reconcile. This ensures that the ClowdApp
  always has the latest, up to date, reconciled env to work with.

### DIFF
--- a/controllers/cloud.redhat.com/clowdapp_controller.go
+++ b/controllers/cloud.redhat.com/clowdapp_controller.go
@@ -140,6 +140,10 @@ func (r *ClowdAppReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	if env.Generation != env.Status.Generation {
+		return ctrl.Result{Requeue: true}, errors.New(fmt.Sprintf("Clowd Environment not yet reconciled: %s", env.Name))
+	}
+
 	if !env.IsReady() {
 		r.Recorder.Eventf(&app, "Warning", "ClowdEnvNotReady", "Clowder Environment [%s] is not ready", app.Spec.EnvName)
 		log.Info("Env not yet ready", "app", app.Name, "namespace", app.Namespace)
@@ -190,7 +194,7 @@ func (r *ClowdAppReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	// Delete all resources that are not used anymore
 	if !requeue {
-		r.Recorder.Eventf(&app, "Normal", "SuccessfulCreate", "created", app.GetClowdName())
+		r.Recorder.Eventf(&app, "Normal", "SuccessfulReconciliation", "Clowdapp reconciled [%s%]", app.GetClowdName())
 		log.Info("Reconciliation successful", "app", fmt.Sprintf("%s:%s", app.Namespace, app.Name))
 		err := cache.Reconcile(&app)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -140,6 +140,11 @@ func (r *ClowdAppReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	if env.Generation != env.Status.Generation {
+		r.Recorder.Eventf(&app, "Warning", "ClowdEnvNotReconciled", "Clowder Environment [%s] has not been reconciled", app.Spec.EnvName)
```
kylape, line 144: I kind of think we shouldn't record an event if this happens.  We're just properly preventing a race condition here, right?

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -190,15 +195,15 @@ func (r *ClowdAppReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	// Delete all resources that are not used anymore
 	if !requeue {
-		r.Recorder.Eventf(&app, "Normal", "SuccessfulCreate", "created", app.GetClowdName())
+		r.Recorder.Eventf(&app, "Normal", "SuccessfulCreate", "Clowdapp created [%s]", app.GetClowdName())
```
kylape, line 198: wouldn't this event also fire for all reconciliations after initial creation?

controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -195,22 +195,23 @@ func (r *ClowdEnvironmentReconciler) Reconcile(req ctrl.Request) (ctrl.Result, e
...
+	env.Status.Generation = env.Generation
 
 	if err := r.Client.Status().Update(ctx, &env); err != nil {
 		return ctrl.Result{}, err
 	}
 
 	if !requeue {
 		// Delete all resources that are not used anymore
-		r.Recorder.Eventf(&env, "Normal", "SuccessfulCreate", "created", env.GetClowdName())
+		r.Recorder.Eventf(&env, "Normal", "SuccessfulCreate", "Environment created [%s]", env.GetClowdName())
```
kylape, line 206: wouldn't this event also fire for all reconciliations after initial creation?

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -140,6 +140,11 @@ func (r *ClowdAppReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	if env.Generation != env.Status.Generation {
+		r.Recorder.Eventf(&app, "Warning", "ClowdEnvNotReconciled", "Clowder Environment [%s] has not been reconciled", app.Spec.EnvName)
+		return ctrl.Result{Requeue: true}, errors.New(fmt.Sprintf("Clowd Environment not yet reconcile: %s", env.Name))
```
kylape, line 145: "not yet reconciled"

kylape (changes requested):

controllers/cloud.redhat.com/clowdapp_controller.go
psav: Yes - it's really a successful reconciliation message - we could omit entirely

kylape (approved):

PR #290, opened 2021-04-07
Removed consumerGroup

psav:
Removes old consumer group - needs all clients updating


BlakeHolifield (commented):
This looks fine as far as spec and schema are concerned. I don't really know enough Kafka to understand how this will affect stage/prod.

psav:
As far as I understand it - it won't - no one uses it

PR #292, opened 2021-04-08
Add new mutating webhook to convert old version to new

psav:
* Currently there are several deprecated features of the spec these are dealt
  with by using certain shims in different places.
* This commit moves all of the 'shim' functionality to a mutating webhook
  so that specs are mutated on creation/edit and Clowder then only deals
  with the latest version of the three specs.
* In time, each shim deprecation will be removed from the webhook, but it is
  less costly to keep it than if it is polluting the codebase.
* kubesetup.sh has been midified to download and install the cert-manager
  required for the creation and security of the webhook endpoint
* Certain YAMLs have been tweaked to provide the mutating webhook endpoint
* All current mutations/shims have been removed from the controller itself.
* Tests have been added to assert the mutation.
* Some tests had a database image removed as it is no longer required.

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -1,0 +1,52 @@
+
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: mutating-webhook-configuration
+webhooks:
+- clientConfig:
+    caBundle: Cg==
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-app
+  failurePolicy: Fail
+  admissionReviewVersions:
+  - "v1alpha1"
+  - "v1beta1"
+  sideEffects: "None"
+  name: clowdapp.cloud.redhat.com
+  rules:
+  - apiGroups:
+    - cloud.redhat.com
+    apiVersions:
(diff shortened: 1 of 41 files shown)


config/webhook/network_policy.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-to-admission-controller
+spec:
+  ingress:
+  - {}
```
BlakeHolifield, line 7: Enables default allow on all ingress traffic to clowder, correct?

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,52 @@
...
+- clientConfig:
+    caBundle: Cg==
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-env
+  failurePolicy: Fail
+  admissionReviewVersions:
+  - "v1alpha1"
+  - "v1beta1"
```
BlakeHolifield, line 40: Did we get this bit resolved? I remember having an issue with the `ReviewVersions`

BlakeHolifield (commented):
Just a few questions, as usual :smile:

BlakeHolifield:
The ticket is closed for this, should we close the PR as well?

PR #289, opened 2021-04-07
Added role binding and per app service accounts

psav:
* Added K8sAccessLevel to Deployment Spec.
* Added GetDeploymentNamespacedName() to ClowdApp.
* Deleted GetDeploymentName() from various places and replaced with the
  ClowdApp.GetDeploymentNamespacedName.
* Upgraded access rights for controller to deal with RoleBindings
* Now create a service account per deployment to allow for more granular
  access levels.

PR #296, opened 2021-04-09
Add Iqe Support via Jobs

BlakeHolifield:
Closes [RHCLOUD-12947](https://issues.redhat.com/browse/RHCLOUD-12947)
* Adds support for IQE Jobs via ClowdJobInvocation
* [design doc](https://docs.google.com/document/d/1ZCmY_50wkoxOu-mRJF3FWluTO1083hvZ5v-GCXQVER8/edit?ts=60535626)
* [env format doc](https://docs.google.com/document/d/1vJ-6g5fC6tGcq9VvVOkKrFrK8g4RKt4blHByutvM-y4/edit#heading=h.cihkmu1ajzgh)

Remaining Tasks:

- [x] Resolve upstream IQE image issue
- [x] Consume and implement ServiceAccount provider changes from #289 
- [x] Add Kuttl tests for IQE job


apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -243,6 +243,10 @@ type KafkaTopicSpec struct {
 	TopicName string `json:"topicName"`
 }
 
+type IqeSpec struct {
```
psav, line 246: Let's name this `TestingSpec` and the corresponding section `Testing`

apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
```diff
@@ -23,13 +23,31 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+type IqeJobSpec struct {
+	// By default, Clowder will set the image on the ClowdJob to be the
+	// baseImage:name-of-iqe-plugin, but can be overridden here
```
psav, line 28: We should point out that only the tag can be changed here

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -150,19 +164,44 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
-		if err := r.InvokeJob(ctx, &job, &app, &env, &cji); err != nil {
+		if err := r.InvokeJob(&cache, &job, &app, &env, &cji); err != nil {
 			r.Log.Info("Job Invocation Failed", "jobinvocation", jobName, "namespace", app.Namespace)
-			r.Recorder.Eventf(&cji, "Error", "JobNotInvoked", "Job [%s] could not be invoked", jobName)
+			r.Recorder.Eventf(&cji, "Warning", "JobNotInvoked", "Job [%s] could not be invoked", jobName)
 			return ctrl.Result{Requeue: true}, err
 		}
 	}
 
+	if cji.Spec.Iqe.Marker != "" {
```
psav, line 174: Why is this wrapped around a marker check?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -75,6 +82,10 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
 		return ctrl.Result{}, err
 	}
 
+	cache := providers.NewObjectCache(ctx, r.Client, r.Scheme)
+
+	var IqeClowdJob = providers.NewSingleResourceIdent("iqeclowdjob", "core_iqeclowdjob", &crd.ClowdJobInvocation{})
```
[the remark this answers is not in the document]
psav: This also should have the `Job` as the template, the cache isn't going to create a CJI, it's going to create multiple Jobs.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -150,19 +164,44 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+	}
+
 	// Short running jobs may be done by the time the loop is ranged,
 	// so we update again before the reconcile ends
 	err = r.setCompletedStatus(ctx, &cji)
 	if err != nil {
 		return ctrl.Result{Requeue: true}, err
 	}
+
+	cacheErr := cache.ApplyAll()
```
psav, line 201:
```
	cacheErr := cache.ApplyAll()
	if cacheErr != nil {
		return ctrl.Result{}, err
	}
```

should become for spacesave + readability

```
	if cacheErr := cache.ApplyAll(); cacheErr != nil {
		return ctrl.Result{}, cacheErr
	}
```

also, you are returning `err`, not `cacheErr`

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -173,102 +212,244 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
-func createJobResource(cji *crd.ClowdJobInvocation, env *crd.ClowdEnvironment, nn types.NamespacedName, job *crd.Job, j *batchv1.Job) {
+func (r *ClowdJobInvocationReconciler) fetchConfig(name types.NamespacedName, ctx context.Context) (config.AppConfig, error) {
+
+	secretConfig := core.Secret{}
+	jsonContent := config.AppConfig{}
+
+	err := r.Client.Get(ctx, name, &secretConfig)
+
+	if err != nil {
+		return jsonContent, err
```
psav, line 244: Personally I would prefer to ensure we send back `config.AppConfig{}`, rather than `jsonContent` - that way we can ensure it's null in the case of an error. Also, jsonContent isn't storing jsonContent, it's a struct, could be a little misleading.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -173,102 +212,244 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+	}
+
+	err = json.Unmarshal(secretConfig.Data["cdappconfig.json"], &jsonContent)
+	if err != nil {
```
psav, line 248: Let's use the same pattern for this and the one above that I described earlier

```
	if err := json.Unmarshal(secretConfig.Data["cdappconfig.json"], &jsonContent); err != nil {
		return jsonContent, err
	}
```

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -173,102 +212,244 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+	iqeSecret := &core.Secret{}
+
+	appList := crd.ClowdAppList{}
+	r.Client.List(ctx, &appList)
+
+	nn := types.NamespacedName{
+		Name:      envName,
+		Namespace: cji.Namespace,
+	}
+	update, err := utils.UpdateOrErr(r.Client.Get(ctx, nn, iqeSecret))
```
psav, line 265: We should use the cache here too

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -173,102 +212,244 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+}
+
+func (r *ClowdJobInvocationReconciler) createAndApplyIqeSecret(ctx context.Context, cji *crd.ClowdJobInvocation, envName string) error {
+	iqeSecret := &core.Secret{}
+
+	appList := crd.ClowdAppList{}
+	r.Client.List(ctx, &appList)
+
+	nn := types.NamespacedName{
+		Name:      envName,
```
psav, line 262: This is potentially going to cause issues, multiple Jobs are going to conflict on this secret name right, one could remove it before another is finished using it.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -173,102 +212,244 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+	iqeSecret.SetName(nn.Name)
+	iqeSecret.SetNamespace(nn.Namespace)
+
+	// This should maybe be owned by the job
+	iqeSecret.SetOwnerReferences([]metav1.OwnerReference{cji.MakeOwnerReference()})
+
+	// loop through secrets and get their appConfig
+	envConfig := make(map[string]interface{})
+	appConfigs := make(map[string]config.AppConfig)
+	for _, app := range appList.Items {
```
psav, line 280: EnvName is indexed iirc so you can use filter options to the List to make this quicker and less nested.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -173,102 +212,244 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
-		Args:         pod.Args,
-		Env:          envvar,
-		Resources:    deployProvider.ProcessResources(&pod, env),
-		VolumeMounts: pod.VolumeMounts,
-		Ports: []core.ContainerPort{{
-			Name:          "metrics",
-			ContainerPort: env.Spec.Providers.Metrics.Port,
-		}},
+		Name:      fmt.Sprintf("%s-iqe", plugin),
+		Image:     fmt.Sprintf("%s:%v", iqeImage, tag),
```
psav, line 369: can't we use `%s` here? It should be a string no?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -173,102 +212,244 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+	// Use edit level service account to create and delete resources
+	// one per app when the app is created
+	case "edit":
+		// TODO: Create secret here
+		j.Spec.Template.Spec.ServiceAccountName = fmt.Sprintf("%v-iqe", app.Name)
+	// Standard view access to the owned resources
+	case "view":
+		j.Spec.Template.Spec.ServiceAccountName = fmt.Sprintf("%v", app.Name)
+	// Standard access
+	case "default":
```
psav, line 361: case default - don't we just omit the service account name entirely?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -173,102 +212,244 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+func constructIqeCommand(cji *crd.ClowdJobInvocation, plugin string) []string {
+	command := []string{}
+	command = append(command, "iqe")
+	command = append(command, "tests")
+	command = append(command, "plugin")
+	command = append(command, fmt.Sprintf("%v", strings.ReplaceAll(plugin, "-", "_")))
+	command = append(command, "-m")
+	command = append(command, fmt.Sprintf("%v", cji.Spec.Iqe.Marker))
+	command = append(command, "-k")
+	command = append(command, fmt.Sprintf("%v", cji.Spec.Iqe.Filter))
```
psav, line 450: I'd prefer to see something like

```
	command := []string{
		"iqe", "tests", "plugin",
		fmt.Sprintf("%v", strings.ReplaceAll(plugin, "-", "_")),
		"-m", fmt.Sprintf("%s", "marker"),
		"-k", fmt.Sprintf("%s", "test-name"),
	}
```

It makes it clearer that the args coming after the `-m` etc are related. Also, Filter is allowed to be empty, which would break this, I think Marker should be allowed to be empty too.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -370,9 +569,28 @@ func (r *ClowdJobInvocationReconciler) setCompletedStatus(ctx context.Context, c
...
 		}
 	}
+	return jobsCompleted == completionsNeeded
+}
 
-	cji.Status.Completed = completionsNeeded == jobsCompleted
+func getIqeStatus(jobs *batchv1.JobList, cji *crd.ClowdJobInvocation) bool {
+	jobCompleted := false
+	if len(cji.Status.Jobs) == 0 {
+		cji.Status.Jobs = []string{}
```
psav, line 578: What's going on here?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -370,9 +569,28 @@ func (r *ClowdJobInvocationReconciler) setCompletedStatus(ctx context.Context, c
...
+		return false
+	}
 
-	return nil
+	if cji.Spec.Iqe.Marker != "" {
```
psav, line 582: Here is the marker check again.....why?

controllers/cloud.redhat.com/providers/iqe/default.go
```diff
@@ -0,0 +1,61 @@
...
+
+	p "cloud.redhat.com/clowder/v2/controllers/cloud.redhat.com/providers"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type iqeProvider struct {
+	p.Provider
+	Config config.IqeConfig
+}
+
```
psav, line 21: Perhaps put all your ProviderResource definitions here?

psav (commented):
I know this is only draft - but thought I'd get in there early.

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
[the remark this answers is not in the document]
BlakeHolifield: Are you proposing a new top level `Testing` struct that has `Iqe` as a member?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
BlakeHolifield: Yeah, that's probably correct. I wasn't sure if the default behavior was to use the `default` service account or not.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
BlakeHolifield: I was going to poll for a cleaner way to do this, so big thanks there.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -173,102 +212,244 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+				return err
+
+			}
+			appConfigs[app.Name] = jsonContent
+		}
+	}
+	envConfig["cdappconfigs"] = appConfigs
+
+	// Marshall the data into the top level "cdenvconfig.json" to be mounted as a single secret
+	envData, err := json.Marshal(envConfig)
```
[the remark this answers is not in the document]
BlakeHolifield: Does the pattern work for multi-return values? If you start the statement with `if envData, err`, the compiler becomes quite cross.

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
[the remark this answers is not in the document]
psav: I was - what do you think?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -152,125 +161,280 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
-			Secret: &core.SecretVolumeSource{
-				SecretName: cji.Spec.AppName,
-			},
-		},
-	})
+func constructIqeCommand(cji *crd.ClowdJobInvocation, plugin string) []string {
+	command := []string{
+		"iqe", "tests", "plugin",
+		fmt.Sprintf("%v", strings.ReplaceAll(plugin, "-", "_")),
+		"-m", cji.Spec.Iqe.Marker,
```
psav, line 434: Still gotta account for these being null....

BlakeHolifield: Marker can't be null via the spec, but I added a check for the filter. Also seems like passing `-k ""` does nothing in the iqe plugins.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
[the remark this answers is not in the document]
psav: Var it outside, then you're good

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
psav: Yeh thanks for the check on that, should we do the same for Marker?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -345,23 +508,43 @@ func (r *ClowdJobInvocationReconciler) cjiToEnqueueUponJobUpdate(a handler.MapOb
...
+	if err := r.Client.List(ctx, &jobs, client.InNamespace(cji.ObjectMeta.Namespace)); err != nil {
+		return err
+	}
+
+	cji.Status.Completed = getInvocationStatus(&jobs, cji)
+
+	return nil
+}
+
+func getInvocationStatus(jobs *batchv1.JobList, cji *crd.ClowdJobInvocation) bool {
```
psav, line 521: I would think that a Status, would return something other than a bool, how do we know what true/false means? If you intend to keep it as a bool, you could encode the meaning into the function name. `isInvocationComplete` ?

BlakeHolifield: Maybe this is a misnomer then. "Status" is a boolean about the completion of a CJI. It doesn't go into more detail about what worked/what didn't work. I can update to reflect the boolean(ness)? :smile:

psav: I think an isDone kinda deal is nice here then

apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
```diff
@@ -315,6 +345,9 @@ type ProvidersConfig struct {
 
 	// Defines the pull secret to use for the service accounts.
 	PullSecrets PullSecrets `json:"pullSecrets,omitempty"`
+
+	// Defines the environment for iqe/smoke testing
+	Iqe IqeConfig `json:"iqe,omitempty"`
```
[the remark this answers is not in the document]
psav: this is bunk

apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
```diff
@@ -266,6 +266,36 @@ type InMemoryDBConfig struct {
 	PVC bool `json:"pvc,omitempty"`
 }
 
+// Sets the access level the iqe job based on the mode
+// +kubebuilder:validation:Enum=none;view;edit
+type AccessLevelMode string
```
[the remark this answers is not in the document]
psav: `K8sAccessLevel` exists in clowdapp types, we should probably use that, enum is slightly different

apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
```diff
@@ -266,6 +266,41 @@ type InMemoryDBConfig struct {
...
+
+type TestingConfig struct {
+	// Defines the environment for iqe/smoke testing
+	Iqe IqeConfig `json:"iqe,omitempty"`
+
+	// The mode of operation of the IQE Pod. Valid options are:
+	// (*_none_*) where no access will be granted, (*_view_*)
+	// which will allow the pod to have the controlling app's view permissions
+	// and (*_edit_*) which grants full create, and edit powers to the pod
+	K8SAccessLevel AccessLevelMode `json:"k8sAccessLevel"`
```
psav, line 285: And so this should change to be the one defined above

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -49,6 +57,10 @@ type ClowdJobInvocationReconciler struct {
 	Recorder record.EventRecorder
 }
 
+var IqeClowdJob = providers.NewSingleResourceIdent("iqeclowdjob", "core_iqe_clowdjob", &batchv1.Job{})
```
psav, line 60: The first argument to this is supposed to be a provider string, so I think setting it to `cji` in all three cases is most appropriate

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -65,23 +77,23 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
 	cji := crd.ClowdJobInvocation{}
-	err := r.Client.Get(ctx, req.NamespacedName, &cji)
-	if err != nil {
+	if err := r.Client.Get(ctx, req.NamespacedName, &cji); err != nil {
 		if k8serr.IsNotFound(err) {
 			// Must have been deleted
 			return ctrl.Result{}, nil
 		}
-		r.Log.Error(err, "CJI not found ")
+		r.Log.Error(err, "CJI not found", "clowdjobinvocation")
```
psav, line 85: This will trigger a warning, because arguments 3+ are supposed to be key/value pairs, like 
```r.Log.Error(err, "message", "controller", "cji")```

File: controllers/cloud.redhat.com/clowdjobinvocation_controller.go [hunk 25]
```diff
@@ -152,125 +164,287 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+		}
+
+		j := batchv1.Job{}
+
+		if err := r.createIqeJobResource(&cache, &cji, &env, &app, nn, ctx, &j); err != nil {
+			r.Log.Error(err, "Iqe Job creation encountered an error", "jobinvocation", nn.Name)
+			r.Recorder.Eventf(&cji, "Warning", "IQEJobFailure", "Job [%s] failed to invoke", j.ObjectMeta.Name)
+			return ctrl.Result{Requeue: true}, err
+		}
+		if err := cache.Create(IqeClowdJob, nn, &j); err != nil {
```
psav, line 191: The pattern is that we should create the object in the cache before we modify and then update it afterwards.
```
obj = emptyObject{}
cache.Create(emptyObject...
manipulate(emptyObject)
cache.Update(emptyObject...
```
Whilst in this case it is unlikely the object already exists, the reason for doing it this way is to preserve other settings on the object. If anything else modifies the object, in a way that doesn't conflict with clowder, then clowder will leave those changes alone, and include them in it's "update" otherwise it will just overwrite them. We _may_ wnt this behaviour in some cases, but to keep with consistency I think we should update this for now.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -152,125 +164,287 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+			r.Log.Error(err, "Iqe Job could not update via cache", "jobinvocation", nn.Name)
+			r.Recorder.Eventf(&cji, "Warning", "IQECacheUpdateFailed", "Job [%s] could not update", j.ObjectMeta.Name)
```
psav, line 193: Personally I think reporting the Cache as a top level issue is a little too granular.....IQEJobFailure is fine, with a message about not being able to update the cache.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -152,125 +164,287 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+		r.Log.Info("Iqe Job Invoked Successfully", "jobinvocation", nn.Name, "namespace", app.Namespace)
+		r.Recorder.Eventf(&cji, "Normal", "IQEJobInvoked", "Job [%s] was invoked successfully", j.ObjectMeta.Name)
+	}
+
 	// Short running jobs may be done by the time the loop is ranged,
 	// so we update again before the reconcile ends
-	err = r.setCompletedStatus(ctx, &cji)
-	if err != nil {
-		return ctrl.Result{Requeue: true}, err
+	if statusErr := r.setCompletedStatus(ctx, &cji); statusErr != nil {
```
psav, line 204: I don't think this is true, at least not since you are using the cache, because your apply to the cache happens after this message, so nothing will have been written out yet, so nothing can complete.....right?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -152,125 +164,287 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
-		Namespace: app.Namespace,
+		Namespace: cji.Namespace,
 	}
-	j := batchv1.Job{}
 
-	createJobResource(cji, env, nn, job, &j)
-	if err := r.Client.Create(ctx, &j); err != nil {
+	j := batchv1.Job{}
+	jobProvider.CreateJobResource(cji, env, nn, job, &j)
+	if err := cache.Create(ClowdJob, nn, &j); err != nil {
```
psav, line 228: Same comment as above

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -152,125 +164,287 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+	case "edit":
+		labeler := utils.GetCustomLabeler(nil, nn, app)
+		if err := svcAccounts.CreateServiceAccount(cache, svcAccounts.CoreAppServiceAccount, env.Spec.Providers.PullSecrets, nn, labeler); err != nil {
+			r.Recorder.Eventf(cji, "Warning", "ServiceAccountNotCreated", "Unable to create service account [%s]", nn.Name)
+			return err
+		}
+		j.Spec.Template.Spec.ServiceAccountName = nn.Name
+	// Standard view access to the owned resources
+	case "view":
+		j.Spec.Template.Spec.ServiceAccountName = app.Name
```
psav, line 365: I think this will have `-app` appended to it

psav: This also assumes that the app has "view" permissions - which by default it doesn't.......I think you need to create an iqe secret in this case also

psav: And thinking about it - The way that we are going to move for doing pull secrets - I think you'll ALWAYS need a service account - which will be populated with the pull secrets

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -152,125 +164,287 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
-	j.Spec.Template.Spec.Volumes = pod.Volumes
-	j.Spec.Template.Spec.Volumes = append(j.Spec.Template.Spec.Volumes, core.Volume{
-		Name: "config-secret",
-		VolumeSource: core.VolumeSource{
-			Secret: &core.SecretVolumeSource{
-				SecretName: cji.Spec.AppName,
-			},
-		},
-	})
+func constructIqeCommand(cji *crd.ClowdJobInvocation, plugin string) ([]string, error) {
```
psav, line 437: This never returns an error

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -152,125 +164,287 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
+		j.Spec.Template.Spec.ServiceAccountName = nn.Name
+	// Standard view access to the owned resources
+	case "view":
+		j.Spec.Template.Spec.ServiceAccountName = app.Name
+	}
+
+	// var constructedIqeCommand []string
+	constructedIqeCommand, err := constructIqeCommand(cji, plugin)
+	if err != nil {
+		return errors.New("IQE Marker must be set")
```
psav, line 371: We should not infer that Marker has not been set from the error, rather wrap up, or fully return the error, plus, looking at the code for the function, it will never return an error anyway ;)

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -152,125 +164,287 @@ func (r *ClowdJobInvocationReconciler) Reconcile(req ctrl.Request) (ctrl.Result,
...
-		VolumeMounts: pod.VolumeMounts,
-		Ports: []core.ContainerPort{{
-			Name:          "metrics",
-			ContainerPort: env.Spec.Providers.Metrics.Port,
-		}},
-		ImagePullPolicy: core.PullIfNotPresent,
+		VolumeMounts: []core.VolumeMount{},
+		// Because the tags on iqe plugins are not commit based, we need to pull everytime we run.
+		// A leftover tag from a previous run is never guaranteed to be up to date
+		ImagePullPolicy: core.PullAlways,
```
psav, line 383: @kylape are we gonna in trouble for this :)

controllers/cloud.redhat.com/providers/testing/provider.go
```diff
@@ -0,0 +1,16 @@
+package testing
```
psav, line 1: I don't see any tests that use this - is that right?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
BlakeHolifield: Lotta oopsies in this function.

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
BlakeHolifield: Thanks for the clarification here. Will update.

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -250,6 +250,10 @@ type KafkaTopicSpec struct {
 	TopicName string `json:"topicName"`
 }
 
+type TestingSpec struct {
+	IqePlugin string `json:"iqe-plugin"`
```
bsquizz, line 254: Maybe a nitpick, but I think we're mostly `camelCase` in the specs -- can we go with name `iqePlugin` for the json here?

apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
```diff
@@ -269,6 +269,35 @@ type InMemoryDBConfig struct {
...
 
+// Describes what amount of app config is mounted to the pod
+// +kubebuilder:validation:Enum={"none", "app", "", "environment"}
+type ConfigAccessMode string
+
+type TestingConfig struct {
+	// Defines the environment for iqe/smoke testing
+	Iqe IqeConfig `json:"iqe,omitempty"`
+
+	// The mode of operation of the IQE Pod. Valid options are:
```
bsquizz, line 280: Technically now `The mode of operation of the testing Pod` :)

bsquizz (changes requested):
I have two minor nitpicks and other than that it looks good to me!

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
BlakeHolifield: Good catch! Fixed.

apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
BlakeHolifield: I almost had it all fixed lol. Updated.

PR #298, opened 2021-04-14
Added local FF tests

psav:


bundle/tests/scorecard/kuttl/test-ff-local/01-pods.yaml
```diff
@@ -0,0 +1,49 @@
...
+      mode: none
+    db:
+      mode: none
+    logging:
+      mode: none
+    objectStore:
+      mode: none
+    inMemoryDb:
+      mode: none
+    pullSecrets:
```
BlakeHolifield, line 26: If this is merged after #304 , we'll need to update this

BlakeHolifield (commented):
Not sure about the test issues, but only one heads up.

psav:
Right - the tests fail because we need to wait for FF server to come up which it can't because we don't have the pull secret in the tests to be able to pull from the quay repo that has the db image that we use for FF

PR #300, opened 2021-04-15
Fix error when trying to get a deleted Job

BlakeHolifield:
* Fixes a bug where CronJobs are causing reconciliations because the ClowdJob controller is watching them. It expects the job name to be there and will error if it does not.
* `{"level":"error","ts":1618516894.2577295,"logger":"controllers.ClowdJobInvocation","msg":"Failed to fetch Job","error":"Job.batch \"crd-pruner-1618506000\" not found"`
* The above entry no longer exists and is causing errors

### DIFF
--- a/controllers/cloud.redhat.com/clowdjobinvocation_controller.go
+++ b/controllers/cloud.redhat.com/clowdjobinvocation_controller.go
@@ -312,7 +312,12 @@ func (r *ClowdJobInvocationReconciler) cjiToEnqueueUponJobUpdate(a handler.MapOb
 
 	job := batchv1.Job{}
 	if err := r.Client.Get(ctx, obj, &job); err != nil {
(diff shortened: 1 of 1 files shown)


kylape:
Is there a difference between returning `nil` and returning and empty `[]reconcile.Request{}`?

BlakeHolifield:
> Is there a difference between returning `nil` and returning and empty `[]reconcile.Request{}`?

Functionally, I'm not sure. I was following the pattern set here: https://github.com/RedHatInsights/clowder/blob/bc0e4c927e70645d1c898e5d37908e94483aca14/controllers/cloud.redhat.com/clowdapp_controller.go#L368

I can't tell how that enqueueHandler handles errors.

kylape (approved):
Okay I just checked another place where this was done, and it checks out.

PR #301, opened 2021-04-16
Added kafka managed mode

psav:


controllers/cloud.redhat.com/providers/kafka/managed.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	kafkaConfig.Brokers = []config.BrokerConfig{broker}
+
+	for _, topic := range app.Spec.KafkaTopics {
+
+		kafkaConfig.Topics = append(
+			kafkaConfig.Topics,
+			config.TopicConfig{
+				Name:          topic.TopicName,
```
BlakeHolifield, line 71: For the other modes, the Name is set to `topicName := fmt.Sprintf("%s-%s-%s", topic.TopicName, k.Env.Name, k.Env.GetClowdNamespace())`. Do we also need to do that here? Or is managed kafka going to use the `Name` and `RequestedName` ?

BlakeHolifield (commented):
One question about topic creation.

controllers/cloud.redhat.com/providers/kafka/managed.go
psav: I made managed use the same mechanism as app-interface, but you raise a good point, we may need to support the situation where we have multiple copies of the same app using the same managed kafka instance.....

PR #305, opened 2021-04-16
Set ready status when a replica count is zero

kylape:
When a specified replica count is zero, we should mark the deployment as
ready.

I've also taken the liberty to add some comments to the
DeploymentStatusChecker function because the different checks were
difficult to understand at first (and maybe second) glance.

### DIFF
--- a/controllers/cloud.redhat.com/utils/utils.go
+++ b/controllers/cloud.redhat.com/utils/utils.go
@@ -254,8 +254,8 @@ func MakePVC(pvc *core.PersistentVolumeClaim, nn types.NamespacedName, labels ma
 	}
 }
 
-// DeploymentStatusChecker takes a deployment and returns True if it is deemed
-// ready by the logic in the function.
+// DeploymentStatusChecker takes a deployment and returns True if it is deemed ready by the logic in
+// the function.
 func DeploymentStatusChecker(deployment *apps.Deployment) bool {
 	if deployment.Generation > deployment.Status.ObservedGeneration {
 		// The deployment controller still needs to reconcile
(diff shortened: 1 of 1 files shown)


controllers/cloud.redhat.com/utils/utils.go
```diff
@@ -254,22 +254,43 @@ func MakePVC(pvc *core.PersistentVolumeClaim, nn types.NamespacedName, labels ma
...
+	}
+
+	if deployment.Spec.Replicas != nil {
+		if *deployment.Spec.Replicas == 0 {
+			// Since there's no replicas, there's nothing to do to get ready
+			return true
 		}
-		if deployment.Status.AvailableReplicas < deployment.Status.UpdatedReplicas {
+
+		if deployment.Status.UpdatedReplicas < *deployment.Spec.Replicas {
```
psav, line 271: I'm guessing if `Replicas` is 0, then `UpdatedReplicas` will never be set at all?

controllers/cloud.redhat.com/utils/utils.go
```diff
@@ -254,22 +254,43 @@ func MakePVC(pvc *core.PersistentVolumeClaim, nn types.NamespacedName, labels ma
 	}
 }
 
-// DeploymentStatusChecker takes a deployment and returns True if it is deemed ready by the logic
-// in the function.
+// DeploymentStatusChecker takes a deployment and returns True if it is deemed
+// ready by the logic in the function.
```
psav, line 258: Shortening the line length? I thought we used 100?

kylape: I can change it to 100

controllers/cloud.redhat.com/utils/utils.go
kylape: I haven't tested it, but if there are zero replicas specified, I wouldn't expect there to be any replicas to update.

BlakeHolifield (approved):

PR #308, opened 2021-04-19
Added generated config for kafka changes

psav:


BlakeHolifield:
The diffs are always a bit hard here. Are these for the new Managed Kafka options are they updating genconfig to match the spec?

BlakeHolifield (approved):

psav:
Yes!

PR #309, opened 2021-04-19
Add Validator

psav:


main.go
```diff
@@ -54,5 +55,5 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseDevMode(false)))
 
-	controllers.Run(metricsAddr, enableLeaderElection, ctrl.GetConfigOrDie(), ctrl.SetupSignalHandler())
+	controllers.Run(metricsAddr, enableLeaderElection, ctrl.GetConfigOrDie(), ctrl.SetupSignalHandler(), !clowder_config.LoadedConfig.Webhooks.Disable)
```
BlakeHolifield, line 58: Any particular reason to do it this way rather than the `boolean` method from the mutating webhook?

apis/cloud.redhat.com/v1alpha1/clowdapp_webhook.go
```diff
@@ -0,0 +1,81 @@
...
+			field.NewPath("spec.Database.Name", "spec.Database.SharedDBAppName"), "cannot set db name and sharedDbApp Name together"),
+		)
+	}
+
+	if len(allErrs) == 0 {
+		return nil
+	}
+
+	return apierrors.NewInvalid(
+		schema.GroupKind{Group: "batch.tutorial.kubebuilder.io", Kind: "CronJob"},
```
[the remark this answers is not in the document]
psav: hahhaha

apis/cloud.redhat.com/v1alpha1/clowdapp_webhook.go
```diff
@@ -0,0 +1,81 @@
...
+		r.Name, allErrs,
+	)
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *ClowdApp) ValidateDelete() error {
+	clowdapplog.Info("validate delete", "name", r.Name)
+
+	// TODO(user): fill in your validation logic upon object deletion.
```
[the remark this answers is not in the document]
psav: And because you need it to fulfill the interface

main.go
psav: Not sure what you mean by this

BlakeHolifield: I guess what's the advantage of this vs [this other way](https://github.com/RedHatInsights/clowder/pull/292/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R57)

psav: Doing it the other means it is ALWAYS on, and was relying on an env var which didn't actually do anything - this new way is much safer and allows you to turn off webhooks when doing `make run`

controllers/cloud.redhat.com/clowder_config/config.go
```diff
@@ -8,6 +8,9 @@ import (
 )
 
 type ClowderConfig struct {
+	Webhooks struct {
+		Disable bool `json:"disable"`
```
kylape, line 12: can we flip the boolean and stick this under the feature flag section now?

kylape (changes requested):
did we want to save the dependency validation (i.e. update introduces unfulfilled dependency) for a later PR?

controllers/cloud.redhat.com/clowder_config/config.go
kylape: just pushed this change up

PR #328, opened 2021-05-07
Move IQE job logic out of invocation controller

BlakeHolifield:
* create an iqe provider shell with only `impl.go`
* relocate logic out of main controller


psav:
I like the essence of this one, but it's not really a provider, not sure if that matters or not, or if we are planning to turn it into one at a later date.

BlakeHolifield:
> I like the essence of this one, but it's not really a provider, not sure if that matters or not, or if we are planning to turn it into one at a later date.

Right. I know we discussed potentially making it a provider, so I didn't want to add it under another utils directory. We can always do that and make it a provider later if we want. Up to you!

PR #334, opened 2021-05-12
More descriptive error messages for elasticache provider

bsquizz:
The current error message of `Missing dependencies: [- in-memory-db-secret: sources-api]` is a little unclear.

Technically the problem is that we're missing the secret named `in-memory-db` in the sources-api app's namespace -- modified the error messages to make this more obvious.

### DIFF
--- a/config/crd/bases/cloud.redhat.com_clowdapps.yaml
+++ b/config/crd/bases/cloud.redhat.com_clowdapps.yaml
@@ -1,13 +1,26 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1
+apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.5.0
+    controller-gen.kubebuilder.io/version: v0.3.0
   creationTimestamp: null
   name: clowdapps.cloud.redhat.com
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.deployments.readyDeployments
+    name: Ready
+    type: integer
+  - JSONPath: .status.deployments.managedDeployments
+    name: Managed
+    type: integer
+  - JSONPath: .spec.envName
+    name: EnvName
+    type: string
+  - JSONPath: .metadata.creationTimestamp
+    name: Age
+    type: date
   group: cloud.redhat.com
   names:
     kind: ClowdApp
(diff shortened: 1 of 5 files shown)


config/crd/bases/cyndi.cloud.redhat.com_cyndipipelines.yaml
```diff
@@ -1,38 +1,13 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
+    controller-gen.kubebuilder.io/version: v0.5.0
```
BlakeHolifield, line 7: Do you need this for cyndi to generate correctly? Trying to determine if you have a different version or if this was an intended version bump.

config/crd/bases/cyndi.cloud.redhat.com_cyndipipelines.yaml
```diff
@@ -1,38 +1,13 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
BlakeHolifield, line 3: Same sort of question here. Seems like you get this from the controller-gen version.

config/crd/bases/cloud.redhat.com_clowdapps.yaml
```diff
@@ -1,26 +1,13 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
+    controller-gen.kubebuilder.io/version: v0.5.0
```
BlakeHolifield, line 7: Here as well

config/crd/bases/cloud.redhat.com_clowdenvironments.yaml
```diff
@@ -1,26 +1,13 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
+    controller-gen.kubebuilder.io/version: v0.5.0
```
BlakeHolifield, line 7: And here

BlakeHolifield (commented):
Good change, just questions about the config gen updates.

psav:
Until we have updated everything in the repo, can you please use the same operator sdk version so that we don't get 7000 lines of file changes?

psav (changes requested):
Need to keep the changes to only relevant ones

bsquizz:
Ugh sorry, too used to using `git add -u`

BlakeHolifield (approved):

controllers/cloud.redhat.com/providers/inmemorydb/elasticache.go
```diff
@@ -50,7 +55,7 @@ func (e *elasticache) Provide(app *crd.ClowdApp, config *config.AppConfig) error
 	if !found {
 		return &errors.MissingDependencies{
 			MissingDeps: map[string][]string{
-				"in-memory-db-secret": {app.Name},
+				"secret": {fmt.Sprintf("name: %s, namespace: %s", secretName, app.Namespace)},
```
psav, line 58: Calling this just `secret` is too generic - we could have multiple "secrets" that are missing dependencies, we either need a more intelligent missing dependency map where we can "add" items to a category, or keep it with a more unique name.

psav (changes requested):

controllers/cloud.redhat.com/providers/inmemorydb/elasticache.go
bsquizz: Ok.. I like your first idea, but sounds like something we probably need a bit more design around ... just reverted the name for now.

psav (approved):

PR #339, opened 2021-05-14
Enable node port

psav:


apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
```diff
@@ -337,6 +337,9 @@ type ClowdEnvironmentSpec struct {
 	// Defines the default resource requirements in standard k8s format in the
 	// event that they omitted from a PodSpec inside a ClowdApp.
 	ResourceDefaults v1.ResourceRequirements `json:"resourceDefaults"`
+
+	// Node port enables the node port for clowder created services
+	NodePort bool `json:"nodePort,omitempty"`
```
BlakeHolifield, line 342: Any reason to make this an object and use the `enabled` pattern? I can't think of one off hand, but I'm just crossing t's and dotting i's

bundle/tests/scorecard/kuttl/test-node-port/00-install.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: test-node-port
+spec:
+  finalizers:
+  - kubernetes
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: test-node-port-secret
```
BlakeHolifield, line 13: I don't see this namespace getting used or deleted in 03. Am I missing something? The env doesn't have a namespacedName.

BlakeHolifield (commented):
Two small comments. NodePorts are going to be a very welcome addition!

apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
kylape: Based on the [k8s service types ](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) page, there are other service types besides NodePort and ClusterIP.  So I would be in favor of renaming this field to `serviceType` and making the value a string bound by an enum

kylape (changes requested):
Let's change the `NodePort bool` to a struct.

```go
type ServiceConfig struct {
    Type string `json:"type"`
}
```
That way we can add additional config if different types need it

PR #340, opened 2021-05-14
Default kafka cluster name to env name

kylape:
If the kafka cluster name is not specified in the env spec, then name
the cluster the same as the environment.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: 127.0.0.1:5000/clowder
-  newTag: 7af0d5f6
+  newTag: d778ee9
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: 127.0.0.1:5000/clowder
-  newTag: d778ee9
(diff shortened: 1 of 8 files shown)


config/manager/kustomization.yaml
```diff
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: 127.0.0.1:5000/clowder
-  newTag: d778ee9
+  newTag: 7af0d5f6
```
bsquizz, line 8: Probably an accidental change -- `git checkout origin/master config/manager/kustomization.yaml`

bsquizz (changes requested):
Spotted 1 thing

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -658,7 +658,7 @@ func (s *strimziProvider) processTopics(app *crd.ClowdApp) error {
 
 		topicName := fmt.Sprintf("%s-%s-%s", topic.TopicName, s.Env.Name, nn.Namespace)
 		knn := types.NamespacedName{
-			Namespace: getKafkaNamespace(s.Env),
+			Namespace: getKafkaName(s.Env),
```
psav, line 661: I think this is erroneous, we should be getting the Namespace here, not the name no?

psav (changes requested):

PR #349, opened 2021-05-25
WIP: Add 'debug' flag for ClowdJobInvocation

bsquizz:
The goal here is to get a pod spun up in the namespace that uses the config of a ClowdJob -- but does not actually run the ClowdJob (so that you can rsh into it and debug)

* Adds 'debug' and 'debugCommand' to ClowdJobInvocation spec
* If CJI is applied with 'debug'=true, a few things happen:
** command of the pod is swapped to debugCommand (by default this is `/bin/sh -c 'while true; do sleep 1; done'`)
** initContainers will not be applied in the podSpec
** pod restart policy is set to `Never`
** name of the container in the job's podSpec will have `-debug` appended to the end

### DIFF
--- a/apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
+++ b/apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
@@ -60,6 +60,12 @@ type ClowdJobInvocationSpec struct {
 
 	// Testing is the struct for building out test jobs (iqe, etc) in a CJI
 	Testing JobTestingSpec `json:"testing,omitempty"`
+
+	// invokes the job pods in "debug" mode -- i.e., command is swapped to the debugCommand
+	Debug bool `json:"debug,omitempty"`
+
+	// indicates pod command to run when in debug mode (default: "/bin/sh -c 'while true; do sleep 1; done'")
+	DebugCommand []string `json:"debugCommand,omitempty"`
 }
 
 // ClowdJobInvocationStatus defines the observed state of ClowdJobInvocation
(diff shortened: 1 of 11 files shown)


apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
```diff
@@ -60,6 +60,12 @@ type ClowdJobInvocationSpec struct {
 
 	// Testing is the struct for building out test jobs (iqe, etc) in a CJI
 	Testing JobTestingSpec `json:"testing,omitempty"`
+
+	// invokes the job pods in "debug" mode -- i.e., command is swapped to the debugCommand
+	Debug bool `json:"debug,omitempty"`
+
+	// indicates pod command to run when in debug mode (default: "/bin/sh -c 'while true; do sleep 1; done'")
+	DebugCommand []string `json:"debugCommand,omitempty"`
```
bsquizz, line 68: I'm still not sure if this is something we want to give people the flexibility to do in the CJI

BlakeHolifield: I like it given that we apparently can't even rely on `cat` lol. I think this should potentially be
```suggestion
	type DebugSpec struct {
	    Enabled bool `json:"enabled"`
	    Command []string `json:"debugCommand,omitempty"`
}
```

bsquizz:
Turns out we can run `oc debug job/<name>` and it works so we're probably gonna scrap this one.

bsquizz:
Handling via #351

PR #351, opened 2021-05-27
Convert IQE CJI to use iqe-core image's entrypoint command

bsquizz:
* Don't set a command on the pod, use the iqe-tests image's entrypoint
* Pass marker/filter/etc. as env vars to the container
* Add 'debug: true' to the IqeSpec which will tell the iqe runner to sleep infinitely (so that tests do not run and you can run `oc rsh` to get into the IQE pod and debug/troubleshoot)

Goes along with https://gitlab.cee.redhat.com/insights-qe/iqe-dockerfiles/-/merge_requests/59 and https://gitlab.cee.redhat.com/insights-qe/iqe-dockerfiles/-/merge_requests/61

### DIFF
--- a/controllers/cloud.redhat.com/providers/iqe/impl.go
+++ b/controllers/cloud.redhat.com/providers/iqe/impl.go
@@ -42,6 +42,7 @@ func CreateIqeJobResource(cache *providers.ObjectCache, cji *crd.ClowdJobInvocat
 	envvar := []core.EnvVar{
 		{Name: "ENV_FOR_DYNACONF", Value: cji.Spec.Testing.Iqe.DynaconfEnvName},
 		{Name: "NAMESPACE", Value: nn.Namespace},
+		{Name: "CLOWDER_ENABLED", Value: "true"},
 		{Name: "IQE_DEBUG_POD", Value: strconv.FormatBool(cji.Spec.Testing.Iqe.Debug)},
 		{Name: "IQE_PLUGINS", Value: app.Spec.Testing.IqePlugin},
 		{Name: "IQE_MARKER_EXPRESSION", Value: cji.Spec.Testing.Iqe.Marker},
--- a/controllers/cloud.redhat.com/providers/iqe/impl.go
+++ b/controllers/cloud.redhat.com/providers/iqe/impl.go
@@ -3,9 +3,8 @@ package iqe
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
-	"strings"
+	"strconv"
 
 	crd "cloud.redhat.com/clowder/v2/apis/cloud.redhat.com/v1alpha1"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


kylape:
My thoughts on this topic.

- I prefer the self-documenting API of having explicit fields for IQE options, e.g. marker and filter
- Making it part of the API should help devs to quickly understand what options are available to configure an IQE job
- Later integrations like Ibutsu or Vault will probably need to be more tightly integrated with Clowder regardless
- I don't understand why you need a debug flag on the IQE pod.  I thought we tentatively agreed yesterday to set a filter that would exclude all tests.  Did that option not work?

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -41,10 +40,12 @@ func CreateIqeJobResource(cache *providers.ObjectCache, cji *crd.ClowdJobInvocat
 	}
 
 	envvar := []core.EnvVar{
-		{Name: "ACG_CONFIG", Value: "/cdapp/cdappconfig.json"},
 		{Name: "ENV_FOR_DYNACONF", Value: cji.Spec.Testing.Iqe.DynaconfEnvName},
 		{Name: "NAMESPACE", Value: nn.Namespace},
-		{Name: "CLOWDER_ENABLED", Value: "true"},
+		{Name: "IQE_DEBUG_POD", Value: strconv.FormatBool(cji.Spec.Testing.Iqe.Debug)},
```
BlakeHolifield, line 45: To avoid some confusion, this line controls how the debug pod starts: https://gitlab.cee.redhat.com/insights-qe/iqe-dockerfiles/-/merge_requests/59/diffs#95dd46e5eb3ea3c869a1eec9b1938e0649aadc52_13_14

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -53,20 +54,13 @@ func CreateIqeJobResource(cache *providers.ObjectCache, cji *crd.ClowdJobInvocat
...
-	if err != nil {
-		return err
-	}
-
 	j.Spec.Template.Spec.ServiceAccountName = fmt.Sprintf("iqe-%s", app.Spec.EnvName)
 
 	c := core.Container{
 		Name:         nn.Name,
 		Image:        fmt.Sprintf("%s:%s", iqeImage, tag),
-		Command:      constructedIqeCommand,
```
BlakeHolifield, line 69: And `command `is removed in favor of the entry point script in the container: https://gitlab.cee.redhat.com/insights-qe/iqe-dockerfiles/-/merge_requests/59/diffs#bf13662e10e3e9768d5f98852899f5970db729dd_4_3

BlakeHolifield (commented):
Regardless of the `debug` or `env` debates, I think moving the command builder out and using the runners script is much more neat.

psav (changes requested):
One comment here, but I'm with @kylape on this particular PR

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -41,10 +40,12 @@ func CreateIqeJobResource(cache *providers.ObjectCache, cji *crd.ClowdJobInvocat
 	}
 
 	envvar := []core.EnvVar{
-		{Name: "ACG_CONFIG", Value: "/cdapp/cdappconfig.json"},
```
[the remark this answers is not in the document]
bsquizz: I could leave it there, but I don't think it is used for the purpose of an IQE job, since IQE looks at the /cdenv/cdenvconfig.json

bsquizz:
> I prefer the self-documenting API of having explicit fields for IQE options, e.g. marker and filter
> Making it part of the API should help devs to quickly understand what options are available to configure an IQE job

Right, I didn't change the spec of a ClowdJobInvocation. I only added 'debug' to the spec.

The main change is clowder no longer builds the `command` that the pod runs. QE has an entrypoint script defined in the Dockerfile of their IQE image, and Clowder will run the pod with no command so that the default entrypoint is invoked, and just make sure to set the env vars that the entrypoint script expects

> I don't understand why you need a debug flag on the IQE pod. I thought we tentatively agreed yesterday to set a filter that would exclude all tests. Did that option not work?

The `debug` flag is used to make IQE gracefully exit 0. If you use the approach of a 'bogus filter' that we talked about, I discovered the pytest run errors with "no tests collected", and the kubernetes `Job` tries to re-run the pod 6 times. If you just `exit 0` the first time, the Job completes successfully and there's extra pods showing 'Error' status

BlakeHolifield (approved):
I think we just need a @kylape  or @psav approval on this now, but it looks good to me.

bsquizz:
> I think we just need a @kylape or @psav approval on this now, but it looks good to me.

Changing behavior of IQE entrypoint command for 'debug mode' with this PR: https://gitlab.cee.redhat.com/insights-qe/iqe-dockerfiles/-/merge_requests/61

PR #356, opened 2021-06-04
fix broken doc and image links

loadtheaccumulator:
* README.adoc
* .../pages/clowder-design
* .../pages/learn-more.adoc

Signed-off-by: Jonathan Holloway <jholloway@redhat.com>


psav:
This failed the tests because your GPG key isn't in the app-interface repo - we should get that setup - but I'll merge this anyway

PR #366, opened 2021-06-11
Use `minikube kubectl` in kube_setup.sh

Ichimonji10:
By default, use `minikube kubectl` instead of `kubectl` in
`build/kube_setup.sh`. This removes a dependency, as users no longer
need to install kubectl system-wide in order to use the script. Users
who wish to use the system-wide kubectl may still do so with little
effort.

### DIFF
--- a/build/kube_setup.sh
+++ b/build/kube_setup.sh
@@ -1,8 +1,6 @@
 #!/bin/bash
+# Configures a local minikube cluster for testing.
 set -e
-
-# Script you can use to set up a local minikube cluster for testing
-# It is assumed you have already run 'minikube start' and your kubectl context is using the minikube cluster
 
 REINSTALL=0
 
@@ -36,6 +34,18 @@ if ! command -v go; then
     exit 1
 fi
 
+# kubectl is required for interactions with the cluster.
+if [ -n "${KUBECTL_CMD}" ]; then
+    :  # already set via env var
+elif command -v kubectl; then
+    KUBECTL_CMD=kubectl
+elif command -v minikube; then
+    KUBECTL_CMD='minikube kubectl --'
+else
+    echo "*** 'kubectl' not found in path. Please install it or minikube, or set KUBECTL_CMD"
+    exit 1
+fi
+
 
 GO_BIN_PATH="$(go env GOPATH)/bin"
 
@@ -62,7 +72,7 @@ function install_strimzi_operator {
     STRIMZI_TARFILE="strimzi-${STRIMZI_VERSION}.tar.gz"
 
     if [ $REINSTALL -ne 1 ]; then
-        STRIMZI_DEPLOYMENT=$(kubectl get deployment strimzi-cluster-operator -n $STRIMZI_OPERATOR_NS --ignore-not-found -o jsonpath='{.metadata.name}')
+        STRIMZI_DEPLOYMENT=$(${KUBECTL_CMD} get deployment strimzi-cluster-operator -n $STRIMZI_OPERATOR_NS --ignore-not-found -o jsonpath='{.metadata.name}')
         if [ ! -z "$STRIMZI_DEPLOYMENT" ]; then
             echo "*** Strimzi operator deployment found, skipping install ..."
             return 0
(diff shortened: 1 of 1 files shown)


bsquizz (changes requested):
Along the lines of what @BlakeHolifield mentioned, can the logic for setting KUBECTL_CMD be this:

1. Use `KUBECTL_CMD` env var if user specified it
2. Use `kubectl` (if user has it installed, i.e. `which kubectl` works)
3. Use `minikube kubectl`

build/kube_setup.sh
```diff
@@ -1,8 +1,10 @@
 #!/bin/bash
+# Configures a local minikube cluster for testing.
 set -e
 
-# Script you can use to set up a local minikube cluster for testing
-# It is assumed you have already run 'minikube start' and your kubectl context is using the minikube cluster
+# Set to `kubectl` if kubectl is installed system-wide, its context is set to your minikube cluster,
+# and you wish to use it instead of the kubectl bundled with minikube.
+KUBECTL_CMD='minikube kubectl --'
```
psav, line 7: Doesn't this need to be something like `?=` to set it as a default but allow override if the env var is set? Plus, @bsquizz mentioned doing a check by default to see if kubectl is installed and available. I think the default should be, try to use `kubectl` from path if available, then fall back to `minikube`, and then finally if the user sets the env var, we should just override it with that.

psav (changes requested):
Some changes requested

psav:
We need to consider this a little more carefully - the `kube_setup` script was so called, because it could be used against both CRC and Minikube. We should discuss if that's still the idea and how to enable both of them.

BlakeHolifield:
> We need to consider this a little more carefully - the `kube_setup` script was so called, because it could be used against both CRC and Minikube. We should discuss if that's still the idea and how to enable both of them.

Should it be considered reasonable for a dev to have one of the Kube binaries installed; be it `oc` or `kubectl` ?  And if so, can we leave the script and document, as a dependency, how to install them. I think if devs are going to be working in a Kubernetes, cloud native environment space, the cli tools are the first step really.

build/kube_setup.sh
Ichimonji10: I think the lookup order would be:

1. Use $`KUBECTL_CMD`, and if it's unset...
2. Use `kubectl`, and if it's unavailable...
3. Use `minikube`.

Right?

Ichimonji10:
@BlakeHolifield It's up to you how high the barriers to entry should be. I don't develop services, and rarely have a reason to interact with `oc` or `kubectl`; I'm here because I'm trying to make a service run tests in ephemeral environments, not because I have any interest in developing with clowder.

build/kube_setup.sh
bsquizz: Yeah IMO that's the best path for the logic

bsquizz (approved):

bsquizz:
The tests are failing because the commits are not gpg signed

Ichimonji10:
Tested with:

```sh
# get fresh minikube deploy
minikube delete
curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-latest.x86_64.rpm
sudo rpm -Uvh minikube-latest.x86_64.rpm
minikube start --driver=kvm2 --cpus=4 --memory=8192MB
minikube kubectl -- get pods -A  # sanity check

# deploy clowder
./build/kube_setup.sh
minikube kubectl -- apply -f https://github.com/RedHatInsights/clowder/releases/download/v0.18.1/clowder-manifest-v0.18.1.yaml --validate=false
minikube kubectl -- get pods -A  # sanity check
```

Ichimonji10:
> The tests are failing because the commits are not gpg signed

I've never signed a commit with GPG. Is there an internal guide for this? Or should I use google?

bsquizz:
It's not too hard to configure git to sign your commits, but then getting your gpg key added into app-interface will be another barrier for you which I don't think is worth the hassle for your small change here. I just co-signed your commit in another branch which should do the trick, and we'll use https://github.com/RedHatInsights/clowder/pull/408 instead

PR #369, opened 2021-06-15
Add status to Env

psav:


apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
```diff
@@ -418,6 +418,7 @@ type MinioStatus struct {
 type ClowdEnvironmentStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
+	Conditions      []ClowdCondition        `json:"conditions,omitempty"`
```
BlakeHolifield, line 421: I see that the root(ish) type has changed. We can change the `status` structs as we please, correct? We aren't tied to the same rigidity as the spec?

BlakeHolifield (approved):
I don't see any issues and this is so so valuable. Do we need unit testing around these? I don't think there's a good way to test this in `kuttl`.

apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
psav: This is an addition - and it's omit-empty

kylape:
Merged manually

PR #374, opened 2021-06-18
Update Redis and Unleash to Use Quay Images

BlakeHolifield:
via: https://coreos.slack.com/archives/C022YV4E0NA/p1624040288441800

The image is rate limited from docker.io. Using our quay image instead will fix this, but will require the use of a dev pull secret from Quay. Docs for that [process are here](https://clouddot.pages.redhat.com/docs/dev/getting-started/backend-local.html#_get_your_quay_pull_secret)

Both images are not mirrored, but built and published publicly.


controllers/cloud.redhat.com/providers/inmemorydb/redis.go
```diff
@@ -125,7 +125,7 @@ func makeLocalRedis(o obj.ClowdObject, objMap providers.ObjectMap, usePVC bool,
 
 	dd.Spec.Template.Spec.Containers = []core.Container{{
 		Name:  nn.Name,
-		Image: "redis:6",
+		Image: "quay.io/cloudservices/redis:6",
```
BlakeHolifield, line 128: Defaulting this to `env.Spec...` instead

psav:
Is this ready now?

BlakeHolifield:
> Is this ready now?

What do you think? I can make it match strimzi where we look for the image first in the env spec or I can do it like this. We will have to start having folks use their own pull secrets when we make this change.

PR #387, opened 2021-07-07
Fix target namespace bug

psav:


psav:
@BlakeHolifield Can you give this a thorough checking out please?

BlakeHolifield:
Verified that the `targetNamespace` will now update in status. I deployed everything into `env-metaverse` in the `metaverse` namespace. After that came up, I edited the same env to use `oopsie` as the `targetNamespace`. The base env pods deployed to the `oopsie` namespace. The `status.targetNamespace` updated to `oopsie`. However, the pods in `metaverse` hang around after the transition to a new `targetNamespace`. Is this intended behavior?

psav:
It's not intended behaviour, but it is expected behaviour right now. There is a label issue that I REALLY want to solve - but havn't found priority to do it yet.

PR #391, opened 2021-07-09
Support for passing vault credentials to IQE ClowdJobInvocation

bsquizz:


bsquizz:
> Won't the CJI always be deployed into one of the apps namespaces, if so, the pull secret provider should copy any secret defined in the clowdenv into that namespace.

Correct it will copy the secret into the NS that the CJI is applied to.

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -125,7 +146,93 @@ func CreateIqeJobResource(cache *providers.ObjectCache, cji *crd.ClowdJobInvocat
...
+
+	if err = cache.Create(VaultSecret, dstSecretRef, vaultSecret); err != nil {
+		logger.Error(err, "Failed to add vault secret to cache")
+		return err, nil
+	}
+
+	return nil, vaultSecret
+}
+
+func addIqeSecretToCache(cache *providers.ObjectCache, ctx context.Context, cji *crd.ClowdJobInvocation, app *crd.ClowdApp, envName string, logger logr.Logger, client client.Client) error {
```
bsquizz, line 235: Renamed this function for consistency

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -125,7 +146,93 @@ func CreateIqeJobResource(cache *providers.ObjectCache, cji *crd.ClowdJobInvocat
...
+	vaultEnvVars := []core.EnvVar{
+		{Name: "DYNACONF_IQE_VAULT_LOADER_ENABLED", Value: "true"},
+		{Name: "DYNACONF_IQE_VAULT_VERIFY", Value: "true"},
+		{
+			Name: "DYNACONF_IQE_VAULT_URL",
+			ValueFrom: &core.EnvVarSource{
+				SecretKeyRef: &core.SecretKeySelector{
+					LocalObjectReference: core.LocalObjectReference{Name: vaultSecret.Name},
+					Key:                  "url",
+					Optional:             utils.BoolPtr(true),
```
bsquizz, line 160: I'm passing `optional: true` on all of these to make the code simpler on the Clowder side. This way we don't have to dig into the secret and actually read it. We'll let the IQE pod parse the env var and fall back to whatever default values it would prefer to use (or abort) if one of these env vars is empty.

BlakeHolifield (commented):
Design doc for the uninitiated: https://docs.google.com/document/d/1ZCmY_50wkoxOu-mRJF3FWluTO1083hvZ5v-GCXQVER8/edit# 

Looks good. Just one question about a goof I made a while back it seems. I think usage docs around this would be super helpful; specifically around 
> For engineers running on minikube, VAULT_ROLE_ID/VAULT_SECRET_ID can be swapped out and their own github token can be used via this env var  `DYNACONF_IQE_VAULT_GITHUB_TOKEN=<secret>` 

but that can go to clouddot.pages

BlakeHolifield (approved):

kylape (approved):

kylape:
Merged in a separate batch.

PR #392, opened 2021-07-13
Use uuid of ClowdEnvironment to generate unique kafka(connect) cluster name

bsquizz:
Strimzi locks appear to be a string composed of [cluster name and cluster namespace](https://github.com/strimzi/strimzi-kafka-operator/blob/500c846d9ec5ce877c0aa12cbad4a67fecfaf304/operator-common/src/main/java/io/strimzi/operator/common/AbstractOperator.java#L155-L157)

I suspect that repeatedly deleting/re-creating clusters with the name name+namespace combo is not helping our "failed to acquire lock" situation on the ephemeral cluster. This PR will generate a unique name for a kafka/kafkaconnect cluster by appending the start of the ClowdEnvironment's UID onto the cluster name. Since the namespace reconciler deletes/re-creates ClowdEnvironments when it prepares namespaces, every time a namespace is prepared it should get a new, fully unique kafka cluster name.

I'm also removing the `-connect` from our auto-generated KafkaConnect cluster names since it is redundant. The `kafkaconnect` type differentiates a connect cluster from a kafka cluster, and the pods appear in the namespace with `name-connect-connect` which I think is ugly though I'm being pedantic :P

So in summary this PR will:
* generate a unique name using ClowdEnvironment UID if a kafka/connect cluster name is not provided in the spec
* if auto-generating the name of the connect cluster, just make it the same as the kafka cluster name (drop `-connect`)
* fix kuttl tests to work w/ this new model. `test-kafka-strimzi-ephemeral` tests the name auto-generation using a unique UID. For all other strimzi tests hard code a cluster name to retain their "prior function". 
* shorten name of the `test-kafka-strimzi-kafka-provisioner-ephemeral` test because the cluster name was too long and was hitting some k8s label length limits. I went ahead and shortened the name of the 'pvc' test in a similar way.


bsquizz:
kuttl tests failing with `invalid label value: "env-test-kafka-strimzi-kafka-provisioner-ephemeral-0cb03bcc-kafka": at key: "strimzi.io/name": must be no more than 63 characters` -- need to shorten the name of the env in some of these tests. I'll also need to fix some other kuttl tests which have assertions that will fail since they were relying on the auto-generated cluster name.

PR #394, opened 2021-07-14
Working annotation db implementation

psav:


kylape (changes requested):
Please prefix the annotation key with `clowder/` similar to https://github.com/RedHatInsights/clowder/blob/8b9a8c4759a434f8c3fc7a2f78251693d024cf3c/controllers/cloud.redhat.com/providers/objectstore/appinterface.go#L109

kylape (approved):

PR #395, opened 2021-07-15
update strimzi client

Jason-RH:


Jason-RH:
@psav Not sure if the update to the strimzi client did what we expected? Unit tests are failing with what (I think) looks like a similar error

psav:
Looking at what I did in the Kafka Metrics piece - I think we are going to have to do something similar here - see the definition of it as a apiextensions.JSON. And the subsequent unmarshal

```	if s.Env.Spec.Providers.Kafka.Cluster.Config != nil {
		k.Spec.Kafka.Config = s.Env.Spec.Providers.Kafka.Cluster.Config
	}

	k.Spec.Kafka.JvmOptions = &s.Env.Spec.Providers.Kafka.Cluster.JVMOptions

	var metrics apiextensions.JSON

	metrics.UnmarshalJSON(metricsData)

	metricsConfig := strimzi.KafkaSpecKafkaMetricsConfig{
		Type: "jmxPrometheusExporter",
		ValueFrom: strimzi.KafkaSpecKafkaMetricsConfigValueFrom{
			ConfigMapKeyRef: &strimzi.KafkaSpecKafkaMetricsConfigValueFromConfigMapKeyRef{
				Key:      common.StringPtr("metrics"),
				Name:     common.StringPtr(cmnn.Name),
				Optional: common.FalsePtr(),
			},
		},
	}
```

PR #398, opened 2021-07-16
Update getting started docs to include latest bonfire commands

bsquizz:
We had a few engineers try to run through these last week and of course they weren't too successful -- these docs are pretty out of date and needed a refresh.


kylape (approved):

kylape:
Merged in a separate batch.

PR #399, opened 2021-07-20
Use correct service account for redis/db deployments

kylape:


BlakeHolifield (approved):

kylape:
Merged in a separate batch.

PR #404, opened 2021-07-23
Changed error message when db name is not found in secrets

quarckster:
I encountered a completely irrelevant error message during deploying a clowder application with a database:
```
Missing dependencies: [- database: <app name>]
```

I think the error message should be more clear and explains what exactly happened. Besides, it should give some idea how to resolve it.

### DIFF
--- a/controllers/cloud.redhat.com/providers/database/appinterface.go
+++ b/controllers/cloud.redhat.com/providers/database/appinterface.go
@@ -52,7 +52,7 @@ func fetchCa() (string, error) {
 	return caBundle, nil
 }
 
-// NewAppInterfaceDBProvider creates a new app-interface DB provider obejct.
+// NewAppInterfaceDBProvider creates a new app-interface DB provider object.
 func NewAppInterfaceDBProvider(p *providers.Provider) (providers.ClowderProvider, error) {
 	provider := appInterface{Provider: *p}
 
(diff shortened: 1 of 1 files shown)


bsquizz:
I have some across a similar issue before of wanting some more context around a `MissingDependencies` error -- but there's a reason the `MissingDependencies` error is returned and structured as a `map[string][]string` (the reason why, I can't exactly remember at the moment, it is used somewhere else in the code)

@psav what do you think is the best way to share more context around these kinds of errors back to the user? Should we re-structure `MissingDeps` to be a `[]MissingDependency` where `MissingDependency` is something like:

```
type MissingDependency struct {
    type string  // i.e. "database"
    appName string  // the ClowdApp that has the missing dependency
    errorDetails string  // custom error msg that gives more info
}
```

and append `errorDetails` to the event that gets emitted?

psav:
Yeh -it's implemented like this to allow us to aggregate missing deps from multiple places, I think we can enhance this for sure. @bsquizz I like your idea

bsquizz:
RHCLOUD-15674 will track this

psav:
Will these changes be made in this PR @bsquizz or hsould we close this off in favour of a new one?

bsquizz:
Will address this as a new feature

PR #408, opened 2021-07-27
Use a configurable KUBECTL_CMD in kube_setup.sh

bsquizz:
In kube_setup.sh use:
* `KUBECTL_CMD` if set, else
* `kubectl` if found on path, else
* `minikube kubectl`


Ichimonji10:
Derived from: #366

psav (approved):

PR #410, opened 2021-07-29
Add pvc check to switch to recreate strategy

psav:


kylape:
Just gotta add tests and documentation

PR #411, opened 2021-07-29
Added disabled flags

psav:


bsquizz:
A kuttl test for this might be feasible, but will take some work to figure out since it's kind of an odd flow for kuttl -- I'll manually test it for now.

psav:
it wouldn't be too hard to write as a kuttl test - you can just set it up, validate it exists, set the disabled flag, make a change to the app, validate that the deployment remained the same, flip the disabled flag off, and validate the deployment changed.

bsquizz:
Wrote a kuttl test for what I think will be the common use case... reducing the deployment of a ClowdEnv's DB deployment to 0 replicas and reducing the deployment of a ClowdApp owned Deployment to 0 replicas.

The test is:
* start w/ a standard env + an app that has a DB, assert app deployment and DB deployment appear with `readyReplicas=1`
* set `disabled: true` on the env and the app
* set replicas on the DB deployment and app deployment to 0
* via an 'errors' test step I'm asserting that `status.readyReplicas` is not present with value=1 (turns out this field is totally absent from the Deployment's status if replicas is set to 0)
* set `disabled: false` on the env and app
* assert that the DB and app deployments once more go back to `readyReplicas=1` once `disabled: false` is re-set on the Clowd

bsquizz:
Added another test to ensure that deleting a ClowdApp/Env with `disabled: true` still causes the Deployments owned by those objects to be removed

BlakeHolifield (approved):

PR #414, opened 2021-08-06
add DeploymentStrategy to Deployment spec

maskarb:
This PR is one idea to control the deployment strategy for deployments. It adds a `deploymentStrategy` field to the `deployments`:
```
    minReplicas: 5
    deploymentStrategy:
        type: RollingUpdate
        rollingUpdate:
          maxUnavailable: 50%
          maxSurge: 50%
```

Adding this section gives us more control over how many pods spin up during deployment. If using these values:
```
    minReplicas: 5
    deploymentStrategy:
        type: RollingUpdate
        rollingUpdate:
          maxUnavailable: 0
          maxSurge: 1
```
this would deploy 1 new pod while keeping the original 5 in place. Then once, the new pod becomes ready, an old one terminates, and another new one spins up.

TODO:
- needs test cases


psav:
Just thinking aloud here would it be more beneficial to describe these just as percentages and change our spec to supply the two percentages, just so that if we implement scaling at some point, ie....between local/ephemeral/prod, we don't have absolute values?

maskarb:
I went this route because:
1. It was very easy. I just took an already valid definition of `deploymentStrategy` and stuffed it into the CRD.
2. This gives us the ability to change the `type` to `Recreate` (if we wanted). This is something we can't do today
3. Percentages lead to some fuzzy math, so it would be nice to be able to supply either percentages or absolutes.

Our team really needs the ability to specify how many pods start during a new deployment. We want 1 pod so that it will run our init-container which is responsible for running our migrations. Now, granted, if we had 20 pods and we were to upgrade, it would take a _loooong_ time to switch over the new deployment (if our unavailable/surge was set to 1/0), but we aren't close to having more than 5 pods running any time soon.

I have thought about other ways we can accomplish running migrations using a one off Job, but it's been quite confusing thinking through how that will work in practice. The changes proposed here will give us a working solution quickly.

I think in the long run, we will want to go a different route, like something along the lines of a `deploymentJob`. But we're still iterating on how our deployment-process will work and thinking of what changes we need to make to get to that new deployment-process.

psav:
@kylape Need your input on this one

psav:
Moved to this implementation https://github.com/RedHatInsights/clowder/pull/546

PR #417, opened 2021-08-06
We should we waiting for the stateful set instead as sometimes a race condition can occur and the pod doesn't exist causing kubectl wait to exit early and never return.

skarekrow:
Verified on a call with Squizz


bsquizz (approved):

bsquizz:
Using #423 so we can fix the pr check failure at the same time.

PR #418, opened 2021-08-07
Added PW for elasticache

bennyturns:
Signed-off-by: bennyturns <bturner@redhat.com>


Jason-RH:
nitpick, but a common pattern to check the existence of a key and operate on it would be:
`if val, ok := dict["foo"]; ok {
    //do something here
}`

bsquizz (approved):

Jason-RH (approved):

psav (approved):

BlakeHolifield (approved):

PR #430, opened 2021-08-26
CJIs now implement conditions

BlakeHolifield:
* Each cji now has a `conditions` block under `status`
* Jobs are now tracked in a map along with their progress as either `Invoked`, `Complete`, or `Failed`


apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
```diff
@@ -169,3 +183,13 @@ func (i *ClowdJobInvocation) SetObjectMeta(o metav1.Object, opts ...omfunc) {
 		opt(o)
 	}
 }
+
+func (i *ClowdJobInvocation) GetInvokedJobs(ctx context.Context, c client.Client) (batchv1.JobList, error) {
+
+	jobs := batchv1.JobList{}
+	if err := c.List(ctx, &jobs, client.InNamespace(i.ObjectMeta.Namespace)); err != nil {
+		return jobs, err
```
psav, line 191: If you make the return a pointer to a JobList, you can return nil here, which I think makes sense from a readability PoV and could make things a littler more optimised too

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -30,7 +30,14 @@ func CreateIqeJobResource(cache *providers.ObjectCache, cji *crd.ClowdJobInvocat
 	labels := cji.GetLabels()
 	cji.SetObjectMeta(j, crd.Name(nn.Name), crd.Labels(labels))
 
+	// Becuase the ns name now has a suffix attached, we need to specify
+	// that the secret name does not include it (and can't because the
+	// env creates the secret)
+	secretName := fmt.Sprintf("%s-iqe", cji.Name)
```
psav, line 36: Why not make this a method on the CJI object?

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -170,55 +168,57 @@ func (r *ClowdJobInvocationReconciler) Reconcile(ctx context.Context, req ctrl.R
 	var emptyTesting crd.IqeJobSpec
 	if cji.Spec.Testing.Iqe != emptyTesting {
 
+		randomString := utils.RandStringLower(7)
 		nn := types.NamespacedName{
-			Name:      fmt.Sprintf("%s-iqe", cji.Name),
+			Name:      fmt.Sprintf("%s-iqe-%s", cji.Name, randomString),
```
psav, line 173: We could have another method on the CJI to generate a name, I would specifically call it "generate" so that people know you can't just recall it with a call to the function

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -298,51 +299,58 @@ func (r *ClowdJobInvocationReconciler) cjiToEnqueueUponJobUpdate(a client.Object
...
 		// job event triggered a reconcile, check our jobs and match
 		// to enable a requeue
-		if contains(cji.Status.Jobs, job.ObjectMeta.Name) {
-			reqs = append(reqs, reconcile.Request{
-				NamespacedName: types.NamespacedName{
-					Name:      cji.Name,
-					Namespace: cji.Namespace,
-				},
-			})
+		for j, _ := range cji.Status.Jobs {
```
psav, line 302: Can drop the `, _` right?

psav (changes requested):

PR #438, opened 2021-09-06
Update strimzi

psav:


psav:
Hoping this could make tests more stable

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -810,6 +812,8 @@ func processTopicValues(
 		}
 	}
 
+	jsonData := "{"
```
bsquizz, line 815: Haha uh oh, if this works then it works but why do we have to build a string like this?

bsquizz: Alright well, my attempt at using `apiextensions.JSON` instead of `apiextensions/v1.JSON` failed so I guess we've just gotta stick with building the strings like this

PR #444, opened 2021-09-14
Downloading fix_namespaces.py if missing

Jason-RH:


build/kube_setup.sh
```diff
@@ -93,7 +94,14 @@ function install_strimzi_operator {
     [[ $PLATFORM == "Darwin" ]] && sed -i '' "s/namespace: .*/namespace: ${STRIMZI_OPERATOR_NS}/" *RoleBinding*.yaml \
         || sed -i "s/namespace: .*/namespace: ${STRIMZI_OPERATOR_NS}/" *RoleBinding*.yaml
 
-    $ROOT_DIR/build/fix_namespace.py 060-Deployment-strimzi-cluster-operator.yaml "$WATCH_NS"
+    if ! test -f $ROOT_DIR/build/${FIX_NAMESPACE_SCRIPT}; then
```
bsquizz, line 97: Should we just always download it?

Otherwise the other problem we run into is: could someone have a very old `fix_namespace.py` sitting in the dir that never gets updated? :)

bsquizz:
In case anyone is wondering why this might be needed:

Some are instructed on how to install clowder onto minikube *without* having to git clone this entire repo, e.x.: https://consoledot.pages.redhat.com/docs/dev/getting-started/local/environment.html#_install_clowder

build/kube_setup.sh
psav: Agreed on this

psav (changes requested):
The one change I see is inline with what @bsquizz  said.

PR #451, opened 2021-09-23
[RHCLOUD-15756] Add tee-shirt-sizing for DB

psav:


psav:
Failing test was unrelated

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -56,6 +60,9 @@ type DatabaseSpec struct {
 
 	// Defines the Name of the app to share a database from
 	SharedDBAppName string `json:"sharedDbAppName,omitempty"`
+
+	// T-shirt size, one of small, medium, large
+	DBTShirtSize TShirtSize `json:"dbTShirtSize,omitempty"`
```
bsquizz, line 65: I am nitpicking here. But could we drop `TShirt` from all the naming to make the field names and code a little cleaner? :) For example could this just be `dbSize` or `volumeSize`? 

Could `TShirtSizeDefinitions` just become `SizeDefinitions`?

psav: I went for `TShirt` because we are using the small, medium, large terminology.

bundle/tests/scorecard/kuttl/test-db-pvc-tee/01-pods.yaml
```diff
@@ -0,0 +1,56 @@
...
+spec:
+  envName: test-db-pvc-tee
+  deployments:
+  - name: processor
+    podSpec:
+      image: quay.io/psav/clowder-hello
+  database:
+    name: app-a
+    version: 12
+    pvc: true
```
BlakeHolifield, line 55: Do we want to add any testing around setting `pvc: false` and trying to have a size enabled? We'll need to update the provider docs as well. I don't have a strong opinion on TShirtSize vs DBSize or VolumeSize. Looks good otherwise!

BlakeHolifield (approved):

psav:
OK @BlakeHolifield I ran this and found two things - 1) the shared db only ever uses the default size - I think what it hsould do is look at all the clowdapps that are trying to use the same db and take the max
2) pvc mode is never set on the clowdenv so the pvc is never created, the pvc was put on the clowdapp, which of course isn't the spec :)

BlakeHolifield:
Closing due to wacky rebase errors. Reopened in #574

PR #452, opened 2021-09-24
Add requirements env vars for IQE CJI

bsquizz:
This adds 3 env vars to the IQE CJI pod that the QE team needs for requirements-based testing.

The iqe_runner within the pod will parse these, Clowder's job is to just pass the string along and set the env var on the pod.

See https://iqe-docs.apps.ocp4.prod.psi.redhat.com/metadata.html#added-functionality for more info.


bsquizz:
`make api-docs` is altering a large portion of `docs/antora/modules/ROOT/pages/api_reference.adoc` -- I think after this goes in we might want to do a separate PR to sync docs.

psav:
What's the need for doing it in a separate PR? If we get a conflict - we just regenerate it, and because it's auto generated a long review of it isn't really necessary as it's just adoc.

psav (approved):
I'm happy - just needs arch review of spec change

kylape (approved):

BlakeHolifield (approved):

bundle/tests/scorecard/kuttl/test-iqe-jobs/05-assert-with-default.yaml
```diff
@@ -48,6 +48,12 @@ spec:
               value: "smoke"
             - name: IQE_FILTER_EXPRESSION
               value: "test_plugin_accessible"
+            - name: IQE_REQUIREMENTS
+              value: ''
```
BlakeHolifield, line 52: I think these are being asserted from a previous job hanging around with the vars filled out. You can update these to match the other tests, or update `05-pods-with-default.yaml` to check the empty values.

bsquizz:
> What's the need for doing it in a separate PR? If we get a conflict - we just regenerate it, and because it's auto generated a long review of it isn't really necessary as it's just adoc.

Because the doc update is updating a bunch of stuff not related to this PR -- but we can handle it in this PR if you want.

bsquizz:
Requirements doc updated here to document spec change: https://docs.google.com/document/d/1ZCmY_50wkoxOu-mRJF3FWluTO1083hvZ5v-GCXQVER8/edit#heading=h.j96vfzgv1nhs

Ran `make api-docs` and pushed via commit https://github.com/RedHatInsights/clowder/pull/452/commits/9add20757a792fc6fe0188588adcc068ecf8123e -- take a look

apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
```diff
@@ -34,17 +34,31 @@ type IqeJobSpec struct {
...
 	DynaconfEnvName string `json:"dynaconfEnvName"`
+
 	// sets pytest -k args
 	Filter string `json:"filter,omitempty"`
+
 	// used when desiring to run `oc debug`on the Job to cause pod to immediately & gracefully exit
 	Debug bool `json:"debug,omitempty"`
+
+	// sets value passed to IQE '--requirements' arg
+	Requirements string `json:"requirements,omitempty"`
```
psav, line 55: Thinking further about this, and seeing the tests, I'm unsure if this API is sufficient. Though I don't want to burden Clowder with having to do translation, I'm unsure if I like the API for this being to pass in escaped python sets. Passing in just CSV as a string would be OK, but passing in an array of strings would be nicer. As the IQE parameters are "probably" unlikely to change, I feel it could be worth the effort to make the API nice here. At the very least we should document the format that is expected. If the goal is to get more developers to start using IQE tests and CJI is one of the interfaces to do that I think we should take the time to look at this a little more carefully.

bundle/tests/scorecard/kuttl/test-iqe-jobs/01-pods.yaml
```diff
@@ -77,4 +77,6 @@ spec:
       marker: "smoke"
       dynaconfEnvName: "clowder_smoke"
       filter: "test_plugin_accessible"
-
+      requirements: "{\"some\",\"requirements\"}"
```
psav, line 80: Could this be expressed with single quotes to avoid the escaping, just like is used in the assertions?

psav (changes requested):
Sorry @bsquizz but I think we should look at the API again briefly before merging this.

bundle/tests/scorecard/kuttl/test-iqe-jobs/01-pods.yaml
bsquizz: Yeah, I did that on purpose for variety to make sure in the assertion it came out as expected :)

apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
bsquizz: TBH I don't like this syntax either, it is how the IQE CLI arguments are formed (see https://iqe-docs.apps.ocp4.prod.psi.redhat.com/metadata.html#added-functionality). IMO, they should make this cleaner on the IQE side and just pass in simple comma,separated,strings or something like that. I wanted us to keep it a string on the Clowder side because then we are free to change the format of the arg on the IQE side without impacting CJI's. I think I'll push for nicer syntax over on IQE... thoughts?

bsquizz:
The last two test failures appear unrelated

apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
psav: The thing I'm thinking here is that if we make it a simple string, then it will "change" without warning for the end user if it is ever updated. I know you are saying this gives you the ability to freely change on the IQE side, but that change would doubtlessly introduce breakage on everyone currently using it. Unless it was done in a backwards compatible way. That seems slightly harder to achieve with a string, than it would with keeping our API consistent and translating. Another point on the "changing" it would actually be far better to introduce a _new_ parameter so that backwards compatibility can be kept. Again, if we did the translation, this would allow us to keep our API clean, and internally move over to the new way once it was established, deprecating the old way safely before the options are removed in IQE.

Thoughts?

bsquizz: But something to consider is that all these env vars get filtered thru the `iqe_runner.sh` which also has the ability to parse/convert the env vars into whatever format the latest iqe-core expects. That script is released within the iqe-core docker image so they go together "1-to-1". How about we make these fields use type `[]string` and set a CSV env var?

So for example:
```yaml
requirements:
- req1
- req2
- req3
```

Would cause Clowder to set IQE_REQUIREMENTS on the pod to `"req1,req2,req3"` -- and @psegedy will edit `iqe_runner.sh` to parse the CSV env var and run iqe with the argument formatted in the right way (which right now needs to look like `{"req1", "req2", "req3"}` even though I am not a fan of this CLI format and hope it can soon be changed to plain CSV)

If IQE changes the way that argument is formatted, we would only update iqe_runner.sh

bsquizz: See https://github.com/RedHatInsights/clowder/pull/452/commits/95e4040b1b927a2f8ea1280ff4954a54c74f0499

bsquizz: Well, ok, can't do the 'empty' struct comparison using slices, switched to using pointers to a slice. https://github.com/RedHatInsights/clowder/pull/452/commits/733bfff3b39e3917af83d84b734535d22a650ccf

bundle/tests/scorecard/kuttl/test-iqe-jobs-vault/02-delete.yaml
```diff
@@ -11,3 +11,29 @@ delete:
...
   kind: ClowdEnvironment
   name: test-iqe-jobs-vault
+
+
+
+apiVersion: cloud.redhat.com/v1alpha1
+kind: ClowdJobInvocation
+metadata:
+ name: host-inventory-smoke
+ namespace: test-iqe-jobs
```
psav, line 21: I think perhaps this is in the wrong place and is the reason for the failure of the test, it tries to delete a CJI from a namespace that is for another test.

psav (changes requested):

psav:
> > What's the need for doing it in a separate PR? If we get a conflict - we just regenerate it, and because it's auto generated a long review of it isn't really necessary as it's just adoc.
> 
> Because the doc update is updating a bunch of stuff not related to this PR -- but we can handle it in this PR if you want.

So is this a failure of running the doc generator before on another PR? Or are you talking about the fact that it reorganises things in the file because of the way it does the ordering?

bsquizz:
> > > What's the need for doing it in a separate PR? If we get a conflict - we just regenerate it, and because it's auto generated a long review of it isn't really necessary as it's just adoc.
> > 
> > 
> > Because the doc update is updating a bunch of stuff not related to this PR -- but we can handle it in this PR if you want.
> 
> So is this a failure of running the doc generator before on another PR? Or are you talking about the fact that it reorganises things in the file because of the way it does the ordering?

Yes I think it was just a case of past PR's not getting the docs updated

bundle/tests/scorecard/kuttl/test-iqe-jobs-vault/02-delete.yaml
bsquizz: Yup, this was an accidental edit

PR #454, opened 2021-09-27
[RHCLOUD-16216] Update k8s batch/v1beta1 to batch/v1

BlakeHolifield:
A good bit going on here, so I'll summarize:

- `CronJob` was promoted to `batch/v1` in `k8s.io/api v0.21.0`.
-  `batch/v1beta1` is being removed in `v0.25.0`. So an API package update was needed. I went ahead and jumped to the latest on `v0.22.0`
- Because of this update, `go1.15` is no longer sufficient and at least `go1.16` is needed. 
- `github.com/go-logr/logr` had to be downgraded from `1.1.0 => 0.4.0` [because of this controller-runtime issue](https://github.com/kubernetes-sigs/controller-runtime/issues/1607), but this is a small upgrade from our current version. Latest won't work yet.

### DIFF
--- a/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
+++ b/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
@@ -19,7 +19,7 @@ import (
 
 	"github.com/RedHatInsights/clowder/apis/cloud.redhat.com/v1alpha1/common"
 	keda "github.com/kedacore/keda/v2/api/v1alpha1"
-	batch "k8s.io/api/batch/v1beta1"
+	batch "k8s.io/api/batch/v1"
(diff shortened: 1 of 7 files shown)


psav:
Starting to shape up - but we can't merge this until UBI8/RHEL8 have the go-toolset for glolang version 1.16.

BlakeHolifield:
Okay I'm gonna put a pin in this for now. We're relying on `ubi8/go-tools:1-16` to get released and I'm not sure when that will happen. I'll dig into the Ubuntu `$GOPATH` stuff when that happens. We have until k8s 1.25. Given that 1.23 is releasing in Q4 2021 I think we have plenty of time.

psav:
@BlakeHolifield I rebased this today and it seems to be good

PR #483, opened 2021-10-29
(needs testing)[RHCLOUD-16525] Use default image pull policy

vkrizan:
Removes ImagePullPolicy on Deployment, Job and CronJob providers in
order to leverage defaults image pull policy.

The default policy is set based on the image tag usage:
* no tag or :latest -> PullAlways
* any other tag -> PullIfNotPresent

See
https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting

This improves a workflow when a deployment uses a single image tag (e.g.
latest) that is constantly being updated during development.

Ref:
* https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy

**This needs testing.** Once I get familiar with the build/test/deploy process I'll update it. Anyhow feel free to help testing this. 
Thank you!

### DIFF
--- a/controllers/cloud.redhat.com/providers/cronjob/impl.go
+++ b/controllers/cloud.redhat.com/providers/cronjob/impl.go
@@ -75,7 +75,6 @@ func buildPodTemplate(app *crd.ClowdApp, env *crd.ClowdEnvironment, pt *core.Pod
 			Name:          "metrics",
 			ContainerPort: env.Spec.Providers.Metrics.Port,
 		}},
-		ImagePullPolicy: core.PullIfNotPresent,
 	}
 
 	// set service account for pod
--- a/controllers/cloud.redhat.com/providers/deployment/impl.go
+++ b/controllers/cloud.redhat.com/providers/deployment/impl.go
@@ -100,7 +100,6 @@ func initDeployment(app *crd.ClowdApp, env *crd.ClowdEnvironment, d *apps.Deploy
 		Env:             envvar,
 		Resources:       ProcessResources(&pod, env),
 		VolumeMounts:    pod.VolumeMounts,
-		ImagePullPolicy: core.PullIfNotPresent,
 	}
 
 	if (core.Probe{}) != livenessProbe {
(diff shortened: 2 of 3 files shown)


psav:
This goes against app-sre policy and was something instigated on purpose. Where are you hoping to leverage this feature?

vkrizan:
@psav 
> Where are you hoping to leverage this feature?

We were developing a new service and were trying it out in EE using just the `latest` tag. Even if we have had a build system/scripts that would tag it with a short commit hash, it would not dismiss the fact that anyone can make a code change w/o a commit deploying it on EE. In our case we were getting some randomly cached image that was previously pulled.

>This goes against app-sre policy 
Do you have some reference? This does not change a regular deployments via AppSRE where a non-latest tag is used -- for those it is still `PullIfNotPresent`.
Alternatively we can make this configurable either in the ClowdEnv and/or within the deployment.

kylape:
I did a quick search to see if this policy was documented in any App SRE repos, but I couldn't find it.  I know we had talked about it before, but that's not very helpful now.

App SRE requirements aside, I think this will require a more in-depth discussion.  We want to tightly control what kind of image tags are allowed on stage and prod for sure, but also in PR checks.

It's not hard to always push a new image tag every time you rebuild (e.g. `$ uuid | cut -f 1 -d -`), but I do understand the convenience of reusing `latest` when doing local development or rapid development in EEs.

bsquizz:
If you all do decide to permit ImagePullPolicy to 'Always' in some cases, I just want to chime in and point out that because the Deployments do not have an ImageChangeTrigger on them (like our old DeploymentConfigs did), pushing the new 'latest' tag into the k8s docker registry won't cause the app to roll out with the newest image automatically. You'll still have to run `kubectl rollout restart deployment <name>`

psav:
> If you all do decide to permit ImagePullPolicy to 'Always' in some cases, I just want to chime in and point out that because the Deployments do not have an ImageChangeTrigger on them (like our old DeploymentConfigs did), pushing the new 'latest' tag into the k8s docker registry won't cause the app to roll out with the newest image automatically. You'll still have to run `kubectl rollout restart deployment <name>`

Good to know - thanks

vkrizan:
@bsquizz indeed, thanks! I guess we are used to restarting services when making local changes.

psav:
Superseded by https://github.com/RedHatInsights/clowder/pull/534

vkrizan:
Correction, superseded by #536

PR #485, opened 2021-11-01
Ensure cyndi has created hbi table

kylape:
This will prevent apps from crashing if they try to read from a table
that doesn't yet exist in their db.


BlakeHolifield (approved):

bennyturns:
This LGTM

PR #488, opened 2021-11-03
Remove Partial Reconciliation

psav:
This PR removes partial reconciliation completely. This was originally introduced to try to circumvent the issue of requiring some resources to be created and acted upon by other operators during a single reconciliation event. Taking the example of deploying Kafka. We require a `KafkaUser` resource to be created, and from strimzi to populate it with a `Status.Username` field. However it is not a good idea to sit and wait for this to happen as:
* The controller is single threaded so this will hold up all other reconciliations of this type
* We do not know if it will successfully complete
To combat this the idea of a partial reconciliation was introduced to allow all the resources up until that point to be written out. The problem with this approach is that if the reconciliation is bombed halfway through, some resources do not receive all their updates, meaning that when a reconciliation IS complete, more change are made, sometimes bouncing a pod.

This PR removes that process and instead INSTANT writes certain resources that are required, and then FAILS the reconciliation. This means nothing is written out until a reconciliation succeeds.

### DIFF
--- a/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
+++ b/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
@@ -340,8 +340,6 @@ const (
 	DeploymentsReady ClowdConditionType = "DeploymentsReady"
 	// ReconciliationSuccessful represents status of successful reconciliation
 	ReconciliationSuccessful ClowdConditionType = "ReconciliationSuccessful"
-	// ReconciliationPartiallySuccessful means the reconciliation is in a partial success state
-	ReconciliationPartiallySuccessful ClowdConditionType = "ReconciliationPartiallySuccessful"
 	// ReconciliationFailed means the reconciliation failed
 	ReconciliationFailed ClowdConditionType = "ReconciliationFailed"
 	// JobInvocationComplete means all the Jobs have finished
@@ -496,7 +494,15 @@ func (i *ClowdApp) GetCronJobNamespacedName(d *Job) types.NamespacedName {
 
 // IsReady returns true when all the ManagedDeployments are Ready
 func (i *ClowdApp) IsReady() bool {
-	return (i.Status.Deployments.ManagedDeployments == i.Status.Deployments.ReadyDeployments)
+	conditionCheck := false
+
+	for _, condition := range i.Status.Conditions {
+		if condition.Type == ReconciliationSuccessful {
+			conditionCheck = true
+		}
+	}
+
+	return i.Status.Ready && conditionCheck
 }
 
 // GetClowdSAName returns the ServiceAccount Name for the App
--- a/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
+++ b/apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
@@ -492,7 +492,7 @@ func (i *ClowdApp) GetCronJobNamespacedName(d *Job) types.NamespacedName {
 	}
 }
 
-// IsReady returns true when all the ManagedDeployments are Ready
+// IsReady returns true when all deployments are ready and the reconciliation is successful
 func (i *ClowdApp) IsReady() bool {
 	conditionCheck := false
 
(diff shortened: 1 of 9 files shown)


controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -65,9 +65,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-var mu sync.RWMutex
```
kylape, line 68: The mutex stuff was from 83fc721 "Stop competition: App vs Env". Looks like it was added to ensure that a given app's environment was not being actively reconciled when the app was also being reconciled.  I think we should keep this in.

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -496,7 +494,15 @@ func (i *ClowdApp) GetCronJobNamespacedName(d *Job) types.NamespacedName {
 
 // IsReady returns true when all the ManagedDeployments are Ready
```
kylape, line 495: Is this comment true anymore?  Even if it is, maybe it would be more accurate to say we are just delegating to the ready bool, assuming reconciliation has been successful

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -164,7 +159,7 @@ func (r *ClowdAppReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	if err != nil {
 		r.Recorder.Eventf(&app, "Warning", "ClowdEnvMissing", "Clowder Environment [%s] is missing", app.Spec.EnvName)
 		SetClowdAppConditions(ctx, r.Client, &app, crd.ReconciliationFailed, err)
-		return ctrl.Result{Requeue: true}, err
+		return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 10}, err
```
kylape, line 162: Why have you decided to start requeueing after 10s now?

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -374,7 +373,8 @@ func (r *ClowdAppReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&core.ConfigMap{}).
 		WithEventFilter(ignoreStatusUpdatePredicate(r.Log, "app")).
 		WithOptions(controller.Options{
-			RateLimiter: workqueue.NewItemExponentialFailureRateLimiter(time.Duration(500*time.Millisecond), time.Duration(60*time.Second)),
+			RateLimiter:             workqueue.NewItemExponentialFailureRateLimiter(time.Duration(500*time.Millisecond), time.Duration(60*time.Second)),
+			MaxConcurrentReconciles: 4,
```
kylape, line 377: As discussed, we were going to save concurrency for another PR.

kylape (changes requested):

controllers/cloud.redhat.com/clowdapp_controller.go
psav: Just to slow things down a little and give a little more time for things to "recover"/steady state

kylape: I understand this may cut down on failed reconciliations, but is that worth the likely slow down from getting an environment reconciled?

bennyturns: @kylape IMO yes it is.  We need stability and if we take a perf hit so be it, we can always optimize later when things have calmed down.  Happy to open a jira for the optimization if we want it in the future.

kylape (approved):

controllers/cloud.redhat.com/clowdapp_controller.go
kylape: Pete and I dsicussed this, and we decided to compromise by reducing the initial requeue delay to 2s.

PR #492, opened 2021-11-09
Fix bad reconcile events

psav:


controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -235,6 +233,13 @@ func (r *ClowdAppReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		managedAppsMetric.Set(float64(len(managedApps)))
```
kylape, line 233: You can remove the if around this code and move it after the `if != nil` block

kylape (approved):
Tiny bit but otherwise looks reasonable

PR #503, opened 2021-11-11
Better write now support

psav:
Instead of deciding at update time if a resource should be written now, we now declare it on the resource type.

### DIFF
--- a/controllers/cloud.redhat.com/providers/resource_cache.go
+++ b/controllers/cloud.redhat.com/providers/resource_cache.go
@@ -37,6 +37,11 @@ type ResourceIdent interface {
 	GetProvider() string
 	GetPurpose() string
 	GetType() client.Object
+	GetWriteNow() bool
+}
+
+type ResourceOptions struct {
+	WriteNow bool
 }
 
 // ResourceIdent is a simple struct declaring a providers identifier and the type of resource to be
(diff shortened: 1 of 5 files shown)


controllers/cloud.redhat.com/providers/resource_cache.go
```diff
@@ -121,22 +136,34 @@ func registerGVK(obj client.Object) {
 }
 
 // NewResourceIdent is a helper function that returns a ResourceIdent object.
-func NewSingleResourceIdent(provider string, purpose string, object client.Object) ResourceIdentSingle {
+func NewSingleResourceIdent(provider string, purpose string, object client.Object, opts ...ResourceOptions) ResourceIdentSingle {
 	registerGVK(object)
+	writeNow := false
+	for _, opt := range opts {
```
kylape, line 142: maybe i'm not familiar with this code construct, but it feels like this allows for bizarre behavior where multiple options are passed in and you just take the value of the last one?

kylape (approved):

controllers/cloud.redhat.com/providers/resource_cache.go
psav: I could wrap around saying only 1 is allowed to be passed

psav: It's a valid concern

psav: This is actually how k8s does it, if you supply two ListOptions it just loops through them one by one. I knew I'd seen the pattern somewhere - so yes thre is the potential confusion, but this seems to be standard pattern.

PR #511, opened 2021-11-15
[RHCLOUD-15296] Pre-Hook Jobs

lindgrenj6:
Adding support for a "pre-hook" job that basically runs a job and then proceeds to rollout the deployment afterwards.

\# TODO:
- [x] build out watcher for job that stores state in annotation `clowder/pre-hook-state` and `clowder/pre-hook-generation`
- [x] keep track of generation of last ran job and only run job if generation changes
- [ ] display deployment status in the k8s status 
- [x] don't run other provider if prehook status isn't completed
- [ ] make clowder NOT DELETE the job, for some reason its nuking it right now. not sure why.

Lots to do yet. Ignore unless you want your eyes to bleed.


apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -540,6 +544,26 @@ func (i *ClowdApp) GetClowdSAName() string {
 	return fmt.Sprintf("%s-app", i.GetClowdName())
 }
 
+func (i *ClowdApp) RunPreHook() bool {
+	// is there a better way to find out if the app has a job?
+	if i.Spec.PreHookJob.Name == "" || i.Annotations["clowder/pre-hook-status"] == "pending" {
```
psav, line 549: Specifically here I think that a comparison to `(i.Spec.PreHookJob{})` may do the job, but checking for the name is probably fine, as long as that is a mandatory field.

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -540,6 +544,26 @@ func (i *ClowdApp) GetClowdSAName() string {
...
+		return false
+	} else if i.Annotations["clowder/pre-hook-generation"] == "" {
```
psav, line 551: I'm still not 100% bought on this. I think it would perhaps be better to shift this state store onto the Job itself. The other problem you have is that it looks as if we are replacing the same job again and again. I'm not sure if this is the best idea, perhaps having a job per generation would be better, if jobs eventually get cleaned up anyway. That way the job name would be easy to search for as it's a contraction of the apps name and its generation. If you wanted to be more granular and only change when the image changed you could use the image tag as a name component. Of course this brings up the question of what happens if two spec changes are written in quick succession, triggering two jobs, but then I'm not 100% sure that this is solved even as we stand currently.

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -380,9 +383,14 @@ func (r *ClowdAppReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 			handler.EnqueueRequestsFromMapFunc(r.appsToEnqueueUponEnvUpdate),
 		).
+		Watches(
+			&source.Kind{Type: &batchv1.Job{}},
+			handler.EnqueueRequestsFromMapFunc(r.appToEnqueueOnJobCompletion),
+		).
 		Owns(&apps.Deployment{}).
 		Owns(&core.Service{}).
 		Owns(&core.ConfigMap{}).
+		// Owns(&batchv1.Job{}).
```
psav, line 393: Are you using the Watch because the Own can't have a filter enqueue on it? If so, I think the logic we have in the enqueuer is like this because of the attributes setup. In my mind rather than watching all Jobs, which could increased overall memory consumption anyway, watching the ones we own and always triggering this apps reconciliation seems like the good way to go. If we are worried about too many reconciliations because the job status changes too often, we have just redone the predicate system to make it FAR easier to define a customer predicate filter to only allow certain changes to the resource to trigger a reconciliation.

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -433,6 +441,67 @@ func (r *ClowdAppReconciler) appsToEnqueueUponEnvUpdate(a client.Object) []recon
...
+	}
+
+	// Filter based on job name
+	for _, app := range appList.Items {
+		// find the app that the job belongs to
+		if job.OwnerReferences[0].Name == app.Name {
+			// update the job with "done" state and mark the generation we last
+			// ran the job for.
+			app.ObjectMeta.Annotations["clowder/pre-hook-status"] = "done"
+			app.ObjectMeta.Annotations["clowder/pre-hook-generation"] = strconv.Itoa(int(app.GetGeneration()))
```
psav, line 490: Depending on when this gets triggered, it is perhaps a little bit of an assumption to say that we know for sure the job was run against this particular generation. If the spec has changed in the meantime, this may not be true.

psav: If however we encode the apps generation into the job when it is started, we know that _that_ generation of the the app was used to build and trigger the job initially.

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -433,6 +441,67 @@ func (r *ClowdAppReconciler) appsToEnqueueUponEnvUpdate(a client.Object) []recon
...
+	err = r.Client.List(ctx, &appList)
+	if err != nil {
+		r.Log.Error(err, "Failed to list ClowdApps")
+		return nil
+	}
+
+	// Filter based on job name
+	for _, app := range appList.Items {
+		// find the app that the job belongs to
+		if job.OwnerReferences[0].Name == app.Name {
```
psav, line 486: The commnt about owned before obviously makes this obsolete.

controllers/cloud.redhat.com/providers/autoscaler/default.go
```diff
@@ -20,6 +20,10 @@ func NewAutoScalerProvider(p *p.Provider) (p.ClowderProvider, error) {
 }
 
 func (asp *autoscalerProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error {
+	if !app.PreHookDone() {
```
psav, line 23: We need to be very careful about skipping here, as skipping parts of providers can mean the cache is not furnished with information for subsequent providers. I'm not quite sure of the solution yet, but this is a little scary.

psav: Oh - And we definitely need to log something here

controllers/cloud.redhat.com/providers/deployment/provider.go
```diff
@@ -17,5 +17,5 @@ func GetDeployment(c *providers.Provider) (providers.ClowderProvider, error) {
 }
 
 func init() {
-	providers.ProvidersRegistration.Register(GetDeployment, 0, ProvName)
+	providers.ProvidersRegistration.Register(GetDeployment, 1, ProvName)
```
psav, line 20: Changing provider ordering is dangerous!

psav: I actually have an idea for a new way of ordering - but it's a little more complex and going to take me some time to figure out.

controllers/cloud.redhat.com/providers/job/default.go
```diff
@@ -0,0 +1,32 @@
...
+	return &jobProvider{Provider: *p}, nil
+}
+
+func (jp *jobProvider) Provide(app *crd.ClowdApp, _ *config.AppConfig) error {
+	if app.RunPreHook() {
+		if err := jp.makePreHookJob(&app.Spec.PreHookJob, app); err != nil {
+			return err
+		}
+	} else if app.PreHookDone() {
+		fmt.Println("YEET")
```
psav, line 28: Comments go bye bye :)

controllers/cloud.redhat.com/providers/job/impl.go
```diff
@@ -1,15 +1,45 @@
...
+	if err := jp.Cache.Create(PreHookJob, nn, &j); err != nil {
+		return err
+	}
+
+	CreatePreHookJobResource(app, nn, job, &j)
+	if err := jp.Cache.Update(PreHookJob, &j); err != nil {
+		return err
+	}
+
+	app.ObjectMeta.Annotations["clowder/pre-hook-status"] = "pending"
```
psav, line 34: Can you elaborate on the choice between annotations and status fields here please?

controllers/cloud.redhat.com/providers/job/provider.go
```diff
@@ -0,0 +1,15 @@
...
+	p "github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/providers"
+)
+
+var ProvName = "job"
+
+func GetJob(c *p.Provider) (p.ClowderProvider, error) {
+	return NewJobProvider(c)
+}
+
+func init() {
```
psav, line 13: Now I see why you did this - to ensure you could go first - you would need to ensure ALL providers are bumped properly.

psav (changes requested):
Some discussion topics here I think

PR #512, opened 2021-11-15
Add support for setting 'resources' on KafkaConnect

bsquizz:
Also I cleared up the "ambiguous section of code" that tripped us up last time we debugged the Kafka cluster's resource limits (which are being set via the env, not hard-coded)

Updated a kuttl test to test for resource settings on kafka + kafka connect


controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -160,13 +150,10 @@ func (s *strimziProvider) configureKafkaCluster() error {
...
-			Version:  &version,
-			Replicas: replicas,
-			Resources: &strimzi.KafkaSpecKafkaResources{
-				Requests: &kRequests,
-				Limits:   &kLimits,
-			},
+			Config:    &kConfig,
+			Version:   &version,
+			Replicas:  replicas,
+			Resources: &s.Env.Spec.Providers.Kafka.Cluster.Resources,
```
psav, line 156: Changing this means we have no defaults set at all though right?

controllers/cloud.redhat.com/providers/kafka/strimzi.go
```diff
@@ -449,6 +435,7 @@ func (s *strimziProvider) configureKafkaConnectCluster() error {
 		Version:          &version,
 		Config:           &config,
 		Image:            &image,
+		Resources:        &s.Env.Spec.Providers.Kafka.Connect.Resources,
```
psav, line 438: Should we have defaults for connect too?

psav (changes requested):

controllers/cloud.redhat.com/providers/kafka/strimzi.go
bsquizz: Remember, the `Spec.Resources` was getting overwritten at a later step of the code (line 223 in the old code)

You're right, I think at the moment (and even before this PR) there was no concept of a default Resources value. So I'll add that.

controllers/cloud.redhat.com/providers/kafka/strimzi.go
bsquizz: Will add

controllers/cloud.redhat.com/providers/kafka/strimzi.go
bsquizz: Added defaults, take a look

psav:
Looks good to me - will hold off on merging until we have discussed the possible fallout

psav (approved):

PR #514, opened 2021-11-18
[RHCLOUD-17085] Added concurrency

psav:


kylape (approved):

kylape (approved):

kylape (changes requested):
* switch back to mutex
* don't forget to set the mutex!
* add mutex to app and cji controller to handle concurrent reconciliation of same resource

PR #518, opened 2021-11-24
[RHCLOUD-16682] Cloudwatch logging

psav:


kylape (changes requested):
Please review if we need to copy all the logging code out of the go-middleware library.

PR #526, opened 2021-11-30
[RHCLOUD-17103] Switched to UBI8 minimal

psav:


kylape (changes requested):
let's at least remove bash from the image

PR #536, opened 2021-12-03
[RHCLOUD-16525] Added pull policy flag

psav:


apis/cloud.redhat.com/v1alpha1/clowdenvironment_types.go
```diff
@@ -406,6 +406,10 @@ type Sidecars struct {
 	TokenRefresher TokenRefresherConfig `json:"tokenRefresher,omitempty"`
 }
 
+type DeploymentConfig struct {
+	OmitPullPolicy bool `json:"omitPullPolicy,omitempty"`
```
vkrizan, line 410: I wonder if this can expose full imagePullPolicy control. For k8s default an empty string value can be used as per [k8s condition](https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/apis/core/v1/defaults.go#L74).

psav: Of course it's possible, I'll check with @kylape but the problem with this approach is that you could set it to Always, which would set it to Always on ALL deployments. If you have many clowdapps in a clowdenv this would have them all set to Always, which could be load inducing.

psav: After discussing this with @kylape, we feel this is the best way forward. There is no reason for someone to want to set Always on a non latest tag. With this option, omitting the pull policy means that k8s will do the "right thing". If you have an image with a tag, it will use PullIfNotPresent, if you have `latest` set, then it will set it to Always. I think this should do what you are asking.

PR #546, opened 2021-12-10
[RHCLOUD-16920] Add deployment strategy

psav:


psav:
Note that merging this will change the deployment strategy on puptoo and others in stage

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -189,6 +189,16 @@ type Deployment struct {
...
+
+	// DeploymentStrategy allows the deployment strategy to be set only if the
+	// deployment has no public service enabled
+	DeploymentStrategy *DeploymentStrategy `json:"deploymentStrategy,omitempty"`
+}
+
+type DeploymentStrategy struct {
+	// OverridePrivate allows a deployment that only uses a private port to use
+	// the Rolling update, as the default is Recreate.
+	OverridePrivate bool `json:"overridePrivate,omitempty"`
```
kylape, line 201: why not just name it `Strategy` or whatever mirrors the actual `DeploymentStrategy` struct?

controllers/cloud.redhat.com/providers/deployment/impl.go
```diff
@@ -72,6 +72,12 @@ func initDeployment(app *crd.ClowdApp, env *crd.ClowdEnvironment, d *apps.Deploy
 	}
 	d.Spec.ProgressDeadlineSeconds = common.Int32Ptr(600)
 
+	if deployment.DeploymentStrategy != nil && !deployment.WebServices.Public.Enabled && deployment.DeploymentStrategy.OverridePrivate {
+		d.Spec.Strategy = *&apps.DeploymentStrategy{
+			Type: "Recreate",
```
kylape, line 77: why not use const like we do above?

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -189,6 +189,16 @@ type Deployment struct {
...
 	AutoScaler *AutoScaler `json:"autoScaler,omitempty"`
+
+	// DeploymentStrategy allows the deployment strategy to be set only if the
+	// deployment has no public service enabled
+	DeploymentStrategy *DeploymentStrategy `json:"deploymentStrategy,omitempty"`
+}
+
+type DeploymentStrategy struct {
+	// OverridePrivate allows a deployment that only uses a private port to use
+	// the Rolling update, as the default is Recreate.
```
kylape, line 200: can we explain why recreate is allowed for private endpoints?

kylape (changes requested):
do we know of any private deployments that host private endpoints that would need to use the rollout strategy?

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
psav: Because if you see - we are not giving the full struct here - we decided to just go with the bool right?

controllers/cloud.redhat.com/providers/deployment/impl.go
psav: Will do

psav: This is all wrong anywya

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
psav: done

controllers/cloud.redhat.com/providers/deployment/impl.go
```diff
@@ -72,6 +72,18 @@ func initDeployment(app *crd.ClowdApp, env *crd.ClowdEnvironment, d *apps.Deploy
 	}
 	d.Spec.ProgressDeadlineSeconds = common.Int32Ptr(600)
 
+	if !deployment.WebServices.Public.Enabled && deployment.WebServices.Private.Enabled {
+		if deployment.DeploymentStrategy != nil && deployment.DeploymentStrategy.PrivateStrategy != "" {
```
BlakeHolifield, line 76: Are we fine with just ignoring the strategy if it's public? Do we want some log or error to appear for debugging?

bundle/tests/scorecard/kuttl/test-deployment-strategy/01-scripts.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- script: sleep 1
+#- script: kubectl apply -f pods.yaml --namespace=test-deployment-strategy 2>&1 | grep "Forbidden"
```
BlakeHolifield, line 6: Commented out because we don't error here, but because we only skip, correct?

BlakeHolifield (commented):
A few questions before we merge.

controllers/cloud.redhat.com/providers/deployment/impl.go
psav: I am personally fine with it since we have said in the docs it's ignored

PR #561, opened 2022-01-19
No longer using random ident in env hostname

bsquizz:
A ClowdEnvironment name is unique on a cluster anyways, so we shouldn't need an additional layer of uniqueness.

This makes it easier for the NS operator to set up a FrontendEnvironment because the hostname follows a predictable format


bsquizz:
We've decided we're going to keep this.

PR #564, opened 2022-01-27
[RHCLOUD-18698] Remove network policy creation

bsquizz:
We are now creating this NetworkPolicy via the project template in OpenShift.

### DIFF
--- a/bundle/tests/scorecard/kuttl/test-minio-app/01-assert.yaml
+++ b/bundle/tests/scorecard/kuttl/test-minio-app/01-assert.yaml
@@ -126,24 +126,3 @@ spec:
     env-app: test-minio-app-minio
   sessionAffinity: None
   type: ClusterIP
---- 
-apiVersion: networking.k8s.io/v1
-kind: NetworkPolicy
-metadata:
-  labels:
-    app: test-minio-app
-  name: allow-from-clowder-system-namespace
-  namespace: test-minio-app
-  ownerReferences:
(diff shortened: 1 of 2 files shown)


psav:
Card number in title please

psav:
I actually think we can keep this - We may need it if someone were to use clowder with CRC for instance

bsquizz:
I don't think a card was ever created for this, it was originally an unexpected toil issue. Created one now.

> I actually think we can keep this - We may need it if someone were to use clowder with CRC for instance

If you want to keep it then that's fine, I opened this because it seems like an unnecessary code block now.

We had to add a network policy to allow the Clowder ns to communicate w/ an ephemeral NS in the OSD project template (via app-interface) so that Clowder would be able to communicate with keycloak. Since we added that policy to the project template, there's no need for the minio provider to create another duplicate policy -- and on minikube all namespaces can communicate with the others anyway.

bsquizz:
OK, let's rework this. Instead of removing the NetworkPolicy creation -- it should live in a more centralized place. This is because a NetworkPolicy needs to be created in more cases than just for minio now.

It might be easiest to just *always* (if on OpenShift) create a NetworkPolicy allowing clowder to talk to the ClowdEnvironment's target namespace.

psav:
That makes complete sense - yeh lets move it to another more centralized place then

psav:
Should I close this PR off in lieu of the new one?

bsquizz:
Will address this differently in another PR

PR #568, opened 2022-02-08
[RHCLOUD-18029] Use ubi-micro

psav:


psav:
This is failing because certs are not included in ubi-micro

PR #571, opened 2022-02-15
[RHCLOUD-18150] Added base-build image

psav:


bsquizz:
Nevermind, I see that it will be rebuilt and re-pushed within the PR check any time a change is detected to go.mod/go.sum/Dockerfile.base

bsquizz (approved):

Dockerfile.base
```diff
@@ -0,0 +1,37 @@
...
+RUN make controller-gen kustomize
+
+COPY hack/boilerplate.go.txt hack/boilerplate.go.txt
+
+COPY main.go main.go
+COPY config/ config/
+COPY apis/ apis/
+COPY controllers/ controllers/
+
+RUN make manifests generate fmt vet release
```
kylape, line 32: why do we do this in the base image build?

pr_check.sh
```diff
@@ -15,10 +15,22 @@ done <<< "$(git log --pretty=format:%s $(git merge-base master HEAD)..HEAD)"
...
 mkdir -p "$DOCKER_CONF"
 docker login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
 
+RESPONSE=$( \
+        curl -Ls -I -o /dev/null -w "%{http_code}" -H "Authorization: Bearer $QUAY_API_TOKEN" \
+        https://quay.io/api/v1/repository/cloudservices/clowder-base/tag/$BASE_TAG/images \
+    )
+    echo "received HTTP response: $RESPONSE"
+if [[ $RESPONSE != 200 ]]; then
+    BASE_IMG=$BASE_IMG make docker-build-and-push-base
```
kylape, line 31: should this check also happen somewhere outside the PR check?

Dockerfile.base
psav: Because vet is costly

PR #572, opened 2022-02-15
[RHCLOUD-14127] Support for running a selenium container when running an IQE CJI

bsquizz:
The CJI will have a new section under `iqe`:
```yaml
     ui:
       selenium:
         deploy: true  # indicates a selenium container will be included                 
         imageTag: "my-custom-image-tag"  # tag can be overridden if desired
```

The ClowdEnvironment `testing` spec has new options added under `iqe` to allow for configuration of selenium containers:
```yaml
     ui:
       # defines options for selenium UI containers
       selenium:
         imageBase: "quay.io/cloudservices/selenium"
         defaultTag: "sometag"
         resources: {}
```

This is outlined in https://docs.google.com/document/d/1ZCmY_50wkoxOu-mRJF3FWluTO1083hvZ5v-GCXQVER8/edit#heading=h.jzpw4e58uo7k

I split some the logic for creating the IQE container and Selenium container out into separate functions within impl.go -- the `createIqeContainer` function was going to become too big


apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
```diff
@@ -76,9 +76,17 @@ type IqeJobSpec struct {
 	TestImportance *[]string `json:"testImportance,omitempty"`
 }
 
-type UiSpec struct {
-	// Indiciates the presence of a selenium container
-	Enabled bool `json:"enabled"`
```
bsquizz, line 81: Since `ui.enabled` didn't do anything it wasn't really used by anyone anywhere as far as I know... for example it was not used in bonfire's CJI template: https://github.com/RedHatInsights/bonfire/blob/master/bonfire/resources/default-iqe-cji.yaml

I can see it is also not used in cloudigrade's release-gating CJI template: https://github.com/cloudigrade/cloudigrade/blob/master/deployment/stage-smoke.yaml

I think just completely removing it should be ok.

bsquizz:
FYI @quarckster

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -119,7 +83,49 @@ func CreateIqeJobResource(cache *rc.ObjectCache, cji *crd.ClowdJobInvocation, en
 		TerminationMessagePolicy: core.TerminationMessageReadFile,
 	}
 
+	return c
+}
+
+func createSeleniumContainer(name string, nn types.NamespacedName, cji *crd.ClowdJobInvocation, env *crd.ClowdEnvironment, app *crd.ClowdApp) core.Container {
+	// create env vars
+	envVars := []core.EnvVar{
+		{Name: "HOME", Value: "/home/selenium"},
```
quarckster, line 92: https://github.com/RedHatQE/selenium-images/blob/master/standalone/Dockerfile#L26

bundle/tests/scorecard/kuttl/test-iqe-jobs-selenium/01-pods.yaml
```diff
@@ -0,0 +1,91 @@
...
+        imageBase: "quay.io/psav/clowder-hello"
+        resources:
+          limits:
+            cpu: 2
+            memory: 600Mi
+          requests:
+            cpu: 200m
+            memory: 256Mi
+        ui:
+          selenium:
```
quarckster, line 25: does it mean that other UI sidecars can be specified? I mean why specify `selenium` under `ui`?

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -119,7 +83,49 @@ func CreateIqeJobResource(cache *rc.ObjectCache, cji *crd.ClowdJobInvocation, en
...
+	}
+
+	// set image tag
+	image := env.Spec.Providers.Testing.Iqe.UI.Selenium.ImageBase
+	if image == "" {
+		image = "quay.io/redhatqe/selenium-standalone"
+	}
+	tag := env.Spec.Providers.Testing.Iqe.UI.Selenium.DefaultImageTag
+	if tag == "" {
+		tag = "ff_91.5.1esr_gecko_v0.30.0_chrome_98.0.4758.80"
```
quarckster, line 102: I have a little concern about the hardcoding tag name here.

apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
psav: While it should be - I would prefer we follow our procedure and deprecate it - especially as we are not trying to reuse it

bundle/tests/scorecard/kuttl/test-iqe-jobs-selenium/01-pods.yaml
psav: I think this is why yes, so that we leave it open in the future for different kinds of UI sidecars

controllers/cloud.redhat.com/providers/iqe/impl.go
psav: It's only a default, but perhaps it would be better to make the default part of clowders config? It's not the only place we hardcode images in Clowder thuogh.

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -167,7 +178,55 @@ func CreateIqeJobResource(cache *rc.ObjectCache, cji *crd.ClowdJobInvocation, en
 		logger.Info("No config mounted to the iqe pod")
 	}
 
-	j.Spec.Template.Spec.Containers = []core.Container{c}
+	return nil
+}
+
+// CreateIqeJobResource will create a Job that contains a pod spec for running IQE
+func CreateIqeJobResource(cache *rc.ObjectCache, cji *crd.ClowdJobInvocation, env *crd.ClowdEnvironment, app *crd.ClowdApp, nn types.NamespacedName, ctx context.Context, j *batchv1.Job, logger logr.Logger, client client.Client) error {
```
psav, line 185: Why not make these methods? It could simplify the prototype considerably I think.

psav (changes requested):

psav:
Can we have the card number in the title pls also?

apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
bsquizz: OK, it literally did nothing previously, so I suppose I can just leave it in the spec but update the comment to say it is not used?

controllers/cloud.redhat.com/providers/iqe/impl.go
bsquizz: Nice, will remove that

bundle/tests/scorecard/kuttl/test-iqe-jobs-selenium/01-pods.yaml
bsquizz: Correct ... just some future-proofing here

controllers/cloud.redhat.com/providers/iqe/impl.go
bsquizz: We will deal w/ default values in a follow-up PR. As Pete mentioned we have a few cases of these hard-coded values and we'll move them all to config via a different PR. The default image tag for selenium is actually normally going to be controlled by the value set in the ClowdEnvironment. This hard-coded value here is just a fallback to ensure it is never empty by accident.

apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
bsquizz: Added 'Enabled' into the IqeUISpec

controllers/cloud.redhat.com/providers/iqe/impl.go
bsquizz: Are you referring to all the functions here in this file or just some pieces of CreateIqeJobResource?

And I'm a little confused on what would be the receiver on the method? (I didn't write this originally so was keeping the structure of the file more or less similar other than splitting up this bug function into a couple smaller ones)

bundle/metadata/annotations.yaml
```diff
@@ -5,7 +5,7 @@ annotations:
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: clowder
   operators.operatorframework.io.bundle.channels.v1: alpha
-  operators.operatorframework.io.metrics.builder: operator-sdk-v1.12.0
+  operators.operatorframework.io.metrics.builder: operator-sdk-v1.6.1+git
```
BlakeHolifield: This doesn't seem intentional and is maybe some local dev environment artifact, yeah?

controllers/cloud.redhat.com/providers/iqe/impl.go
bsquizz: We chatted on slack and agreed to backlog this refactor backlog that refactor since jobs and CJI's don't currently follow the model of registering a provider

BlakeHolifield (approved):

bundle/metadata/annotations.yaml
bsquizz: Yeah accidental commit, fixed

PR #578, opened 2022-03-17
Add ability to override IQE plugins on CJI

bsquizz:
This adds a `plugins` field to the CJI IQE spec. By default, Clowder will use the plugin specified on the ClowdApp's `.spec.testing.iqePlugin` field (as it always has) -- but it can now be overridden using the 'plugins' field if desired when applying a CJI.

### DIFF
--- a/apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
+++ b/apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
@@ -50,6 +50,11 @@ type IqeJobSpec struct {
 	// baseImage:name-of-iqe-plugin, but only the tag can be overridden here
 	ImageTag string `json:"imageTag,omitempty"`
 
+	// By default, Clowder will use the plugin name indicated in the ClowdApp's
+	// spec.testing.iqePlugin field. A comma,separated,list of plugins can be supplied
+	// here if you wish you override the plugins.
+	IqePlugins string `json:"plugins,omitempty"`
+
 	// Indiciates the presence of a selenium container
 	// Note: currently not implemented
 	UI IqeUISpec `json:"ui,omitempty"`
--- a/controllers/cloud.redhat.com/providers/iqe/impl.go
+++ b/controllers/cloud.redhat.com/providers/iqe/impl.go
@@ -35,12 +35,16 @@ func joinNullableSlice(s *[]string) string {
 
 func createIqeContainer(j *batchv1.Job, nn types.NamespacedName, cji *crd.ClowdJobInvocation, env *crd.ClowdEnvironment, app *crd.ClowdApp) *core.Container {
 	// create env vars
+	iqePlugins := app.Spec.Testing.IqePlugin
+	if cji.Spec.Testing.Iqe.IqePlugins != "" {
+		iqePlugins = cji.Spec.Testing.Iqe.IqePlugins
+	}
 	envVars := []core.EnvVar{
 		{Name: "ENV_FOR_DYNACONF", Value: cji.Spec.Testing.Iqe.DynaconfEnvName},
 		{Name: "NAMESPACE", Value: nn.Namespace},
 		{Name: "CLOWDER_ENABLED", Value: "true"},
 		{Name: "ACG_CONFIG", Value: "/cdapp/cdappconfig.json"},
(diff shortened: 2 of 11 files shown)


apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
```diff
@@ -50,6 +50,11 @@ type IqeJobSpec struct {
 	// baseImage:name-of-iqe-plugin, but only the tag can be overridden here
 	ImageTag string `json:"imageTag,omitempty"`
 
+	// By default, Clowder will use the plugin name indicated in the ClowdApp's
+	// spec.testing.iqePlugin field. A comma,separated,list of plugins can be supplied
+	// here if you wish you override the plugins.
+	IqePlugins string `json:"plugins,omitempty"`
```
BlakeHolifield, line 56: So is it default + this list or only this list? With an inventory example is it:
`value: "e2e"` => inventory + e2e

or

`value: host-inventory,e2e` => inventory + e2e

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -35,12 +35,16 @@ func joinNullableSlice(s *[]string) string {
...
+	if cji.Spec.Testing.Iqe.IqePlugins != "" {
+		iqePlugins = cji.Spec.Testing.Iqe.IqePlugins
+	}
 	envVars := []core.EnvVar{
 		{Name: "ENV_FOR_DYNACONF", Value: cji.Spec.Testing.Iqe.DynaconfEnvName},
 		{Name: "NAMESPACE", Value: nn.Namespace},
 		{Name: "CLOWDER_ENABLED", Value: "true"},
 		{Name: "ACG_CONFIG", Value: "/cdapp/cdappconfig.json"},
-		{Name: "IQE_PLUGINS", Value: app.Spec.Testing.IqePlugin},
+		{Name: "IQE_PLUGINS", Value: iqePlugins},
```
BlakeHolifield, line 47: Same sort of question as above. Do you get your app's IQE plugin included or does it need to be explicitly added with your extra plugins?

BlakeHolifield (commented):
Looks good! Just a few semantics questions

apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
bsquizz: Setting this would completely override, it wouldn't add the default to it.

controllers/cloud.redhat.com/providers/iqe/impl.go
bsquizz: If you specify `plugins` on the CJI it totally overrides `app.Spec.Testing.IqePlugin`

BlakeHolifield (approved):

apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
psav: I really wish we hadn't already set the precedence of using a CSV string here. I actually would argue that as this is a simple list of plugin names, that from a readability PoV it would make sense to turn it into an array. But given that most of the other IQE fields are not using this precedence I guess we can forgo it.

bsquizz: Yeah we could have it as a list of strings in the spec, but this gets painful. We have to do the work in Clowder of converting it back into a CSV string before setting the env var since that's how the code that runs in the container parses the env var. Example here: https://github.com/RedHatInsights/clowder/blob/master/controllers/cloud.redhat.com/providers/iqe/impl.go#L50-L51

Also templating gets a little uglier. For ex right now it's easy to just type `bonfire deploy-iqe-cji --plugins plugin1,plugin2` and simply pass that string into the template as a parameter: https://github.com/RedHatInsights/bonfire/blob/master/bonfire/resources/default-iqe-cji.yaml#L19

Compare that to how the REQUIREMENTS and REQUIREMENTS_PRIORITY values are set.

So we end up having to add code on both sides to convert CLI args into a list and then over on the clowder side back out of a list into a plain old CSV string... when we could just pass a CSV string straight through :man_shrugging:

PR #580, opened 2022-03-23
Create non-prefixed svc for mbop/keycloak/mocktitlements

bsquizz:
This PR creates an additional Service resource named without a prefix in a ClowdEnvironment's target namespace for the following components:
* mbop
* mocktitlements
* keycloak


psav:
Can we have a JIRA in the title please

bsquizz:
Turns out this isn't going to work as easily as expected because `CachedMakeComponent` adds resources into the cache using the prefixed name.

PR #583, opened 2022-03-24
Rhcloud 18319

adamrdrew:
This change implements the request from RHCLOUD-18319 - it adds resource limits and requests to `DatabaseSpec` that are in line with what's in `PodSpec`. Changes to `MakeLocalDB` and surrounding code to handle the new opts and provide sane defaults. Updated unit test, added new `test-localdb` kuttl test because there wasn't one. Docs updated.


adamrdrew:
I had to set up my GPG config. Will send a new PR with signed commits.

PR #584, opened 2022-03-24
[RHCLOUD-18319] Added resource requests and limits to database provid…

adamrdrew:
…er along with tests and docs.

Signed-off-by: Adam Drew <addrew@redhat.com>


psav:
You'll need to run ``make build-template`` and commit the ``deploy.yaml`` and ``deploy_mutate.yaml`` before this will pass tests

psav:
@BlakeHolifield Do we want to leave it completely open to set resource limits here? Or do we want to t-shirt size it?

adamrdrew:
Thanks for looking! And good idea about the sizes. Would you want that in a new PR or in this one if we do it.

psav:
As we are intending to modify the spec, we would want to do it here I think - apologies - we can discuss in standup briefly today and spin out into another meeting if we feel necessary

BlakeHolifield:
@psav Yeah that's an interesting point. My only reservation here is that we don't do any t-shirt sizing for `resources` currently. It might be a bit confusing to be able to set the full `resources key on pods and jobs, but not on databases.

adamrdrew:
Let me know if there's anything else yall need me to do on this. Tests were failing because md5sum on my deploy.yaml didn't match what jenkins was generating. It was because my pyyaml was way out of date. Confirmed that md5 on mine matches jenkins now. Noticed test status has disappeared from this PR - not sure if that's normal or not.

adamrdrew:
retest ran, it failed in:
```
11:14:00         --- FAIL: kuttl/harness/test-multi-db-shared (3.83s)
11:14:00         --- FAIL: kuttl/harness/test-web-services (13.39s)
```
So I guess I need to track down if my changes caused that. I mean, I assume they do. I'll catch up with @BlakeHolifield

adamrdrew:
Kuttle failed again with:
```
resource Deployment:test-localdb/app-a-db: .spec.template.spec.containers.resources.limits.memory: value mismatch, expected: 1024Mi != actual: 1Gi
```
Kinda surprised. Why doesn't this affect anything else - I copied and pasted those limits.  Fixed and pushed will see what happens.

adamrdrew:
OK, so I think everything I touched it working. Builds are still failing though 
```


kuttl.test-multi-db        Error Details            Error Details           failed in step 1-pods Stack Trace            Stack Trace           Internal error occurred: failed calling webhook "vclowdapp.kb.io": Post "https://clowder-webhook-service.clowder-system.svc:443/validate-cloud-redhat-com-v1alpha1-clowdapp?timeout=10s": context deadline exceeded | 13 sec | 2
-- | -- | --
kuttl.test-multi-db-shared        Error Details            Error Details           failed in step 1-pods Stack Trace            Stack Trace           Internal error occurred: failed calling webhook "vclowdapp.kb.io": Post "https://clowder-webhook-service.clowder-system.svc:443/validate-cloud-redhat-com-v1alpha1-clowdapp?timeout=10s": dial tcp 10.106.227.93:443: connect: connection refused

[kuttl.test-multi-db](https://ci.ext.devshift.net/job/RedHatInsights-clowder-pr-check/2218/testReport/(root)/kuttl/test_multi_db/)
 Error Details  Error Details

failed in step 1-pods

 Stack Trace  Stack Trace

Internal error occurred: failed calling webhook "vclowdapp.kb.io": Post "https://clowder-webhook-service.clowder-system.svc:443/validate-cloud-redhat-com-v1alpha1-clowdapp?timeout=10s": context deadline exceeded

	13 sec	[2](https://ci.ext.devshift.net/job/RedHatInsights-clowder-pr-check/2217/)
 [kuttl.test-multi-db-shared](https://ci.ext.devshift.net/job/RedHatInsights-clowder-pr-check/2218/testReport/(root)/kuttl/test_multi_db_shared/)
 Error Details  Error Details

failed in step 1-pods

 Stack Trace  Stack Trace

Internal error occurred: failed calling webhook "vclowdapp.kb.io": Post "https://clowder-webhook-service.clowder-system.svc:443/validate-cloud-redhat-com-v1alpha1-clowdapp?timeout=10s": dial tcp 10.106.227.93:443: connect: connection refused
```

So, not sure what's up with that. I haven't touched those tests (other than to copy stuff out of them) and that looks like an external dependency?

bundle/tests/scorecard/kuttl/test-localdb/01-assert.yaml
```diff
@@ -0,0 +1,29 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: app-b-db
+  namespace: test-localdb
+spec:
+  template:
+    spec:
+      containers:
+        - image: quay.io/cloudservices/postgresql-rds:10-9ee2984
```
bsquizz, line 29: Should we be asserting that the 'defaults' were used here for resources on app B's DB? I'm looking at https://github.com/RedHatInsights/clowder/pull/584/files#diff-5df7ec685a29a1eed6f24d8f66cf2129089eff23408cff9e91e550aa611f7a0fR113

config/manager/kustomization.yaml
```diff
@@ -17,5 +17,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: 127.0.0.1:5000/clowder
-  newTag: b7db066
+  newName: quay.io/cloudservices/clowder
+  newTag: 0a44742
```
bsquizz, line 21: I think we usually omit changes to this file (it's an auto-generated change from the operator-sdk that we don't actually want to push AFAIK). See Blake's new Makefile helper for this here: https://github.com/RedHatInsights/clowder/pull/587

Run a `git checkout origin/master config/manager/kustomization.yaml`

controllers/cloud.redhat.com/providers/database/localdb.go
```diff
@@ -106,7 +106,14 @@ func (db *localDbProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error
 	}
 
 	labels := &map[string]string{"sub": "local_db"}
-	provutils.MakeLocalDB(dd, nn, app, labels, &dbCfg, image, db.Env.Spec.Providers.Database.PVC, app.Spec.Database.Name, nil)
+
+	databaseResourceRequirements := app.Spec.Database.Resources
+
+	if databaseResourceRequirements.Limits == nil || databaseResourceRequirements.Requests == nil {
```
bsquizz, line 112: I wonder if we should instead be checking if `app.Spec.Database.Resources` is "empty". That would look like:
```
emptyResourceRequirements := v1.ResourceRequirements{}
if databaseResourceRequirements == emptyResourceRequirements  {
    // use defaults
}
```

Reason is, it is valid to define a resource block which sets no value for requests but DOES set limits (it may rarely be used that way but if we want it to be truly a 'pass-thru' config we'd probably want to consider that).

controllers/cloud.redhat.com/providers/utils/utils.go
```diff
@@ -15,6 +15,20 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+func DatabaseResourceDefaults() core.ResourceRequirements {
```
bsquizz, line 18: Don't need to address this now ... but IMO we should have a follow-up PR that obtains this value from a configuration block in the ClowdEnvironment... so that we can adjust the defaults if desired without needing a new code push.

controllers/cloud.redhat.com/version.txt
```diff
@@ -1 +1 @@
-master
\ No newline at end of file
+v0.29.0-15-gdfa3d6e
```
bsquizz, line 1: This is another auto-generated change you want to revert: `git checkout origin/master controllers/cloud.redhat.com/version.txt`

bsquizz (changes requested):
Looking good! Mostly nitpicking here ;)

bundle/tests/scorecard/kuttl/test-localdb/01-assert.yaml
adamrdrew: Yeah that makes sense. I'll add a test for defaults on app b.

controllers/cloud.redhat.com/providers/database/localdb.go
adamrdrew: I don't know what `v1` is here. Its not in scope. There are a couple of packages we import that have `v1` in their paths like `apps "k8s.io/api/apps/v1"` and  `core "k8s.io/api/core/v1"` but neither of them expose a `ResourceRequirements` - where should I be pulling that from?

adamrdrew: NVM - its core. VSCode's completion just wasn't catching up.

adamrdrew: Here's my code btw

```
	emptyResourceRequirements := core.ResourceRequirements{}
	databaseResourceRequirements := app.Spec.Database.Resources

	if databaseResourceRequirements == emptyResourceRequirements {
		databaseResourceRequirements = provutils.DatabaseResourceDefaults()
	}
```

adamrdrew: Should this work? Internet seems to say that our options for struct comparisons are either slow (reflect.DeepCopy), dangerous (cmp.Equal), or roll your own (check equality of struct members.) I have no idea if that's true or not.

adamrdrew: @bsquizz let me know what you think re: the above. If we can't just compare the structs directly that might be why we did the initial code checking for the members being null.

adamrdrew:
We talked a bit in chat and decided on a different change for that comparison. I think I've fulfilled everything requested, but I'm new to this interface so let me know if I missed something or need something else. Tests are massing so we're either ready for more review or merge I think.

bsquizz (approved):
My requests were addressed! Thanks.

bundle/tests/scorecard/kuttl/test-localdb/01-assert.yaml
```diff
@@ -0,0 +1,41 @@
...
+          resources:
+            limits:
+              cpu: 1200m
+              memory: 1Gi
+            requests:
+              cpu: 600m
+              memory: 512Mi
```
psav, line 41: New line on this file pls

controllers/cloud.redhat.com/providers/database/localdb.go
```diff
@@ -106,7 +106,14 @@ func (db *localDbProvider) Provide(app *crd.ClowdApp, c *config.AppConfig) error
 	}
 
 	labels := &map[string]string{"sub": "local_db"}
-	provutils.MakeLocalDB(dd, nn, app, labels, &dbCfg, image, db.Env.Spec.Providers.Database.PVC, app.Spec.Database.Name, nil)
+
+	databaseResourceRequirements := app.Spec.Database.Resources
+
+	if databaseResourceRequirements.Limits == nil && databaseResourceRequirements.Requests == nil {
+		databaseResourceRequirements = provutils.DatabaseResourceDefaults()
```
psav, line 113: Is there the potential for confusing behaviour here, if they only set one, their choices will get overwritten by the defaults?

psav (changes requested):
Originally we talked about t-shirt sizing these @BlakeHolifield did you decide against doing that?

controllers/cloud.redhat.com/providers/database/localdb.go
adamrdrew: We originally had it `||`, brandon suggested `&&`. I think there are arguments for both approaches and I don't know enough about kubernetes to adjudicate between them. Logical `or` makes sense because if they provide resources but not limits what does the system use for limits? Logical `and` makes sense because principle of least surprise would require that we never stomp over user input, even if partial.

I'm happy to implement either, y'all just let me know what you think.

BlakeHolifield: I personally think you should always set both and that only setting one would be more confusing. If your app has specific resource limits, they're likely higher than usual. Setting high limits and standard requests can lead to node overcommits. The other way around can lead to high requests and low limits that might lead to OOM kills. You either set your standards or take the defaults, but that's just my take.

psav: I agree we could absolutely use the validating webhook here..............

adamrdrew: Sounded like we agreed to address webhook in a follow up MR. So, I think I'm going to revert back to `||`.

adamrdrew:
We talked about this before the holiday; I'm going to move this to the t-shirt sizing system we use elsewhere.

adamrdrew:
Hmmm, well, I started going down the road of implementing t shirt sizes for CPU and RAM but it gets murky. We have 2 options:

1. Extend the spec with new `DBCPUSize` and `DBRAMSize` options (and corresponding JSON opts) to match the currently existing `DBVolumeSize`
2. Replace `DBVolumeSize` with `Size` and have volume, RAM, and CPU all just use that

Option 1 doesn't require a spec change from any user configs, it just provides new options. However, it makes config (arguably) confusing and messy with a 3x3x3 matrix of possible size combos. Option 2 is parsimonious and thus my aesthetic preference but it will require a spec change for customers. So, that's worth considering and discussing. 

I spoke with @psav this evening about implementation and we talked about going with the single size, but the spec change requirement didn't hit me then so I'm noting it here now and we'll talk about it tomorrow.

adamrdrew:
Upon further consideration it may make sense for me to just close this PR and start over because I'm pretty sure t-shirt sizing wont share a single line of code with this PR! Well catch up tomorrow :)

adamrdrew:
We talked a bit this AM and decided that we'd just add to the spec rather than change it. DBVolumeSize will stay as is; we'll add a sibling DBResourceSize that will deal with compute and RAM.

adamrdrew:
I'm closing this in favor of an alternative approach in https://github.com/RedHatInsights/clowder/pull/597

PR #587, opened 2022-03-30
[RHCLOUD-18694] Use pre-push to remove excessive files being updated

BlakeHolifield:
Since we have lots of errant files being modified locally and pushed up, we should clear those out as part of our pre-push target. These files can't really be git ignored, but they need to be restricted when they update instead of updated by default.

* config/manager/kustomization.yaml
* bundle/metadata/annotations.yaml 
* cloud.redhat.com/version.txt 
* config/manifests/bases/clowder.clusterserviceversion.yaml

Will now be reverted to the master version and not pushed up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,9 +153,16 @@ deploy-minikube: bundle docker-build-no-test docker-push-minikube deploy
 
 deploy-minikube-quick: bundle docker-build-no-test-quick docker-push-minikube deploy
 
+# we can't git ignore these files, but we want to avoid overwriting them
+no-update:
+	git checkout origin/master -- config/manager/kustomization.yaml \
+								  bundle/metadata/annotations.yaml \
+								  controllers/cloud.redhat.com/version.txt \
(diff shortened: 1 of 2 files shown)


psav:
Card number please

bsquizz (approved):
Yes. Awesome. Thank you :D

Makefile
```diff
@@ -153,9 +153,16 @@ deploy-minikube: bundle docker-build-no-test docker-push-minikube deploy
 
 deploy-minikube-quick: bundle docker-build-no-test-quick docker-push-minikube deploy
 
+# we can't git ignore these files, but we want to avoid overwriting them
+no-update:
+	git checkout origin/master -- config/manager/kustomization.yaml \
```
bsquizz, line 158: One suggestion. We probably want to `git fetch origin` before we do this.

bsquizz (changes requested):

PR #597, opened 2022-04-20
[RHCLOUD-18319] Add localdb resource requirements T Shirt sizing

adamrdrew:
We did a bunch of work in https://github.com/RedHatInsights/clowder/pull/584 but at the end of review we decided that t shirt sizing was a better UX and more consistent with Clowder philosophy. I started the work in that same branch but the work really represented a do-over sharing not a single line of code in common. I figured rather than trying to make the new changes while undoing my old changes I'd just start over.

I'm submitting this in draft at first so I can use jenkins for kuttl tests, will flip it for review later.


adamrdrew:
When I get this passing tests we''ll need to figure out what the sizes and defaults should be. It is worth noting that in #584 the defaults were set to:
```go
func DatabaseResourceDefaults() core.ResourceRequirements {
	return core.ResourceRequirements{
		Limits: core.ResourceList{
			"memory": resource.MustParse("1Gi"),
			"cpu":    resource.MustParse("1200m"),
		},
		Requests: core.ResourceList{
			"memory": resource.MustParse("512Mi"),
			"cpu":    resource.MustParse("600m"),
		},
	}

}
```

Right now the default is `small` with a limit of `medium` which would be `600Mhz/1Gi - 1.2Ghz/2Gi` - that's the same vis-a-vis CPU but twice as much RAM.

adamrdrew:
Tests were failing because my CPU requirements were using `M` instead of `m` resulting in significant CPU over commit. Also, fixed defaults to look like what we had before.

adamrdrew:
Something was wrong with ci today - I had to run the tests over and over again because it kept failing even though all the tests passed:

```
10:46:01 --- FAIL: kuttl (1157.48s)
10:46:01     --- PASS: kuttl/harness (0.00s)
10:46:01         --- PASS: kuttl/harness/test-kafka-managed (24.51s)
10:46:01         --- PASS: kuttl/harness/test-web-services (31.50s)
10:46:01         --- PASS: kuttl/harness/test-app-interface-s3 (31.55s)
10:46:01         --- PASS: kuttl/harness/test-sidecars (26.55s)
10:46:01         --- PASS: kuttl/harness/test-target-namespace (26.61s)
10:46:02         --- PASS: kuttl/harness/test-validator (39.68s)
10:46:02         --- PASS: kuttl/harness/test-rename-pod (26.07s)
10:46:02         --- PASS: kuttl/harness/test-recreate-strategy (20.01s)
10:46:02         --- PASS: kuttl/harness/test-service-mesh (32.16s)
10:46:02         --- PASS: kuttl/harness/test-node-port (26.11s)
10:46:02         --- PASS: kuttl/harness/test-pull-policy (26.12s)
10:46:02         --- PASS: kuttl/harness/test-multiple-app-endpoints (41.71s)
10:46:02         --- PASS: kuttl/harness/test-multi-app-interface-db (24.91s)
10:46:02         --- PASS: kuttl/harness/test-multi-db (51.18s)
10:46:02         --- PASS: kuttl/harness/test-metrics-operator (16.88s)
10:46:02         --- PASS: kuttl/harness/test-metrics-app-interface (16.78s)
10:46:02         --- PASS: kuttl/harness/test-minio-app (81.47s)
10:46:02         --- PASS: kuttl/harness/test-local-kafka (51.17s)
10:46:02         --- PASS: kuttl/harness/test-multi-db-shared (145.62s)
10:46:02         --- PASS: kuttl/harness/test-kafka-strimzi-topic-auth (206.08s)
10:46:02         --- PASS: kuttl/harness/test-kafka-strimzi-topic-deprecated-spec (239.25s)
10:46:02         --- PASS: kuttl/harness/test-kafka-strimzi-topic-basic (231.32s)
10:46:02         --- PASS: kuttl/harness/test-kafka-strimzi-pvc (175.12s)
10:46:02         --- PASS: kuttl/harness/test-ephemeral-gateway (177.03s)
10:46:02         --- PASS: kuttl/harness/test-kafka-strimzi-ephemeral (197.37s)
10:46:02         --- PASS: kuttl/harness/test-k8s-access (21.85s)
10:46:02         --- PASS: kuttl/harness/test-iqe-jobs-vault (20.06s)
10:46:02         --- PASS: kuttl/harness/test-iqe-jobs-selenium (19.75s)
10:46:02         --- PASS: kuttl/harness/test-iqe-jobs (31.51s)
10:46:02         --- PASS: kuttl/harness/test-kafka-app-autoscale (163.39s)
10:46:02         --- PASS: kuttl/harness/test-init-containers (42.11s)
10:46:02         --- PASS: kuttl/harness/test-kafka-app (165.02s)
10:46:02         --- PASS: kuttl/harness/test-init-container-image (19.21s)
10:46:02         --- PASS: kuttl/harness/test-db-pvc-tee (19.59s)
10:46:02         --- PASS: kuttl/harness/test-ff-app-interface (22.76s)
10:46:02         --- PASS: kuttl/harness/test-disabled-delete (16.24s)
10:46:02         --- PASS: kuttl/harness/test-elasticache (26.70s)
10:46:02         --- PASS: kuttl/harness/test-deployment-strategy (26.53s)
10:46:02         --- PASS: kuttl/harness/test-db-resource-tee (18.79s)
10:46:02         --- PASS: kuttl/harness/test-ff-local (159.40s)
10:46:02         --- PASS: kuttl/harness/test-disabled (124.96s)
10:46:02         --- PASS: kuttl/harness/test-cyndi-local-db (16.83s)
10:46:02         --- PASS: kuttl/harness/test-clowder-jobs (113.80s)
10:46:02         --- PASS: kuttl/harness/test-basic-app (27.67s)
10:46:02         --- PASS: kuttl/harness/test-autoscaler (18.37s)
10:46:02         --- PASS: kuttl/harness/test-cyndi-strimzi (182.42s)
10:46:02 FAIL
10:46:02 KUTTL_RESULT=$?
10:46:02 + KUTTL_RESULT=1
```

It seemed like one of the scripts couldn't make a directory or couldn't save a file or something:

```
10:46:02 + KUTTL_RESULT=1
10:46:02 mv kuttl-test.xml artifacts/junit-kuttl.xml
10:46:02 + mv kuttl-test.xml artifacts/junit-kuttl.xml
10:46:02 mv: cannot stat 'kuttl-test.xml': No such file or directory


10:46:01     harness.go:511: cleaning up
10:46:01     harness.go:511: cleaning up
10:46:01     harness.go:593: fatal error writing report: mkdir : no such file or directory
```

Re-ran and re-ran until eventually it worked after 4 or 5 hours.

adamrdrew:
Alight, we're green on tests so I think this is ready for review.

controllers/cloud.redhat.com/providers/sizing/sizing.go
```diff
@@ -0,0 +1,151 @@
...
+
+//Get the default volume capacity
+func GetDefaultVolCapacity() string {
+	return getVolSizeToCapacityMap()[GetDefaultSizeVol()]
+}
+
+//Get resource requirements - request and limits - for a given size
+func GetResourceRequirementsForSize(tShirtSize string) core.ResourceRequirements {
+	requestSize := useDefaultIfEmptySize(tShirtSize, GetDefaultSizeCPURAM())
+	cpu := getCPUSizeToCapacityMap()
```
psav, line 62: Why return the entire map here, when the function could accept an arg and be used directly on line 67, 68 etc?

controllers/cloud.redhat.com/providers/sizing/sizing.go
```diff
@@ -0,0 +1,151 @@
...
+
+//For a givin vol size get the capacity. Providing "" gets the default.
+func GetVolCapacityForSize(size string) string {
+	requestSize := useDefaultIfEmptySize(size, GetDefaultSizeVol())
+	return getVolSizeToCapacityMap()[requestSize]
+}
+
+//Accepts 2 sizes. Returns true if first size is larger than second
+func IsSizeLarger(capacityA string, capacityB string) bool {
+	capacities := map[string]int{
```
psav, line 85: This is slightly inefficient since this map will be created every time this function is invoked. Perhaps consider making this a top level definition.

controllers/cloud.redhat.com/providers/sizing/sizing.go
```diff
@@ -0,0 +1,151 @@
...
+		XLARGE: 4,
+	}
+	return capacities[capacityA] > capacities[capacityB]
+}
+
+// Private methods
+
+//Get a map of CPU T-Shirt sizes to capacities
+func getCPUSizeToCapacityMap() map[string]string {
+	return map[string]string{
```
psav, line 99: Again here, there is no real reason to return these as functions, they could be const definitions and referred to as such

```go

SizeMap := map[string]string{
  SMALL: "600m"
  ...
}

func getCPUSizeToCapaticy(string size){
  return SizeMap[size]
}
```

However, then the function is perhaps superfluous as you _could_ just interrogate the map directly. Perhaps you feel that is not a good API, either way, creating the maps as consts will be more efficient.

controllers/cloud.redhat.com/version.txt
```diff
@@ -1 +1 @@
-master
\ No newline at end of file
+v0.30.0-26-g1b4289d
```
psav, line 1: This file should not be committed/updated `make pre-push` could help here

psav (changes requested):

controllers/cloud.redhat.com/providers/sizing/sizing.go
adamrdrew: See my response to `getCPUSizeToCapacityMap()` because its the same pattern there.

controllers/cloud.redhat.com/providers/sizing/sizing.go
adamrdrew: Yeah, so I figured you'd ask about this. I don't love it, but lemme tell you what I was thinking and you can tell me what you think the right way to handle this is.

The problem is that Go doesn't support const or otherwise immutable composite types. You can't have a const map for example. So the problem I wanted to solve is I wanted these maps to be immutable; the recommendations I found for how to do this in golang is to create a method that returns a map. The map returned is still mutable obviously but at least no code can go and change the value for everyone else because the map instantiation is encapsulated in the method call.

So, that's why you see all these methods in this code that define maps and either return them or just pull something off them. I think considering Go's limits this is probably OK, but here are some ways we could change it if you like, but they're all tradeoffs:

- Just define top level maps. More efficient but more dangerous.
- Look and see if there's any kind of memoization in Go - if so we could get the safety with more efficiency
- Some other completely different way I don't know about!

Let me know what you think. Based on what I know about Go at this point I do think this is the best implementation - but I'm new to the lang so I'm very much open to a better implementation idea.

adamrdrew: I found multiple resources saying the same thing but this sort of article is what I based the approach on https://medium.com/bootdotdev/how-to-global-constant-maps-and-slices-in-go-30bf3eaf10db

adamrdrew: We could use something like this : https://github.com/kofalt/go-memoize but I don't think its a good fit. I have to imagine that the memoization code is less efficient than instantiating a map. Memoization is usually used for caching expensive methods; creating a map, though more costly than simply referencing one, should be cheaper than caching it and all the complexity that affords.

So upon further review I do think my implementation is best considering Go's limitations - but, again, I am cognizant of my own ignorance so if you know a way to solve this I'm happy to take it!

psav: OK, so, I think my preference here is this. Top level map, but not exported, and ensure we put this in a new module. This then means that nothing outside our module can access the map directly and that our functions to return a particular size are already inside that module, and hence we should be bale to trust what is in that small module.

controllers/cloud.redhat.com/version.txt
adamrdrew: I promise that at some point in the next few years I'll get this

controllers/cloud.redhat.com/providers/sizing/sizing.go
adamrdrew: Cool - I think this actually tidies things up a bit.

adamrdrew:
Alright, I think this is ready for re-review!

adamrdrew:
Rebased from master as stuff over there had changed.

PR #598, opened 2022-04-21
Update caddy plugin image

bsquizz:
This updates to a caddy plugin image that uses the latest crcauthlib -- which sends `auth_type` in the identity header.


BlakeHolifield (approved):

bsquizz:
Tested on crcd cluster via updating the clowder ConfigMap -- working

PR #619, opened 2022-05-24
[RHCLOUD-19621] Add external listener to strimzi kafka resource

ckyrouac:
This adds an external listener to the Kafka instance in ephemeral. This will be useful when forwarding the port to connect to Kafka locally. By using the external port (9094), Kafka can be connected at `localhost:9094` without requiring a DNS mapping in /etc/hosts.


subpop:
Is this still waiting on anything? I'd love to see this added as a feature to clowder.

PR #621, opened 2022-05-26
[RHCLOUD-19605] Add kubelinter ignore annotations on CronJob/Job resources

bsquizz:
App-SRE runs kubelinter which complains if a liveness or readiness probe is not present on any pod. Apparently it's not smart enough to tell if the pod was created via a Job or CronJob -- so we will have Clowder add the proper kubelinter "ignore" annotation onto our Job/CronJob pod templates.


controllers/cloud.redhat.com/providers/cronjob/impl.go
```diff
@@ -166,6 +166,10 @@ func applyCronCronJob(app *crd.ClowdApp, env *crd.ClowdEnvironment, cj *batch.Cr
 	labels["pod"] = nn.Name
 	app.SetObjectMeta(cj, crd.Name(nn.Name), crd.Labels(labels))
 
+	// add kubelinter annotations to ignore liveness/readiness probes on CronJobs
+	pt.ObjectMeta.Annotations["ignore-check.kube-linter.io/no-liveness-probe"] = "probes not required on Job pods"
+	pt.ObjectMeta.Annotations["ignore-check.kube-linter.io/no-readiness-probe"] = "probes not required on Job pods"
```
gburges, line 171: can you change it to "probes not required for CronJob pods" ? just to please my ADHD :)

bsquizz: I used this wording because technically a CronJob creates a Job :P

PR #622, opened 2022-05-27
doc: use github API to get the latest release

subpop:
Use GitHub API to get the download URL of the latest release.


subpop:
Oh I didn't realize that. I'll open an MR to get that added.

PR #627, opened 2022-06-10
[RHCLOUD-19829] Fixed logging type

psav:


cryptomilk:
We are hitting this issue that the logging type isn't set. This could be fixed since 2022.

bundle/tests/scorecard/kuttl/test-log-none/02-json-asserts.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- script: sleep 5
+- script: kubectl get secret --namespace=test-log-none puptoo -o json > /tmp/test-log-none
+- script: jq -r '.data["cdappconfig.json"]' < /tmp/test-log-none | base64 -d > /tmp/test-log-none-json
+
+- script: jq -r '.logging.type == "null"' -e < /tmp/test-log-none-json
```
bsquizz, line 9: I thought we left null keys absent these days, but I can see this has been in place in the code for awhile so probably not worth changing just for this simple fix.

bsquizz (approved):

PR #630, opened 2022-06-15
Bump MBOP version to add support for org_id

Victoremepunto:


Victoremepunto:
@vbelchio I think the Clowder pipelines are busted for good due to an ongoing issue - so we're out of options to make this pipeline to run :(

psav:
This is now done :)

PR #632, opened 2022-06-15
[RHCLOUD-19663] Implement Simple HPA for clowder

adamrdrew:
This patch adds simple HPA (k8s included HPA) to Clowder. Configuration is simple, requiring only an `autoScaler` mode of `simple` on a ClowdEnvironment as well as new `simpleAutoScaler` options for Deployments. Scaling can be performed on CPU and RAM resources reaching thresholds defined as either average utilization or an average value (or both!)

Examples:
```
  - name: highcpu
    podSpec:
      image: quay.io/rh_ee_addrew/highload:latest
    simpleAutoScaler:
      replicas:
        min: 1
        max: 4
      cpu:
        scaleAtUtilization: 80
```
The above deployment will scale to a maximum of 4 replicas as each reaches an average CPU utilization of 80%

```
  - name: highram
    podSpec:
      image: quay.io/psav/clowder-hello
    simpleAutoScaler:
      replicas:
        min: 1
        max: 4
      ram:
        scaleAtValue: 1Mi
```
The above deployment will scale to a maximum of 4 replicas when RAM utlization reaches an average value of 1Mi.

A kuttl test that validates the functionality is included that observes replica counts as assertions.

This is a fairly large change so I'm sending the code up now even though I intend to amend the PR with further changes to add unit tests. 

The HPA functionality requires metrics server to work. I've tried to update docs, scripts, etc that run minikube to include metrics-server wherever possible, but I could have missed something - we'll see. I also have no idea if this works on open shift yet because of that metrics-server dependency.


adamrdrew:
```
 --- FAIL: kuttl/harness/test-ephemeral-gateway (403.89s)
```

adamrdrew:
```
15:12:13         --- FAIL: kuttl/harness/test-kafka-strimzi-topic-basic (404.56s)
15:12:13         --- FAIL: kuttl/harness/test-kafka-strimzi-ephemeral (405.26s)
15:12:13         --- FAIL: kuttl/harness/test-kafka-strimzi-pvc (405.62s)
```

:shrug: :question: 

We had a kafka / ephemeral outage yesterday so maybe it was that? Let's /retest again!

adamrdrew:
All my clowder PRs are failign with the same tests (listed above) failing. Trouble is afoot at the cirlcle k.

tests/kuttl/test-autoscaler-simple/01-pods.yaml
```diff
@@ -0,0 +1,95 @@
...
+    simpleAutoScaler:
+      replicas:
+        min: 1
+        max: 4
+      cpu:
+        scaleAtUtilization: 80
+  - name: nohpa
+    podSpec:
+      image: quay.io/rh_ee_addrew/highload:latest
+---
```
psav, line 70: The pull secrets are superfluous here

tests/kuttl/test-autoscaler-simple/03-delete.yaml
```diff
@@ -0,0 +1,13 @@
+---
```
psav, line 1: Please rename to `02-delete.yaml`

psav (changes requested):

psav:
We should also ensure that there is a validating webhook configuration that prevents trying to set simple and the pass through or keda at the same time

adamrdrew:
> We should also ensure that there is a validating webhook configuration that prevents trying to set simple and the pass through or keda at the same time

Now that's spicy - I don't know how to do that. I'll poke around and see if I can figure it out.

adamrdrew:
Wait, the way its implemented you can't set multiple HPA types on a given `ClowdEnvironment` :

```
kind: ClowdEnvironment
metadata:
  name: test-autoscaler-simple
spec:
  targetNamespace: test-autoscaler-simple
  providers:
    autoScaler:
      mode: simple
```

You set the `ClowdEnvironment` `autoScaler` mode to `simple` and then you set `simpleAutoScaler` on any deployments you want in that env's apps. So, based on that you shouldn't be able to ever have 2 auto scaler modes active in a given environment. 

Is that good enough? Or do we need to worry about one `ClowdEnv` using keda and another using simple. I would think that would be OK.

adamrdrew:
Found an old option in clowdapp_types.go under `AutoScaler` that was left in from earlier in development but doesn't actually function. Also, pre-push.

adamrdrew:
Alright mate, this turned out to be a *slightly* larger change than I think we expected in review. But, I think the end result is significantly better in basically every way. Waiting on tests, will mop up any problems and should be g2g tomorrow.

adamrdrew:
Ugh, I'm going to change it again. I changed AutoScaler.Mode of type string to `AutoScaler.Enabled` of type boolean because I thought it was cleaner and more accurate. But I notice a couple of things:

1. Most ClowdEnvironment options are set with a `mode` flag that is set to `none` if we don't want it active
2. As I have it now this is a breaking change which is probably not a good thing

So, at least one more commit incoming. Grr.

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -200,7 +200,9 @@ type Deployment struct {
 	K8sAccessLevel K8sAccessLevel `json:"k8sAccessLevel,omitempty"`
 
 	// AutoScaler defines the configuration for the auto scaler
-	AutoScaler *AutoScaler `json:"autoScaler,omitempty"`
+	AutoScalerKeda *AutoScalerKeda `json:"autoScalerKeda,omitempty"`
```
psav, line 203: We can't change the name because it is already a part of the spec, we need to perhaps have both for a while, and deprecate one of them.

controllers/cloud.redhat.com/providers/autoscaler/router.go
```diff
@@ -0,0 +1,33 @@
...
+type autoScaleProviderRouter struct {
+	providers.Provider
+	Config config.DatabaseConfig
+}
+
+func NewAutoScaleProviderRouter(p *providers.Provider) (providers.ClowderProvider, error) {
+	return &autoScaleProviderRouter{Provider: *p}, nil
+}
+
+func (db *autoScaleProviderRouter) Provide(app *crd.ClowdApp, c *config.AppConfig) error {
```
psav, line 19: Let's not use `db` here, I imagine this was from a copy pasta

controllers/cloud.redhat.com/providers/autoscaler/router.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	crd "github.com/RedHatInsights/clowder/apis/cloud.redhat.com/v1alpha1"
+	"github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/config"
+	"github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/providers"
+)
+
+//autoScaleProviderRouter is a wrapper for the different autoscaler providers.
+type autoScaleProviderRouter struct {
+	providers.Provider
+	Config config.DatabaseConfig
```
psav, line 12: Database Config?

controllers/cloud.redhat.com/providers/autoscaler/router.go
```diff
@@ -0,0 +1,33 @@
...
+func NewAutoScaleProviderRouter(p *providers.Provider) (providers.ClowderProvider, error) {
+	return &autoScaleProviderRouter{Provider: *p}, nil
+}
+
+func (db *autoScaleProviderRouter) Provide(app *crd.ClowdApp, c *config.AppConfig) error {
+	for _, deployment := range app.Spec.Deployments {
+		//If we find a SimpleAutoScaler config create one
+		if deployment.AutoScalerSimple != nil {
+			ProvideSimpleAutoScaler(app, c, &db.Provider, deployment)
+			continue
```
psav, line 24: `continue` is superfluous here I think

controllers/cloud.redhat.com/providers/autoscaler/router.go
```diff
@@ -0,0 +1,33 @@
...
+		}
+		//If we find a SimpleAutoScaler config create one
+		if deployment.AutoScalerKeda != nil {
```
psav, line 27: Will need to handle the fact that we may have the old API spec field here still

controllers/cloud.redhat.com/providers/autoscaler/router.go
```diff
@@ -0,0 +1,33 @@
...
+
+func NewAutoScaleProviderRouter(p *providers.Provider) (providers.ClowderProvider, error) {
+	return &autoScaleProviderRouter{Provider: *p}, nil
+}
+
+func (db *autoScaleProviderRouter) Provide(app *crd.ClowdApp, c *config.AppConfig) error {
+	for _, deployment := range app.Spec.Deployments {
+		//If we find a SimpleAutoScaler config create one
+		if deployment.AutoScalerSimple != nil {
+			ProvideSimpleAutoScaler(app, c, &db.Provider, deployment)
```
psav, line 23: The error that could be returned from `ProvideSimpleAutoScaler` is ignored

controllers/cloud.redhat.com/providers/autoscaler/router.go
```diff
@@ -0,0 +1,33 @@
...
+			continue
+		}
+		//If we find a SimpleAutoScaler config create one
+		if deployment.AutoScalerKeda != nil {
+			ProvideKedaAutoScaler(app, c, &db.Provider, deployment)
```
psav, line 28: The error that could be returned from ProvideKedaAutoScaler is ignored

controllers/cloud.redhat.com/providers/autoscaler/router.go
```diff
@@ -0,0 +1,33 @@
...
+			continue
```
psav, line 29: `continue` is superfluous here I think

controllers/cloud.redhat.com/providers/autoscaler/simple.go
```diff
@@ -0,0 +1,148 @@
...
+	apps "k8s.io/api/apps/v1"
+	v2 "k8s.io/api/autoscaling/v2"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func ProvideSimpleAutoScaler(app *crd.ClowdApp, appConfig *config.AppConfig, sp *providers.Provider, deployment crd.Deployment) error {
+	cachedDeployment, err := getDeploymentFromCache(&deployment, app, sp)
+	if err != nil {
+		log.Println(err)
```
psav, line 22: Logging here is unnecessary. The standard is to return a wrapped error message. These are then passed all the way down the chain, and logging only takes place in the reconcile function. For example
```
if err != nil {
  return errors.Wrap("could not get deployment from cache", err)
}
```

controllers/cloud.redhat.com/providers/autoscaler/simple.go
```diff
@@ -0,0 +1,148 @@
...
+		return err
+	}
+	deploymentHPA := makeDeployemntSimpleHPA(&deployment, app, appConfig, cachedDeployment)
```
psav, line 25: `makeDeploy(e)mntSimpleHPA`

controllers/cloud.redhat.com/providers/autoscaler/simple.go
```diff
@@ -0,0 +1,148 @@
...
+func getDeploymentFromCache(clowdDeployment *crd.Deployment, app *crd.ClowdApp, sp *providers.Provider) (*apps.Deployment, error) {
+	nn := app.GetDeploymentNamespacedName(clowdDeployment)
+	d := &apps.Deployment{}
+	if err := sp.Cache.Get(deployProvider.CoreDeployment, d, nn); err != nil {
+		return d, err
+	}
+	return d, nil
+}
+
+func makeDeployemntSimpleHPA(deployment *crd.Deployment, app *crd.ClowdApp, appConfig *config.AppConfig, coreDeployment *apps.Deployment) deployemntSimpleHPA {
```
psav, line 46: As a factory func I think we use the word `new` more than `make` to denote the factory creation function. Also the name is a little ambiguous, consider a rename to something like (because my suggestion is terrible) HPAConfigurator. Which shows more that we are returning something that will DO more processing.

controllers/cloud.redhat.com/providers/autoscaler/simple.go
```diff
@@ -0,0 +1,148 @@
...
+func ProvideSimpleAutoScaler(app *crd.ClowdApp, appConfig *config.AppConfig, sp *providers.Provider, deployment crd.Deployment) error {
+	cachedDeployment, err := getDeploymentFromCache(&deployment, app, sp)
+	if err != nil {
+		log.Println(err)
+		return err
+	}
+	deploymentHPA := makeDeployemntSimpleHPA(&deployment, app, appConfig, cachedDeployment)
+	hpaResource := deploymentHPA.getResource()
+
+	err = sp.Client.Create(sp.Ctx, &hpaResource)
```
psav, line 28: This appears to be using the client directly to create resource and hence it will not be tracked by Clowder, the resource cache should be used here so that Clowder can clean up the HPA if the mode is changed or teh app is deleted.

controllers/cloud.redhat.com/providers/autoscaler/simple.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		fmt.Println("HPA Error: ", err)
```
psav, line 30: Don't print errors, wrap and return.

controllers/cloud.redhat.com/providers/autoscaler/simple.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	return nil
+}
+
+//Get the core apps.Deployment from the provider cache
+//This is in AutoScalerSimpleProvider because we need access to the provider cache
+func getDeploymentFromCache(clowdDeployment *crd.Deployment, app *crd.ClowdApp, sp *providers.Provider) (*apps.Deployment, error) {
```
psav, line 37: Throughout the code, when we `get/generate` a k8s resource for the first time, we tend to follow the k8s pattern of population, rather than returning it. Throughout the Clowder code base, particularly when using the cache and clients you will see that we create an empty object, and send a pointer for population, rather than expect an object to be returned to us.

controllers/cloud.redhat.com/providers/autoscaler/simple.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+type deployemntSimpleHPA struct {
+	deployment     *crd.Deployment
+	app            *crd.ClowdApp
+	appConfig      *config.AppConfig
+	coreDeployment *apps.Deployment
+}
+
+func (d *deployemntSimpleHPA) getResource() v2.HorizontalPodAutoscaler {
```
psav, line 62: We should probably be passing pointers to these structs, and not creating them each time.

controllers/cloud.redhat.com/providers/autoscaler/simple.go
```diff
@@ -0,0 +1,148 @@
...
+	hpa := d.makeHPA()
+	hpa.Spec.Metrics = d.makeMetricsSpecs()
+	return hpa
+}
+
+func (d *deployemntSimpleHPA) makeHPA() v2.HorizontalPodAutoscaler {
+	hpa := v2.HorizontalPodAutoscaler{
+		TypeMeta: metav1.TypeMeta{
```
psav, line 70: Kind shouldn't need setting - it is done by the client.

controllers/cloud.redhat.com/providers/autoscaler/simple.go
```diff
@@ -0,0 +1,148 @@
...
+			Kind:       "HorizontalPodAutoscaler",
+			APIVersion: "autoscaling/v2",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      d.app.Name + "-" + d.coreDeployment.Name + "-" + "hpa",
+			Namespace: d.coreDeployment.Namespace,
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: d.coreDeployment.TypeMeta.APIVersion,
```
psav, line 79: I would have the owner of the object here be the ClowdApp, is there a particular reason not to do this?

controllers/cloud.redhat.com/providers/autoscaler/simple.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (d *deployemntSimpleHPA) makeHPA() v2.HorizontalPodAutoscaler {
+	hpa := v2.HorizontalPodAutoscaler{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "HorizontalPodAutoscaler",
+			APIVersion: "autoscaling/v2",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      d.app.Name + "-" + d.coreDeployment.Name + "-" + "hpa",
```
psav, line 75: Can we use `fmt.Sprintf` usage 
`fmt.Sprintf("%s-%s-hpa", d.app.Name, d.coreDeployment.Name)`

psav: We also have a function called `GetDeploymentNamespacedName` which takes a deployment and gives you a properly formatted name - perhaps consider using this.

controllers/cloud.redhat.com/providers/autoscaler/wrapper.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/providers"
+)
+
+type autoScaleProvider struct {
+	providers.Provider
+	Config config.DatabaseConfig
+}
+
+// NewNoneDBProvider returns a new none db provider object.
+func NewAutoScaleWrapper(p *providers.Provider) (providers.ClowderProvider, error) {
```
psav, line 15: I don't remember seeing this used anywhere - I think the file is perhaps old and mistakenly added here.

tests/kuttl/test-autoscaler-simple/01-pods.yaml
psav: The are still here....

psav (changes requested):
A few changes on this one :)

controllers/cloud.redhat.com/providers/autoscaler/wrapper.go
adamrdrew: Yup, you're right. Removed.

controllers/cloud.redhat.com/providers/autoscaler/router.go
adamrdrew: More copy/paste. Messy, messy. I wasn't sure what this was or if something else depended on it. Should have looked instead of just leaving it. Removed in next push.

controllers/cloud.redhat.com/providers/autoscaler/router.go
adamrdrew: I *believe* this is fixed by the change you suggested above where I had changed `AutoScaler` to `AutoScalerKeda` - now that it is `AutoScaler` again this should be good. Will not resolve convo until you give :+1:

controllers/cloud.redhat.com/providers/autoscaler/simple.go
adamrdrew: Good, I knew I should have handled this better but wasn't sure how. Add in next push.

controllers/cloud.redhat.com/providers/autoscaler/router.go
adamrdrew: Added in next push.

controllers/cloud.redhat.com/providers/autoscaler/simple.go
adamrdrew: Weapons grade ooftonium. Fixed in next push.

controllers/cloud.redhat.com/providers/autoscaler/simple.go
adamrdrew: Yeah I was trying to be idiomatic - I saw go devs use `make` in lots of places and sort of copied that. But if `new` is better that's good with me. Changed in next push.

Also, I changed the name of `deployemntSimpleHPA` to `simpleHPAMaker` because as you point out that's more accurate.

controllers/cloud.redhat.com/providers/autoscaler/router.go
adamrdrew: Fixed. I'm returning all errors now.

controllers/cloud.redhat.com/providers/autoscaler/simple.go
adamrdrew: Changed to `fmt.Sprintf` to keep it simple but will keep that function in mind for the future. Will leave this open for :+1:  or :-1:

controllers/cloud.redhat.com/providers/autoscaler/simple.go
adamrdrew: Looking at:

```
			ScaleTargetRef: v2.CrossVersionObjectReference{
				APIVersion: d.coreDeployment.APIVersion,
				Kind:       d.coreDeployment.Kind,
				Name:       d.coreDeployment.Name,
			},
```

I actually think you are totally right. The owner ref could be clowd app but target can be a deployment.

adamrdrew:
Today did not go well. Ran into problem after problem. I think we're going to need to team up on this. Stuff I hit:

- HPA doesn't want to be owned by ClowdApp; it complains about not being able to find the target deployments
- wrapping errors in `ProvideSimpleAutoScaler` causes the HPAs to never get created

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -199,9 +199,11 @@ type Deployment struct {
 	// K8sAccessLevel defines the level of access for this deployment
 	K8sAccessLevel K8sAccessLevel `json:"k8sAccessLevel,omitempty"`
 
-	// AutoScaler defines the configuration for the auto scaler
+	// AutoScaler defines the configuration for the Keda auto scaler
```
psav, line 202: So leaving as is for now - am cool with that

controllers/cloud.redhat.com/providers/autoscaler/simple.go
adamrdrew: This is done in the next push

adamrdrew:
OK I *believe* I hit all your input. Pretty happy with this last set of changes: much refine, very quality. Passing tests locally so we'll see if Jenkins is happy. If he's cool then I'll ping you.

psav (approved):

PR #636, opened 2022-06-17
[RHCLOUD-19111] - Change to rhc-osdk-utils status checking

adamrdrew:
This patch changes the resource status checking system to the features from rhc-osdk-utils


adamrdrew:
```
13:33:01         --- FAIL: kuttl/harness/test-kafka-strimzi-topic-basic (408.52s)
13:33:01         --- FAIL: kuttl/harness/test-kafka-strimzi-pvc (404.18s)
13:33:01         --- FAIL: kuttl/harness/test-kafka-strimzi-ephemeral (404.34s)
```

So, problem is my other clowder PR - which doesn't share this code - is failing in exactly the same tests and only these tests. See here https://github.com/RedHatInsights/clowder/pull/632#issuecomment-1158802056

Not sure what to think here.

adamrdrew:
```
getprov: kafka: unable to determine kafka broker addresses
```
Lots of these in the logs.

adamrdrew:
Rebased.

Also, we talked about adding an explicit test for kafka counts but as I started to try and implement it, it occurred to me we probably don't need to. If the status checking code didn't work we'd end up with failed reconciliations - it would manifest in every or nearly every test. If there was a problem in the code I think the existing test suite (kuttl) would catch it. We know the status checking code itself works because of the test suite in rhc-osdk-utils, so the kuttl test suite passing should leave us feeling good about this, right?

Lemme know what you think.

adamrdrew:
Rebased. Testing now. This should be gtg.

psav (approved):

PR #638, opened 2022-06-24
[BRANCH PR] Loop improvement idea

psav:
@adamrdrew Just thinking of future expandability - let me know if this works and what you think


adamrdrew:
:+1: I like.

PR #650, opened 2022-07-01
Removed two references to old bundle make command.

maknop:


psav:
Or just sign you commit :)

PR #656, opened 2022-07-06
RHCLOUD-20301 Allow configurable images for web provider via ClowdEnvironment

bsquizz:
Allows the images for caddy, mocktitlements, mbop, and keycloak to be configurable in the ClowdEnvironment. The logic is as follows:

When looking up the image for each of these components:
1. Check if it is defined in the ClowdEnvironment and use that if so
2. Check if it is defined in the LoadedConfig and use that if so
3. Use a default hard-coded value


controllers/cloud.redhat.com/providers/utils/utils.go
```diff
@@ -154,3 +164,55 @@ func MakeLocalDBService(s *core.Service, nn types.NamespacedName, baseResource o
 func MakeLocalDBPVC(pvc *core.PersistentVolumeClaim, nn types.NamespacedName, baseResource obj.ClowdObject, capacity string) {
 	utils.MakePVC(pvc, nn, providers.Labels{"service": "db", "app": baseResource.GetClowdName()}, capacity, baseResource)
 }
+
+// GetCaddyImage returns the caddy image to use in a given environment
+func GetCaddyImage(env *crd.ClowdEnvironment) string {
```
psav, line 169: Can we put these in the web provider?

psav (changes requested):

controllers/cloud.redhat.com/providers/web/impl.go
```diff
@@ -249,11 +245,8 @@ func makeKeycloak(o obj.ClowdObject, objMap providers.ObjectMap, usePVC bool, no
 		FailureThreshold:    3,
 	}
 
-	image := fmt.Sprintf("quay.io/keycloak/keycloak:%s", KEYCLOAK_VERSION)
-
-	if clowderconfig.LoadedConfig.Images.Keycloak != "" {
-		image = clowderconfig.LoadedConfig.Images.Keycloak
-	}
+	env := o.(*crd.ClowdEnvironment)
```
psav, line 248: This type assert could fail, and it will panic if it does. As the code stands a) it shouldn't and b) MakeCachedComponent doesn't take an error for the `fn` argument so I can understand why you don't try to return an error here. We should fix this SOON!

controllers/cloud.redhat.com/providers/web/impl.go
```diff
@@ -397,11 +390,8 @@ func makeBOP(o obj.ClowdObject, objMap providers.ObjectMap, usePVC bool, nodePor
 		TimeoutSeconds:      2,
 	}
 
-	image := IMAGE_WEB_MBOP
-
-	if clowderconfig.LoadedConfig.Images.MBOP != "" {
-		image = clowderconfig.LoadedConfig.Images.MBOP
-	}
+	env := o.(*crd.ClowdEnvironment)
```
psav, line 393: As above

controllers/cloud.redhat.com/providers/web/impl.go
```diff
@@ -462,8 +452,11 @@ func makeMocktitlements(o obj.ClowdObject, objMap providers.ObjectMap, usePVC bo
 
 	dd.Spec.Template.ObjectMeta.Labels = labels
 
+	env := o.(*crd.ClowdEnvironment)
```
psav, line 455: As above

controllers/cloud.redhat.com/providers/web/impl.go
```diff
@@ -540,15 +533,11 @@ func makeMocktitlements(o obj.ClowdObject, objMap providers.ObjectMap, usePVC bo
 		TimeoutSeconds:      2,
 	}
 
-	image := IMAGE_WEB_MOCKTITLEMENTS
-
-	if clowderconfig.LoadedConfig.Images.MBOP != "" {
-		image = clowderconfig.LoadedConfig.Images.MBOP
-	}
+	mocktitlementsImage := provutils.GetMocktitlementsImage(env)
```
psav, line 536: Thanks for fixing my bug here

psav (approved):
Some comments to be fixed soon - but overall looks good

controllers/cloud.redhat.com/providers/utils/utils.go
bsquizz: Sure. Still not clear what you mean when you say put these in the web provider though.

You want them to be receiver methods on that provider? Or just locate these funcs in the web provider package and adjust the imports everywhere else?

PR #660, opened 2022-07-11
Apply Subscription

psav:


bsquizz (approved):

bsquizz:
If that kuttl assert is passing, seems pretty low risk. Only thing that popped into my head: are the versions of prom operator the same in all clusters? Is the one you're installing in minikube the same as what is running elsewhere?

PR #672, opened 2022-07-22
Ephemeral Managed Kafka Provider refactor

adamrdrew:
Here's a proposal for a refactor of the new ephemeral managed kafka provider code. 

There are no logic changes here. Rather, this refactor focuses on organization. The goal is to keep methods small (~ 10 lines or less) and single purpose with names that clearly communicate what they do. I think the result is easier to read, reason about, and maintain.

Let me know what you think!


adamrdrew:
Note, as of creating this PR I still haven't refactored ephemProcessTopicValues - I'll do that I just wanted to push something up before it got too late

adamrdrew:
OK, A first pass at this proposed refactor is done (I think, I could have missed something). Have a look. Let's discuss it next week. Have a great weekend!

adamrdrew:
I ran a diff between f6a40d6eb626cbd742b100952927e3085fec804d  9c5f9229324af1484f059e84994289831e5fe447  which *should* get me the changes you made to your branch between when I started and what you've done since. There's no way for me to merge them with git because the code paths are different in this patch so I added them by hand. It wasn't much, just partition limits.

PR #675, opened 2022-07-26
[RHCLOUD-20060] - This adds custom annotations to the podspec

adamrdrew:
I missed part of the customer request. We need annotations on podspec. This patch adds that along with a kuttl test.


psav (changes requested):
This looks reasonable to me - but let's remove all the env vars in the tests to keep them focused.

psav (approved):

PR #687, opened 2022-08-10
[RHCLOUD-15674] Change missing dependency errors to be more explicit

adamrdrew:
This patch changes missing dependency errors from a wrapper around a list of string maps to a wrapper around a list of new MissingDependency structs that have fields for specific datapoints. Both MissingDependency and MissingDepdencies are enhanced with small convenience methods to make common tasks DRY. Changed all places in providers where we throw missing dep errors to use the new stuff. Also re-wrote the error text itself to make them more consistent and hopefully more helpful.


bsquizz:
Will take a look soon since I opened the card related to this :)

controllers/cloud.redhat.com/providers/inmemorydb/elasticache.go
```diff
@@ -58,11 +58,12 @@ func (e *elasticache) Provide(app *crd.ClowdApp, config *config.AppConfig) error
 	}
 
 	if !found {
-		return &errors.MissingDependencies{
-			MissingDeps: map[string][]string{
-				"in-memory-db-secret": {fmt.Sprintf("name: %s, namespace: %s", secretName, app.Namespace)},
-			},
-		}
+		missingDeps := errors.MakeMissingDependencies(errors.MissingDependency{
+			Source:  "elasticsearch",
```
BlakeHolifield, line 62: I think this should be "elasticache" or "inmemorydb"

BlakeHolifield (changes requested):
Quick typo check, otherwise I dig it!

controllers/cloud.redhat.com/providers/database/appinterface.go
```diff
@@ -149,11 +149,12 @@ func GetDbConfig(ctx context.Context, pClient client.Client, namespace, appName,
...
 		if matched == (config.DatabaseConfigContainer{}) {
-			return nil, &errors.MissingDependencies{
-				MissingDeps: map[string][]string{
-					"database": {appName},
-				},
-			}
+			missingDep := errors.MakeMissingDependencies(errors.MissingDependency{
+				Source:  "database",
+				App:     appName,
+				Details: "No app interface database provider config found",
```
bsquizz, line 155: This can probably be addressed in a follow-up -- technically I think we could go deeper on the details here... If you look at `genDbConfigs` and `resolveDb` we are looking for a specific secret in the namespace that has specific details in there. We should be able to say something like `DB secret named '%s' not found in namespace '%s'` or something like that -- but need to read the logic of those functions more closely to determine what to sub in for the strings in that message :)

controllers/cloud.redhat.com/providers/kafka/appinterface.go
```diff
@@ -69,9 +69,12 @@ func validateKafkaTopic(ctx context.Context, cl client.Client, nn types.Namespac
 	err := cl.Get(ctx, nn, &t)
 
 	if err != nil {
-		return &errors.MissingDependencies{
-			MissingDeps: map[string][]string{"topics": {nn.Name}},
-		}
+		missingDeps := errors.MakeMissingDependencies(errors.MissingDependency{
+			Source:  "kafka",
+			App:     nn.Name,
+			Details: fmt.Sprintf("No topics found for app '%s' found in namespace '%s'", nn.Name, nn.Namespace),
```
bsquizz, line 75: It looks like we tried to `Get` a specific KafkaTopic resource up above -- is this description better?

```suggestion
			Details: fmt.Sprintf("No KafkaTopic named '%s' found in namespace '%s'", nn.Name, nn.Namespace),
```

controllers/cloud.redhat.com/providers/kafka/appinterface.go
```diff
@@ -69,9 +69,12 @@ func validateKafkaTopic(ctx context.Context, cl client.Client, nn types.Namespac
 	err := cl.Get(ctx, nn, &t)
 
 	if err != nil {
-		return &errors.MissingDependencies{
-			MissingDeps: map[string][]string{"topics": {nn.Name}},
-		}
+		missingDeps := errors.MakeMissingDependencies(errors.MissingDependency{
+			Source:  "kafka",
+			App:     nn.Name,
```
bsquizz, line 74: I believe `nn.Name` is the name of the KafkaTopic we tried to fetch, not the app's name.

controllers/cloud.redhat.com/providers/kafka/appinterface.go
```diff
@@ -92,11 +95,12 @@ func validateBrokerService(ctx context.Context, cl client.Client, nn types.Names
...
 
 	if err != nil {
-		errors.LogError(ctx, "kafka", errors.New("Cannot find kafka bootstrap service"))
-		qualifiedName := fmt.Sprintf("%s:%s", nn.Namespace, nn.Name)
-		return &errors.MissingDependencies{
-			MissingDeps: map[string][]string{"kafka": {qualifiedName}},
-		}
+		missingDeps := errors.MakeMissingDependencies(errors.MissingDependency{
+			Source:  "kafka",
+			App:     nn.Name,
```
bsquizz, line 100: I believe nn.Name is the name of the KafkaTopic we tried to fetch, not the app's name.

controllers/cloud.redhat.com/providers/kafka/appinterface.go
```diff
@@ -92,11 +95,12 @@ func validateBrokerService(ctx context.Context, cl client.Client, nn types.Names
...
+			Details: fmt.Sprintf("No kafka bootstrap service for app '%s' found in namespace '%s'", nn.Name, nn.Namespace),
```
bsquizz, line 101: Since we tried to 'Get' a `core.Service` resource with the client up above...

```suggestion
			Details: fmt.Sprintf("No Service named '%s' found in namespace '%s'", nn.Name, nn.Namespace),
```

controllers/cloud.redhat.com/providers/kafka/cyndi.go
```diff
@@ -42,15 +42,21 @@ func validateCyndiPipeline(
 	err := cl.Get(ctx, nn, &pipeline)
 
 	if err != nil {
-		return &errors.MissingDependencies{
-			MissingDeps: map[string][]string{"cyndiPipeline": {nn.Name}},
-		}
+		missingDeps := errors.MakeMissingDependencies(errors.MissingDependency{
+			Source:  "cyndi",
+			App:     nn.Name,
```
bsquizz, line 47:
```suggestion
			App:     app.Name,
```

controllers/cloud.redhat.com/providers/kafka/cyndi.go
```diff
@@ -42,15 +42,21 @@ func validateCyndiPipeline(
...
+			Details: fmt.Sprintf("No Cyndi pipeline found for app '%s' found in namespace '%s'", nn.Name, nn.Namespace),
```
bsquizz, line 48:
```suggestion
			Details: fmt.Sprintf("CyndiPipeline named '%s' not found in namespace '%s'", nn.Name, nn.Namespace),
```

controllers/cloud.redhat.com/providers/kafka/cyndi.go
```diff
@@ -42,15 +42,21 @@ func validateCyndiPipeline(
...
 	}
 
 	if pipeline.Status.ActiveTableName == "" {
-		return &errors.MissingDependencies{
-			MissingDeps: map[string][]string{"cyndiPipelineActiveTable": {nn.Name}},
-		}
+		missingDeps := errors.MakeMissingDependencies(errors.MissingDependency{
+			Source:  "cyndi",
+			App:     nn.Name,
+			Details: fmt.Sprintf("No Cyndi pipeline Active Table found for app '%s' found in namespace '%s'", nn.Name, nn.Namespace),
```
bsquizz, line 57:
```suggestion
			Details: fmt.Sprintf("CyndiPipeline '%s' in namespace '%s' has no 'active table' in status", nn.Name, nn.Namespace),
```

controllers/cloud.redhat.com/providers/kafka/cyndi.go
```diff
@@ -42,15 +42,21 @@ func validateCyndiPipeline(
...
+		return &missingDeps
 	}
 
 	if pipeline.Status.ActiveTableName == "" {
-		return &errors.MissingDependencies{
-			MissingDeps: map[string][]string{"cyndiPipelineActiveTable": {nn.Name}},
-		}
+		missingDeps := errors.MakeMissingDependencies(errors.MissingDependency{
+			Source:  "cyndi",
+			App:     nn.Name,
```
bsquizz, line 56:
```suggestion
			App:     app.Name,
```

bsquizz (changes requested):
Thanks for adding this! I spotted some small tweaks/corrections I think we need to make here.

adamrdrew:
Oof, a lot to fix. I'll send up a revision today.

controllers/cloud.redhat.com/providers/kafka/appinterface.go
adamrdrew: You're right - but I'm not sure how to get the app name here. I'm looking at it now - I likely can get it from the ctx some how but I don't immediately know how to do it.

adamrdrew: I amended the method signature to accept the clowd app.

controllers/cloud.redhat.com/providers/kafka/appinterface.go
adamrdrew: This one is more challenging. AFAICT there's no way for me to get a ClowdApp reference in this method without making changes more than one hop into the code. `appinterface.validateBrokerService` is called by `appinterface.NewAppInterface` which only has a reference to a Provider, which has no ClowdApp reference. And `appinterface.NewAppInterface` is `provider.GetKafka`, again with no ClowdApp reference. That's about as far as I'd be comfortable going to get this right. 

I'm going to make my next push without a resolution to this. I don't see a way to do it, so I'm open to suggestions.

adamrdrew: That said I did change the error to `Details: fmt.Sprintf("No kafka bootstrap service for KafkaTopic '%s' found in namespace '%s'", nn.Name, nn.Namespace),`

controllers/cloud.redhat.com/providers/kafka/appinterface.go
adamrdrew: Oh, lol - this invalidates my change above. Changed to this :)

adamrdrew:
As per the group discussion I recast the env kafka broker error as a normal error.

bsquizz (approved):

controllers/cloud.redhat.com/errors/errors.go
```diff
@@ -73,22 +73,38 @@ func Wrap(msg string, err error) *ClowderError {
 	return clowderErr
 }
 
-// MissingDependencies is a struct containing a map of string lists, it is intended to hold
-// information about missing dependencies. The key is a resource type, or a provider as a string,
-// and the list is a list of keys or items that are missing.
+//MissingDependency is a struct that holds information about a missing dependency
+type MissingDependency struct {
+	Source  string
+	App     string
```
psav, line 79: Is the app strictly necessary here? since we will always be throwing the error in the context of a specific application, and its place in the status field places it inside of an application, is the app name here redundant?

controllers/cloud.redhat.com/errors/errors.go
```diff
@@ -73,22 +73,38 @@ func Wrap(msg string, err error) *ClowderError {
...
+func MakeMissingDependencies(missingDep MissingDependency) MissingDependencies {
+	return MissingDependencies{
+		MissingDeps: []MissingDependency{missingDep},
+	}
+}
+
+//MissingDependencies is a struct that holds a list of MissingDependency structs
 type MissingDependencies struct {
-	MissingDeps map[string][]string
+	MissingDeps []MissingDependency
```
psav, line 96: I am pretty sure that this was a problem before, but as we are refactoring this, the missing deps can show up in a status field. Since this is a list the order should be determinate, but should we throw a sort in here to ensure that we don't get a string that can be different from the last time, such operations reduce the amount of reconciliations. I think there are potentially other places in the subsequent functions where we are picking up items from a loop. If those loops are looping over things which are not guaranteed to arrive in order, we may need to sort in there also to avoid creating reordered error lists.

controllers/cloud.redhat.com/errors/errors.go
```diff
@@ -73,22 +73,38 @@ func Wrap(msg string, err error) *ClowderError {
...
-		depList := strings.Join(vals, ",")
-		typeList = append(typeList, fmt.Sprintf("- %s: %s", t, depList))
+	for _, missingDep := range e.MissingDeps {
+		typeList = append(typeList, missingDep.ToString())
 	}
 
-	body := strings.Join(typeList, ", ")
+	body := strings.Join(typeList, "; ")
 
 	return fmt.Sprintf("Missing dependencies: [%s]", body)
```
psav, line 109: This used square brackets, I'm wondering if we should remove the square brackets from around here, and actually use them instead of the single quotes that we have used in the format strings.

controllers/cloud.redhat.com/providers/kafka/appinterface.go
```diff
@@ -40,7 +40,7 @@ func (a *appInterface) Provide(app *crd.ClowdApp, c *config.AppConfig) error {
 			Name:      topic.TopicName,
 		}
 
-		err := validateKafkaTopic(a.Ctx, a.Client, topicName)
+		err := validateKafkaTopic(a.Ctx, a.Client, topicName, app)
```
psav, line 43: With the above comment, this would make this change here unnecessary

psav (changes requested):
Just a few things I thought of, thanks @bsquizz for the other thorough review!

controllers/cloud.redhat.com/errors/errors.go
adamrdrew: I believe it is because the MissingDependency struct provides its own ToString method for producing the error message which is derives from the source, app, and details. This ensures the messages are consistent, and that this logic is only implemented once.

controllers/cloud.redhat.com/errors/errors.go
adamrdrew: I'm afraid I don't understand this input. What would I sort by? Would I add a timestamp to `MissingDepdency` and then when we iterate through the `MissingDependencies.MissingDeps` (in the `MissingDepdencies.Error()` method) sort by time before creating the error strings? What does this solve? I thought this code just produced log messages? 

That said, this got me thinking about the `MissingDepdencies` object itself and I'm not sure why it contains a list at all. When I was refactoring this code I wasn't thinking about the larger picture of like how these structs are used, I was just trying to take what was there and improve it while changing as little as possible. But, as I go through the code I see no place where we ever add another `MissingDependency` to `MissingDepdencies.MissingDeps`. Every single place we create a `MissingDepdencies` instance we just create it with one `MissingDependency` - and we never add anything to it. Heck, there isn't even a method for adding a dependency after instantiation, and you can bet I would have added a method rather than allow access to raw object state.

So, while I don't fully understand your input I wonder if we should just scrap `MissingDependencies` all together and just work with `MissingDependency` instances, and add the `Error` method to `MissingDepdency` and then refactor `errors.HandleError` to expect the `MissingDepdency` type instead of `MissingDependencies` - I wonder if maybe this system was intended to handle cases where a single `MissingDependencies` instance would have multiple `MissingDependency` instances popped into it, but looking at the code it doesn't appear this ever actually came true. If this isn't a real world use case we should remove it and simplify.

controllers/cloud.redhat.com/errors/errors.go
adamrdrew: My input re: the lists would impact here because seing as though we never actually concatenate anything as there's never more than one MissingDepdency per MissingDepdencies struct if we decide to unroll that we could just recast this as a simple string rather than trying to represent a concatenated list.

controllers/cloud.redhat.com/providers/kafka/appinterface.go
adamrdrew: Yeah, I definitely think this is a small price to pay for consistency and DRY. I could make this a pointer if you are concerned about memory?

controllers/cloud.redhat.com/errors/errors.go
adamrdrew: I can be too long winded: TL;DR - maybe we get rid of `MissingDependencies` all together and just pass `MissingDepdency` around because that list only ever contains one entry.

psav: Yeh I can see that, I think we kept it as a list in case we wanted to ever do a concurrent reconcile. - I'm happy to rename it. Now I remember the context, it was actually what was happening inside of the error itself. When we were trying to report kafka topic errors, the topics had to be sorted because otherwise each time we wrote the status, it changed and was different, causing needless writes to the etc database

controllers/cloud.redhat.com/errors/errors.go
psav: I thought the app was already present in the log message by virtue f the fact that it was being emitted from the app itself. If so, I see no need to pass it in here.

controllers/cloud.redhat.com/errors/errors.go
psav: Do as you will with the rename

psav (changes requested):
As discussed

controllers/cloud.redhat.com/errors/errors.go
adamrdrew: Oh! I get what you are saying now. I totally didn't understand. Because the log call is made in the context of an app the app name will be there automatically and our missing dep error is added to it. That makes sense. I thought you were asking me to interpolate the app name into the error string within each provider, which would have been messy. Removed app.

controllers/cloud.redhat.com/providers/kafka/appinterface.go
adamrdrew: See above. I misunderstood what you meant. Removed.

controllers/cloud.redhat.com/errors/errors.go
adamrdrew: I was wrong:

`providers/depdencies/impl.go`
```golang
	missingDeps := makeDepConfig(
		&depConfig,
		&privDepConfig,
		dep.Provider.Env.Spec.Providers.Web.Port,
		dep.Provider.Env.Spec.Providers.Web.PrivatePort,
		app,
		apps,
	)
	if len(missingDeps) > 0 {
		missingDepStructs := []errors.MissingDependency{}
		for _, dep := range missingDeps {
			missingDepStructs = append(missingDepStructs, errors.MissingDependency{
				Source:  "service",
				Details: dep,
			})
		}
		return &errors.MissingDependencies{MissingDeps: missingDepStructs}
	}
```

So there is one place where we use an missing deps with a list of missing deps; and all you need is one to justify leaving the API consistent. So, we should keep with what we have.

So I rescind my response about removing the list feature. So that brings me back to your input. I don't know how you'd want these sorted or in what way. It looks like `providers/depdencies/impl.go` would be where we'd do it. Maybe in `makeDepConfig` - let me know if this is still something you want for this patch and if so how to do it.

psav: I mean - it does now :) - that change is from your PR :)
Actually looking at it - in this case - we are looping over a static list of deps that comes from the clowdapp spec so it's not likely to change order. The problem will be if we bring in a new provider or something that sends back missing deps. If we do that we may want to put in a new object that has an `add` method, that will auto sort it after adding.

psav (approved):
Please see comment - I think this is OK

adamrdrew:
Awesome! Thanks everyone!

PR #688, opened 2022-08-15
Added check for CJI name length.

maknop:


controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -263,11 +263,18 @@ func (r *ClowdJobInvocationReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 	randomString := utils.RandStringLower(7)
+	jobName := fmt.Sprintf("%s-%s", job.Name, randomString)
+
 	nn := types.NamespacedName{
-		Name:      fmt.Sprintf("%s-%s", job.Name, randomString),
 		Namespace: cji.Namespace,
 	}
 
+	if len(jobName) > 63 {
+		r.Recorder.Eventf(cji, "Error", "ClowdJobLabelError", "Job [%s] contains a label with character length greater than 63", jobName)
```
bsquizz, line 273: I am not sure if returning an error here will cause an Event to be emitted. Maybe try to just return an error in a dev environment and see if you get an event produced in the NS?

Whether or not you end up having to call `r.Recorder.Eventf` here though, I think you want to return an err here at this point so that the function does not continue.

bsquizz (changes requested):

controllers/cloud.redhat.com/clowdjobinvocation_controller.go
maknop: I removed `r.Recorder...` and had it return an error instead.

bsquizz:
I know you were testing this out on minikube -- not sure where you are at with things now so let me know when it's ready for another look.

bsquizz (approved):
LGTM, pr check error seems unrelated, will try a retest

PR #690, opened 2022-08-29
Reduce complexity of Clowder methods

adamrdrew:
The goal of this PR is to reduce the complexity of Clowder methods as much as possible during the tech debt sprint. In an effort to go about this in a data driven way I'm using complexity measures based on cyclonic complexity and halstead metrics as provided by the go-complexity-analysis go vet plugin. The goal to to get as many high complexity methods under a cyclonic measure of 10 as possible within the time frame.

### DIFF
--- a/controllers/cloud.redhat.com/clowdapp_controller.go
+++ b/controllers/cloud.redhat.com/clowdapp_controller.go
@@ -16,7 +16,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -33,7 +32,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
@@ -42,7 +40,6 @@ import (
 
 	"github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/clowderconfig"
 	"github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/config"
-	"github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/providers"
 
 	// These imports are to register the providers with the provider registration system
 	_ "github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/providers/autoscaler"
@@ -65,12 +62,9 @@ import (
 	_ "github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/providers/sidecar"
 	_ "github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/providers/web"
 
-	provutils "github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/providers/utils"
-
 	crd "github.com/RedHatInsights/clowder/apis/cloud.redhat.com/v1alpha1"
 	"github.com/RedHatInsights/clowder/controllers/cloud.redhat.com/errors"
 
-	rc "github.com/RedHatInsights/rhc-osdk-utils/resource_cache"
 	"github.com/RedHatInsights/rhc-osdk-utils/utils"
 )
 
@@ -138,192 +132,27 @@ func (r *ClowdAppReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	ctx = context.WithValue(ctx, errors.ClowdKey("recorder"), &r.Recorder)
 	app := crd.ClowdApp{}
 
-	if getAppErr := r.Client.Get(ctx, req.NamespacedName, &app); getAppErr != nil {
-		if k8serr.IsNotFound(getAppErr) {
-			// Must have been deleted
-			return ctrl.Result{}, nil
-		}
-		log.Info("App not found", "env", app.Spec.EnvName, "app", app.GetIdent(), "err", getAppErr)
-		return ctrl.Result{}, getAppErr
-	}
-
-	if _, ok := presentApps[app.GetIdent()]; !ok {
-		presentApps[app.GetIdent()] = true
-	}
-	presentAppsMetric.Set(float64(len(presentApps)))
-
-	delete(managedApps, app.GetIdent())
-
 	defer func() {
 		managedAppsMetric.Set(float64(len(managedApps)))
 	}()
 
-	log = log.WithValues("env", app.Spec.EnvName)
-
-	isAppMarkedForDeletion := app.GetDeletionTimestamp() != nil
-	if isAppMarkedForDeletion {
-		if contains(app.GetFinalizers(), appFinalizer) {
-			if finalizeErr := r.finalizeApp(log, &app); finalizeErr != nil {
-				log.Info("Cloud not finalize", "err", finalizeErr)
-				return ctrl.Result{}, finalizeErr
-			}
-
-			controllerutil.RemoveFinalizer(&app, appFinalizer)
-			removeFinalizeErr := r.Update(ctx, &app)
-			if removeFinalizeErr != nil {
-				log.Info("Cloud not remove finalizer", "err", removeFinalizeErr)
-				return ctrl.Result{}, removeFinalizeErr
-			}
-		}
-		return ctrl.Result{}, nil
-	}
-
-	// Add finalizer for this CR
-	if !contains(app.GetFinalizers(), appFinalizer) {
-		if addFinalizeErr := r.addFinalizer(log, &app); addFinalizeErr != nil {
-			log.Info("Cloud not add finalizer", "err", addFinalizeErr)
-			return ctrl.Result{}, addFinalizeErr
-		}
-	}
-
-	if ReadEnv() == app.Spec.EnvName {
-		r.Recorder.Eventf(&app, "Warning", "ClowdEnvLocked", "Clowder Environment [%s] is locked", app.Spec.EnvName)
-		log.Info("Env currently being reconciled")
-		return ctrl.Result{Requeue: true}, nil
-	}
-
 	log.Info("Reconciliation started")
-	reconciliationMetrics := ReconciliationMetrics{}
-	reconciliationMetrics.init(app.Name, app.Spec.EnvName)
-	reconciliationMetrics.start()
 
-	if clowderconfig.LoadedConfig.Features.PerProviderMetrics {
-		requestMetrics.With(prometheus.Labels{"type": "app", "name": app.GetIdent()}).Inc()
-	}
-
-	if app.Spec.Disabled {
-		log.Info("Reconciliation aborted - set to be disabled")
-		return ctrl.Result{}, nil
-	}
-
-	ctx = context.WithValue(ctx, errors.ClowdKey("obj"), &app)
-
-	env := crd.ClowdEnvironment{}
-
-	if getEnvErr := r.Client.Get(ctx, types.NamespacedName{Name: app.Spec.EnvName}, &env); getEnvErr != nil {
-		log.Info("ClowdEnv missing", "err", getEnvErr)
-		r.Recorder.Eventf(&app, "Warning", "ClowdEnvMissing", "Clowder Environment [%s] is missing", app.Spec.EnvName)
-		if setClowdStatusErr := SetClowdAppConditions(ctx, r.Client, &app, crd.ReconciliationFailed, getEnvErr); setClowdStatusErr != nil {
-			log.Info("Set status error", "err", setClowdStatusErr)
-			return ctrl.Result{Requeue: true}, setClowdStatusErr
+	reconciliation := ClowdAppReconciliation{
+		ctx:      &ctx,
+		client:   r.Client,
+		recorder: r.Recorder,
+		app:      &app,
+		log:      &log,
+		req:      &req}
+	res, err := reconciliation.Reconcile()
+	if err != nil {
+		if shouldSkipReconciliation(err) {
+			return res, nil
 		}
-		return ctrl.Result{}, getEnvErr
+		return res, err
 	}
 
-	ens := &core.Namespace{}
-	if getNSErr := r.Client.Get(ctx, types.NamespacedName{Name: env.Status.TargetNamespace}, ens); getNSErr != nil {
-		return ctrl.Result{Requeue: true}, getNSErr
-	}
-
-	if ens.ObjectMeta.DeletionTimestamp != nil {
-		log.Info("Env target namespace is to be deleted - skipping reconcile")
-		return ctrl.Result{}, nil
-	}
-
-	ans := &core.Namespace{}
-	if getAppNSErr := r.Client.Get(ctx, types.NamespacedName{Name: app.Namespace}, ans); getAppNSErr != nil {
-		return ctrl.Result{Requeue: true}, getAppNSErr
-	}
-
-	if ans.ObjectMeta.DeletionTimestamp != nil {
-		log.Info("App namespace is to be deleted - skipping reconcile")
-		return ctrl.Result{}, nil
-	}
-
-	if env.Generation != env.Status.Generation {
-		r.Recorder.Eventf(&app, "Warning", "ClowdEnvNotReconciled", "Clowder Environment [%s] is not reconciled", app.Spec.EnvName)
-		log.Info("Env not yet reconciled")
-		if setClowdStatusErr := SetClowdAppConditions(ctx, r.Client, &app, crd.ReconciliationFailed, fmt.Errorf("clowd env not reconciled")); setClowdStatusErr != nil {
-			log.Info("Set status error", "err", setClowdStatusErr)
-			return ctrl.Result{Requeue: true}, setClowdStatusErr
-		}
-		return ctrl.Result{}, nil
-	}
-
-	if !env.IsReady() {
-		r.Recorder.Eventf(&app, "Warning", "ClowdEnvNotReady", "Clowder Environment [%s] is not ready", app.Spec.EnvName)
-		log.Info("Env not yet ready")
-		if setClowdStatusErr := SetClowdAppConditions(ctx, r.Client, &app, crd.ReconciliationFailed, fmt.Errorf("clowd env not ready")); setClowdStatusErr != nil {
-			log.Info("Set status error", "err", setClowdStatusErr)
-			return ctrl.Result{Requeue: true}, setClowdStatusErr
-		}
-
-		return ctrl.Result{Requeue: true}, nil
-	}
-
-	cacheConfig := rc.NewCacheConfig(Scheme, errors.ClowdKey("log"), ProtectedGVKs, DebugOptions)
-
-	cache := rc.NewObjectCache(ctx, r.Client, cacheConfig)
-	//cache := providers.NewObjectCache(ctx, r.Client, scheme)
-
-	provider := providers.Provider{
-		Client: r.Client,
-		Ctx:    ctx,
-		Env:    &env,
-		Cache:  &cache,
-		Log:    log,
-	}
-
-	if provErr := r.runProviders(log, &provider, &app); provErr != nil {
-		r.Recorder.Eventf(&app, "Warning", "FailedReconciliation", "Clowdapp requeued [%s]", app.GetClowdName())
-		if setClowdStatusErr := SetClowdAppConditions(ctx, r.Client, &app, crd.ReconciliationFailed, provErr); setClowdStatusErr != nil {
-			log.Info("Set status error", "err", setClowdStatusErr)
-			return ctrl.Result{Requeue: true}, setClowdStatusErr
-		}
-		log.Info("Provider error", "err", provErr)
-		return ctrl.Result{Requeue: true}, provErr
-	}
-
-	cacheErr := cache.ApplyAll()
-
-	if cacheErr != nil {
-		r.Recorder.Eventf(&app, "Warning", "FailedReconciliation", "Clowdapp requeued [%s]", app.GetClowdName())
-		if setClowdStatusErr := SetClowdAppConditions(ctx, r.Client, &app, crd.ReconciliationFailed, cacheErr); setClowdStatusErr != nil {
-			log.Info("Set status error", "err", setClowdStatusErr)
-			return ctrl.Result{Requeue: true}, setClowdStatusErr
-		}
-		log.Info("Cache error", "err", cacheErr)
-		return ctrl.Result{Requeue: true}, cacheErr
-	}
-
-	if statusErr := SetAppResourceStatus(ctx, r.Client, &app); statusErr != nil {
-		log.Info("Set status error", "err", statusErr)
-		return ctrl.Result{Requeue: true}, statusErr
-	}
-
-	opts := []client.ListOption{
-		client.MatchingLabels{app.GetPrimaryLabel(): app.GetClowdName()},
-		client.InNamespace(app.Namespace),
-	}
-
-	// Delete all resources that are not used anymore
-	rErr := cache.Reconcile(app.GetUID(), opts...)
-	if rErr != nil {
-		log.Info("Reconcile error", "err", rErr)
-		return ctrl.Result{Requeue: true}, nil
-	}
-
-	if setClowdStatusErr := SetClowdAppConditions(ctx, r.Client, &app, crd.ReconciliationSuccessful, nil); setClowdStatusErr != nil {
-		log.Info("Set status error", "err", setClowdStatusErr)
-		return ctrl.Result{Requeue: true}, setClowdStatusErr
-	}
-
-	managedApps[app.GetIdent()] = true
-
-	r.Recorder.Eventf(&app, "Normal", "SuccessfulReconciliation", "Clowdapp reconciled [%s]", app.GetClowdName())
-	log.Info("Reconciliation successful")
-
-	reconciliationMetrics.stop()
 	return ctrl.Result{}, nil
 }
 
@@ -401,31 +230,6 @@ func (r *ClowdAppReconciler) appsToEnqueueUponEnvUpdate(a client.Object) []recon
 	return reqs
 }
 
-func (r *ClowdAppReconciler) finalizeApp(reqLogger logr.Logger, a *crd.ClowdApp) error {
-
-	delete(managedApps, a.GetIdent())
-	managedAppsMetric.Set(float64(len(managedApps)))
-
-	delete(presentApps, a.GetIdent())
-	presentAppsMetric.Set(float64(len(presentApps)))
-
-	reqLogger.Info("Successfully finalized ClowdApp")
-	return nil
-}
-
-func (r *ClowdAppReconciler) addFinalizer(reqLogger logr.Logger, a *crd.ClowdApp) error {
-	reqLogger.Info("Adding Finalizer for the ClowdApp")
-	controllerutil.AddFinalizer(a, appFinalizer)
-
-	// Update CR
-	err := r.Update(context.TODO(), a)
-	if err != nil {
-		reqLogger.Error(err, "Failed to update ClowdApp with finalizer")
-		return err
-	}
-	return nil
-}
-
 func contains(list []string, s string) bool {
 	for _, v := range list {
 		if v == s {
@@ -449,35 +253,4 @@ func updateMetadata(app *crd.ClowdApp, appConfig *config.AppConfig) {
 	appConfig.Metadata = &metadata
 	appConfig.Metadata.Name = &app.Name
 	appConfig.Metadata.EnvName = &app.Spec.EnvName
-}
-
-func (r *ClowdAppReconciler) runProviders(log logr.Logger, provider *providers.Provider, a *crd.ClowdApp) error {
-
-	c := config.AppConfig{}
-
-	// Update app metadata
-	updateMetadata(a, &c)
-
-	for _, provAcc := range providers.ProvidersRegistration.Registry {
-		provutils.DebugLog(log, "running provider:", "name", provAcc.Name, "order", provAcc.Order)
-		start := time.Now()
-		prov, err := provAcc.SetupProvider(provider)
-		elapsed := time.Since(start).Seconds()
-		providerMetrics.With(prometheus.Labels{"provider": provAcc.Name, "source": "clowdenv"}).Observe(elapsed)
-		if err != nil {
-			return errors.Wrap(fmt.Sprintf("getprov: %s", provAcc.Name), err)
-		}
-		start = time.Now()
-		err = prov.Provide(a, &c)
-		elapsed = time.Since(start).Seconds()
-		providerMetrics.With(prometheus.Labels{"provider": provAcc.Name, "source": "clowdapp"}).Observe(elapsed)
-		if err != nil {
-			reterr := errors.Wrap(fmt.Sprintf("runapp: %s", provAcc.Name), err)
-			reterr.Requeue = true
-			return reterr
-		}
-		provutils.DebugLog(log, "running provider: complete", "name", provAcc.Name, "order", provAcc.Order, "elapsed", fmt.Sprintf("%f", elapsed))
-	}
-
-	return nil
 }
--- a/controllers/cloud.redhat.com/clowdapp_reconciliation.go
+++ b/controllers/cloud.redhat.com/clowdapp_reconciliation.go
@@ -172,7 +172,7 @@ func (r *ClowdAppReconciliation) isEnvLocked() (ctrl.Result, error) {
 	if ReadEnv() == r.app.Spec.EnvName {
 		r.recorder.Eventf(r.app, "Warning", "ClowdEnvLocked", "Clowder Environment [%s] is locked", r.app.Spec.EnvName)
 		r.log.Info("Env currently being reconciled")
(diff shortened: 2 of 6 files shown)


adamrdrew:
Gave up on keda provider method `getTriggerRoute` - I tried to refactor it multiple times and there's something wrong with my approach. I don't see the problem no matter how much I look and I can't reproduce the failures locally. Decided to just give up on it becuase the method has a high complexity score, but it is artificial - there's a huge empty switch block in it.

adamrdrew:
OK, so its passing tests now. I'll do some tidying up and then get it ready for review.

controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+func (r *ClowdAppReconciliation) steps() []func() (ctrl.Result, error) {
+	return []func() (ctrl.Result, error){
+		r.getApp,
+		r.setPresentAndManagedApps,
+		r.startMetrics,
+		r.isAppMarkedForDeletion,
+		r.addFinalizer,
+		r.isEnvLocked,
```
psav, line 63: I wonder if we could/should rename some of these a little and combine them. Right now there are a lot of steps to visually go through, for example, what about
`getClowdEnv` also chain calling, `isEnvCoreNamespaceDeleted`, `isClowdEnvReady`, and `isClowdEnvReconciled`

Perhaps it needs a different name, but I think that would collapse things down a little in this list of steps.

It could also be argued that the same could be done for `getApp`.

controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -0,0 +1,381 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *ClowdAppReconciliation) stopMetrics() (ctrl.Result, error) {
+	r.reconciliationMetrics.stop()
+	return ctrl.Result{}, nil
+}
+
+func (r *ClowdAppReconciliation) setPresentAndManagedApps() (ctrl.Result, error) {
+	if _, ok := presentApps[r.app.GetIdent()]; !ok {
```
psav, line 106: In looking at this - I think we can simplify this - I know it's copy pasta, but why not

``presentApps[r.app.GetIdent()] = true``

can replace the if block

controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -0,0 +1,381 @@
...
+		presentApps[r.app.GetIdent()] = true
+	}
+	presentAppsMetric.Set(float64(len(presentApps)))
+
+	delete(managedApps, r.app.GetIdent())
```
psav, line 111: Can you add a comment here - it took me a moment to grok what I did - we remove it from the managed list because it _may_ have been managed before, but it may not be after this reconcile.

controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -0,0 +1,381 @@
...
+		if k8serr.IsNotFound(getAppErr) {
+			// Must have been deleted
+			return ctrl.Result{}, errors.New(SKIPRECONCILE)
+		}
+		r.log.Info("App not found", "env", r.app.Spec.EnvName, "app", r.app.GetIdent(), "err", getAppErr)
+		return ctrl.Result{}, getAppErr
+	}
+	//This is kinda side-effecty but I couldn't think of a better place
+	//to put it.
+	logWithEnv := r.log.WithValues("env", r.app.Spec.EnvName)
```
psav, line 127: This is a side effect of the steps. Piece, one could argue it could become its own step. It also doesn't matter when it happens, so we could move this little piece out and into the actual reconciliation function as it references the EnvName from the spec and not from a real ClowdEnv, which is only gotten half way through the steps.

controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -0,0 +1,381 @@
...
+	r.log = &logWithEnv
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ClowdAppReconciliation) isAppMarkedForDeletion() (ctrl.Result, error) {
+	isAppMarkedForDeletion := r.app.GetDeletionTimestamp() != nil
```
psav, line 134: As soon as we get this in, I want to migrate this out and into the rhc-osdk-utils - No action, just an FYI

controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -0,0 +1,381 @@
...
+			r.log.Info("Set status error", "err", setClowdStatusErr)
+			return ctrl.Result{Requeue: true}, setClowdStatusErr
+		}
+		return ctrl.Result{}, getEnvErr
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ClowdAppReconciliation) isEnvCoreNamespaceDeleted() (ctrl.Result, error) {
```
psav, line 223: Do we need `Core` in here?

controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	if ens.ObjectMeta.DeletionTimestamp != nil {
+		r.log.Info("Env target namespace is to be deleted - skipping reconcile")
+		return ctrl.Result{}, nil
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *ClowdAppReconciliation) isAppCoreNamespaceDeleted() (ctrl.Result, error) {
```
psav, line 236: Same as above

controllers/cloud.redhat.com/clowdapp_reconciliation.go
psav, line 223: These two functions are identical apart from the name, should we create a combined function for this?

controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -0,0 +1,381 @@
...
+
+func (r *ClowdAppReconciliation) isEnvReady() (ctrl.Result, error) {
+	if !r.env.IsReady() {
+		r.recorder.Eventf(r.app, "Warning", "ClowdEnvNotReady", "Clowder Environment [%s] is not ready", r.app.Spec.EnvName)
+		r.log.Info("Env not yet ready")
+		if setClowdStatusErr := SetClowdAppConditions(*r.ctx, r.client, r.app, crd.ReconciliationFailed, fmt.Errorf("clowd env not ready")); setClowdStatusErr != nil {
+			r.log.Info("Set status error", "err", setClowdStatusErr)
+			return ctrl.Result{Requeue: true}, setClowdStatusErr
+		}
+
```
psav, line 271: blank line remove

controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -107,237 +107,70 @@ func ReadEnv() string {
...
-//Reconcile fn
-func (r *ClowdEnvironmentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+//Acts as a guard for a reconciliation cycle, as well as initatizes a bunch of required objects
+func (r *ClowdEnvironmentReconciler) initReconciliation(ctx context.Context, req ctrl.Request) (logr.Logger, context.Context, crd.ClowdEnvironment, ctrl.Result, error) {
 	log := r.Log.WithValues("env", req.Name).WithValues("rid", utils.RandString(5))
 	ctx = context.WithValue(ctx, errors.ClowdKey("log"), &log)
 	ctx = context.WithValue(ctx, errors.ClowdKey("recorder"), &r.Recorder)
 	env := crd.ClowdEnvironment{}
 
+	silentFailError := errors.New(SILENTFAIL)
```
psav, line 117: Why pass this back at all? Why not just pass back nil? Then there is no need to do the conditional later in the `Reconcile` function

controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -107,237 +107,70 @@ func ReadEnv() string {
...
 
 	delete(managedEnvironments, env.Name)
+}
 
-	defer func() {
-		managedEnvsMetric.Set(float64(len(managedEnvironments)))
-	}()
+//Reconcile fn
+func (r *ClowdEnvironmentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log, ctx, env, res, err := r.initReconciliation(ctx, req)
```
psav, line 142: A few things around this
1) I'm not sure I understand why we are doing it differently to the clowdapp
2) Passing in the context and getting a modified version back seems rather strange to me
I propose making it like ClowdApp

controllers/cloud.redhat.com/clowdenvironment_controller.go
```diff
@@ -107,237 +107,70 @@ func ReadEnv() string {
...
-		log.Info("GetEnvResourceStatus error", "err", getEnvResErr)
-		return ctrl.Result{Requeue: true}, getEnvResErr
-	}
-
-	envStatus := core.ConditionFalse
-	successCond := cond.Get(&env, crd.ReconciliationSuccessful)
-	if successCond != nil {
-		envStatus = successCond.Status
-	}
+	r.initMetrics(env)
```
psav, line 154: Is there a reason that metrics is handled differently here?

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
```diff
@@ -0,0 +1,489 @@
...
+
+//During reconciliation we handle errors in 2 ways: sometimes we want to error out of reconciliation and sometimes we want to skip reconciliation.
+func shouldSkipReconciliation(err error) bool {
+	if err != nil && err.Error() == SKIPRECONCILE {
+		return true
+	}
+	return false
+}
+
+func MakeClowdEnvReconciliation(cache *rc.ObjectCache,
```
psav, line 36: Same question as for ClowdApp

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
```diff
@@ -0,0 +1,489 @@
...
+	cache    *rc.ObjectCache
+	recorder record.EventRecorder
+	ctx      *context.Context
+	client   client.Client
+	env      *crd.ClowdEnvironment
+	log      *logr.Logger
+}
+
+//Returns a list of step methods that should be run during reconciliation
+func (r *ClowdEnvironmentReconciliation) steps() []func() (ctrl.Result, error) {
```
psav, line 66: In thinking about the steps piece here - the only thing I'm concerned about is that if we need to inject something in the middle it's a whole new step - I guess that's OK, yeh I guess that's fine - sorry stream of consciousness

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
```diff
@@ -0,0 +1,489 @@
...
+		r.deleteUnusedResources,
+		r.setClowdEnvConditions,
+		r.logSuccess,
+	}
+}
+
+//Public method to iterate through the steps of the reconciliation process
+func (r *ClowdEnvironmentReconciliation) Reconcile() (ctrl.Result, error) {
+	r.log.Info("Reconciliation started")
+	//The env stays locked for the entire reconciliation
```
psav, line 90: Not a problem with this change

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
```diff
@@ -0,0 +1,489 @@
...
+		result, err := step()
+		if err != nil {
+			return result, err
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+//Determine if env is marked for deletion, and if so finalize and end resonciliation
+func (r *ClowdEnvironmentReconciliation) markedForDeletion() (ctrl.Result, error) {
```
psav, line 105: Comment as in app

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
```diff
@@ -0,0 +1,489 @@
...
+func (r *ClowdEnvironmentReconciliation) initTargetNamespace() (ctrl.Result, error) {
+	result := ctrl.Result{}
+	var err error = nil
+
+	if r.env.Status.TargetNamespace != "" {
+		return result, err
+	}
+
+	if r.env.Spec.TargetNamespace == "" {
+		result, err = r._makeTargetNamespace()
```
psav, line 199: Why the underscores?

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
```diff
@@ -0,0 +1,489 @@
...
+	if ens.ObjectMeta.DeletionTimestamp != nil {
+		r.log.Info("Env target namespace is to be deleted - skipping reconcile")
+		return ctrl.Result{}, errors.New(SKIPRECONCILE)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ClowdEnvironmentReconciliation) runProviders() (ctrl.Result, error) {
+	provider := r._makeProvider()
```
psav, line 279: Is this just to reduce line count of the function? I'd perhaps argue in a case like this, that being able to see the literal would help understanding.

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
```diff
@@ -0,0 +1,489 @@
...
+		}
+		return ctrl.Result{Requeue: true}, setAppErr
+	}
+	return ctrl.Result{}, nil
+}
+
+//Performs the actual business logic for the setAppInfo step
+//TODO this is a bear of a method. Factor this out into pieces.
+func (r *ClowdEnvironmentReconciliation) _setAppInfoImplementation() error {
+	p := r._makeProvider()
```
psav, line 335: You don't actually need a provider object now, you already have everything you need on `r` :)

psav (changes requested):
First pass review - love love love :heart: the work in here, if has made things ~im~measurably better, a few questions though and a few things that I think are bugs

controllers/cloud.redhat.com/providers/deployment/impl.go
```diff
@@ -152,45 +138,95 @@ func initDeployment(app *crd.ClowdApp, env *crd.ClowdEnvironment, d *apps.Deploy
 		if readinessProbe.FailureThreshold == 0 {
 			readinessProbe.FailureThreshold = 3
 		}
-	} else if bool(deployment.Web) || deployment.WebServices.Public.Enabled {
-		readinessProbe = baseProbe
+
+		c.ReadinessProbe = &readinessProbe
+	}
+	if bool(deployment.Web) || deployment.WebServices.Public.Enabled {
```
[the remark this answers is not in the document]
adamrdrew: Good call! Changed

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -138,192 +133,24 @@ func (r *ClowdAppReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
-
-	cacheErr := cache.ApplyAll()
+	log.Info("Reconciliation started")
 
-	if cacheErr != nil {
-		r.Recorder.Eventf(&app, "Warning", "FailedReconciliation", "Clowdapp requeued [%s]", app.GetClowdName())
-		if setClowdStatusErr := SetClowdAppConditions(ctx, r.Client, &app, crd.ReconciliationFailed, cacheErr); setClowdStatusErr != nil {
-			log.Info("Set status error", "err", setClowdStatusErr)
-			return ctrl.Result{Requeue: true}, setClowdStatusErr
+	reconciliation := MakeClowdAppReconciliation(ctx, &cache, r.Client, r.Recorder, &app, log, &req)
```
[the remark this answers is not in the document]
adamrdrew: Yeah I tried to avoid judgement calls where possible but this was one of them so I figured I'd try it and see what happens.

I don't have any kind of mental model for resource usage and performance impacts re: Clowder running in real life. Other than "try to use as few resources as possible" I don't have any sort of heuristics around that in my head about like what is and is not expensive relatively speaking. So in moving this code I didn't know if I was doing something as cheap as allocating an empty struct or as expensive as spinning up some complex system. Considering the comment I suspect its more like the latter than the former.

I'll change this to be a step method in `ClowdAppReconciliation` and I'll do it as late as possible in the process.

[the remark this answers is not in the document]
adamrdrew: Moved it to a step!

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -138,192 +133,24 @@ func (r *ClowdAppReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+	res, err := reconciliation.Reconcile()
+	if err != nil {
+		if shouldSkipReconciliation(err) {
+			return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
adamrdrew: Ah ok, it should be 

```
	res, err := reconciliation.Reconcile()
	if err != nil {
		if shouldSkipReconciliation(err) {
			return res, nil
		}
		return res, err
	}
```

Because some skips requeue:
```go
func (r *ClowdAppReconciliation) isEnvLocked() (ctrl.Result, error) {
	if ReadEnv() == r.app.Spec.EnvName {
		r.recorder.Eventf(r.app, "Warning", "ClowdEnvLocked", "Clowder Environment [%s] is locked", r.app.Spec.EnvName)
		r.log.Info("Env currently being reconciled")
		return ctrl.Result{Requeue: true}, errors.New(SKIPRECONCILE)
	}
	return ctrl.Result{}, nil
}
```

And some don't:
```go
func (r *ClowdAppReconciliation) isAppDisabled() (ctrl.Result, error) {
	if r.app.Spec.Disabled {
		r.log.Info("Reconciliation aborted - set to be disabled")
		return ctrl.Result{}, errors.New(SKIPRECONCILE)
	}
	return ctrl.Result{}, nil
}
```

Which matches the behavior of the old code:
```go

	if ReadEnv() == app.Spec.EnvName {
		r.Recorder.Eventf(&app, "Warning", "ClowdEnvLocked", "Clowder Environment [%s] is locked", app.Spec.EnvName)
		log.Info("Env currently being reconciled")
		return ctrl.Result{Requeue: true}, nil
	}
...
	if app.Spec.Disabled {
		log.Info("Reconciliation aborted - set to be disabled")
		return ctrl.Result{}, nil
	}
```

I changed it to reture the result - the steps were already doing the right thing but the skip eval wasn't respecting requeues

controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -0,0 +1,381 @@
...
+	ctx                   *context.Context
+	client                client.Client
+	log                   *logr.Logger
+	app                   *crd.ClowdApp
+	req                   *ctrl.Request
+	env                   *crd.ClowdEnvironment
+	reconciliationMetrics ReconciliationMetrics
+}
+
+func MakeClowdAppReconciliation(ctx context.Context,
```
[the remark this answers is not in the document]
adamrdrew: Good catch re: public method. It is a hold over from an earlier design where this code was in its own package. Changed to private.

Its a method because you slapped my hand once for doing a literal declaration instead of using a constructor function :) There was probably a more subtle point to your asking me to use a function in the past, maybe like init logic or something, but it stuck in my head as "Pete wants me to use methods when creating instances of complex types. I can change it.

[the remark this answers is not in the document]
adamrdrew: Removed

controllers/cloud.redhat.com/clowdapp_reconciliation.go
adamrdrew: I would like to push back on this suggestion. My goal was for each step to (as much as possible) do a single thing and to have a complexity score of 1 or 2 - 3 max. This makes each step easy to reason about, and will make it very easy to write tests for each step. If the steps do more then there will be high conditionality, more ways to return, and will make testing more complex. (Note: I know tests aren't included in the patch. I ran out of time. But in future I'd like to add tests for every step.)

I do think this is the best way to bake this particular cake.

controllers/cloud.redhat.com/clowdapp_reconciliation.go
adamrdrew: Good call. I had noticed that during the refactor and was like "Oh god, what am I not understanding about this block. It must be like this for a reason." I ended up leaving it out of fear lol. Changed.

controllers/cloud.redhat.com/clowdapp_reconciliation.go
adamrdrew: Done

controllers/cloud.redhat.com/clowdapp_reconciliation.go
adamrdrew: Yeah this is a tough little one. Where I wanted to put this was at the very top of `ClowdAppReconciler.Reconcile` where we do all of those `r.Log.WithValues` calls, but I couldn't figure out how to get access to the env name before getting the app.  If there is a way to get the env name before getting the app then that would be a good change I think, but I don't know how.

So then the most strictly idiomatic thing would be to put it in a step, right? Yeah, it would be, except that this log update has no conditionality associated with it. It can't (so far as I can tell) fail. Additionally, there's no implementation to test, as this is a single public method on the logging library and thus outside of scope for testing. So, in this particular case a step would buy us nothing: no improvement to testing and  no reduction in complexity.

So long story short (too late) if you can tell me a way to grab env name before loading the app I'll move it out of the step. If not then I suggest we leave it as is.

controllers/cloud.redhat.com/clowdapp_reconciliation.go
adamrdrew: Good call :+1:

controllers/cloud.redhat.com/clowdapp_reconciliation.go
adamrdrew: That was some cargo cult naming on my part. I didn't know if the fact that we were dealing with a core namespace was in some way significant. Judging from context I'm assuming it isn't :) I'll remove it..

adamrdrew: Oh yeah good call

adamrdrew: Consolidated them

controllers/cloud.redhat.com/clowdenvironment_controller.go
adamrdrew: Yeah that's a little complex. I'll explain why it is that way and then let me know if you see a better way to handle it.

The core problem here is that the Reconcile function takes 3 different actions depending on the result of attempting to get the env from the client:
- If there's no error move on with the reconcile
- If the error is that the environment was deleted return an empty result and nil, terminating the reconcile but not returning an error
- If the error is any other error then terminate the reconcile and return the error

So the case of the env having been deleted is a special case where we silently fail. That's how it worked in the original code:

```go
if getEnvErr := r.Client.Get(ctx, req.NamespacedName, &env); getEnvErr != nil {
		if k8serr.IsNotFound(getEnvErr) {
			// Must have been deleted
			return ctrl.Result{}, nil
		}
		log.Info("Namespace not found", "err", getEnvErr)
		return ctrl.Result{}, getEnvErr
	}
```

So when I factored this out into a method I had to provide some way to tell the Reoncile function "Hey this is the special case where you want to bail based on an error but where you don't want to return the error. I didn't see any other way to handle this, and looking at the code I still don't (except to forgo using the special SILENTFAIL error I created and just move the `k8serr.IsNotFound` check into the error handler in Reconcile.

So open to suggestions on this one for sure.

controllers/cloud.redhat.com/clowdenvironment_controller.go
adamrdrew: Done

controllers/cloud.redhat.com/clowdenvironment_controller.go
adamrdrew: To try and match the original implementation as closely as possible. Metrics are initialized in the clowd environment far earlier in the Reconcile method than clowd app, before the point where the step system takes over. In clowd app the metrics are initialized in the middle of the process. I'm happy to put it wherever you suggest but this was me trying to preserve as much of the original logic flow as I could.

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
adamrdrew: Removed

controllers/cloud.redhat.com/clowdapp_reconciliation.go
adamrdrew: removed

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
adamrdrew: Yeah I would actually say that's a huge benefit. The reconcile function before was very hard to understand for someone new (or even not so knew to the project) - really had to take it in like a novel and understand it cover to cover. You had to read it and say "OK we're doing this here, and doing this here" etc. Now you can just read the step list and its like "Oh, ok, I get that." And if you want to add a step you know exactly where you are in the process, what's come before and will happen after, etc. Its maybe my favorite piece of this refactor. Seeing reconciliation as this list of steps I think is a bonus.

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
adamrdrew: Awesome! I was super nervous about this.

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
adamrdrew: Yeah, I wasn't sure how you'd react to that. I can change it if you don't like it and I'm not married to it. Basically, I wanted some way to denote what is and is not a step method. I could preface every step method with `step` but that would get ugly fast, and considering steps are 90% of the methods that didn't seem sensible. Scope can't help us because all these methods are private. If each step was its own type then each step's Run method would be public and any helper methods would be private and limited to that type. So, the best worst solution I landed on was to use an underscore to hint that this method is a private implementation detail and not a step.

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
adamrdrew: No it was just being DRY. We create the provider in two places and if I notice something being done in 2 places I'll factor it out. None of the logic in the code deals with the internal state of the provider, we just pass it out somewhere so I figured this made sense.

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
adamrdrew: I would request that we do that later because this method is a straight copy and paste from the original code and I don't understand what it is doing. Considering the size and complexity of the method I'd prefer to leave it as is (which is exactly what was in clowder before) and then we tackle it later.

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
adamrdrew: :+1:

controllers/cloud.redhat.com/providers/deployment/impl.go
```diff
@@ -119,6 +93,11 @@ func initDeployment(app *crd.ClowdApp, env *crd.ClowdEnvironment, d *apps.Deploy
 		SuccessThreshold:    1,
 		TimeoutSeconds:      1,
 	}
+}
+
+func setLivenessProbe(pod crd.PodSpec, deployment crd.Deployment, env *crd.ClowdEnvironment, c *core.Container) {
```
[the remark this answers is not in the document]
psav: Discussed - resolving as we need to ensure Clowder politely corrects anything the user may have changed/altered.

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
adamrdrew: Renamed the underscore methods to funcNameImplementation in cases where funcName already existed as a step. Otherwise just gave them sane names.

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
adamrdrew: Refactored. You were right, way easier than I expected. Just deleted the provider and let the red squiggles tell me what to fix.

controllers/cloud.redhat.com/clowdenvironment_reconciliation.go
adamrdrew: Discussed. Removed function

controllers/cloud.redhat.com/clowdenvironment_controller.go
adamrdrew: Discussed. Leaving it as is.

controllers/cloud.redhat.com/clowdapp_controller.go
```diff
@@ -138,192 +132,27 @@ func (r *ClowdAppReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+	reconciliation := ClowdAppReconciliation{
+		ctx:      &ctx,
+		client:   r.Client,
+		recorder: r.Recorder,
+		app:      &app,
+		log:      &log,
+		req:      &req}
+	res, err := reconciliation.Reconcile()
+	if err != nil {
+		if shouldSkipReconciliation(err) {
```
psav, line 150: Sorry @adamrdrew I think there are some issues with this still. The main problem as far as I see is how the requeue is to be handled. In k8s, if you return a `nil` error from a reconciliation, the requeue request is ignored. It actually won't reconcile again ever, (until another requests comes in to reconcile). So this code returning `res` along with `nil` is actually equivalent to passing `ctrl.Result{}, nil`.

controllers/cloud.redhat.com/clowdapp_reconciliation.go
psav: I don't see a comment :)

psav (changes requested):
Tiny changes here I think

controllers/cloud.redhat.com/clowdapp_controller.go
adamrdrew: Hmmm, is this a problem with the new implementation or what the problem always there?

This is the logic in the original code that I based this on:
```golang
	if ens.ObjectMeta.DeletionTimestamp != nil {
		log.Info("Env target namespace is to be deleted - skipping reconcile")
		return ctrl.Result{}, nil
	}
```

In our new step code that is recast as:
```golang
	if ens.ObjectMeta.DeletionTimestamp != nil {
		r.log.Info("Env target namespace is to be deleted - skipping reconcile")
		return ctrl.Result{}, errors.New(SKIPRECONCILE)
	}
```

Which when it is passed through the code you highlighted will result in returning `ctrl.Result{}, nil` which should be exactly the same thing we did in the original code.

The other place in the steps we do this is:
```golang
func (r *ClowdEnvironmentReconciliation) setToBeDisabled() (ctrl.Result, error) {
	if r.env.Spec.Disabled {
		r.log.Info("Reconciliation aborted - set to be disabled")
		return ctrl.Result{}, errors.New(SKIPRECONCILE)
	}
	return ctrl.Result{}, nil
}
```

Which was based on and should be logically equivilent to:
```golang
	if app.Spec.Disabled {
		log.Info("Reconciliation aborted - set to be disabled")
		return ctrl.Result{}, nil
	}
```

So, from the way I parse all this we should be doing exactly what the old implementation was doing. Am I parsing this out wrong? If not, then is this a problem that already existed in Clowder? And if so, what so I do in these cases?

psav:
```
func (r *ClowdAppReconciliation) isEnvLocked() (ctrl.Result, error) {
	if ReadEnv() == r.app.Spec.EnvName {
		r.recorder.Eventf(r.app, "Warning", "ClowdEnvLocked", "Clowder Environment [%s] is locked", r.app.Spec.EnvName)
		r.log.Info("Env currently being reconciled")
		return ctrl.Result{Requeue: true}, errors.New(SKIPRECONCILE)
	}
	return ctrl.Result{}, nil
}
```

I think this function is the one that breaks. It says the Clowder Env is Locked, and it should be requeued. But it we squash the error and return `nil`, this won't be the case.

adamrdrew: OK, I changed it but that's what we did before this change:

```golang
	if ReadEnv() == app.Spec.EnvName {
		r.Recorder.Eventf(&app, "Warning", "ClowdEnvLocked", "Clowder Environment [%s] is locked", app.Spec.EnvName)
		log.Info("Env currently being reconciled")
		return ctrl.Result{Requeue: true}, nil
	}
}
```

adamrdrew: We're failing on merge conflict again. There were more changes to the deployment implementation so we're dying on that. I tried to resolve it but I'm not sure what changed - the diff output is a mess again due to how big the delta is in this branch.

psav: I may have a large apology to make

psav: I think I had forgotten the state we had where this was wrong
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

adamrdrew: Cool cool - flipped it back around

PR #696, opened 2022-09-03
Scale Deployments to 0 when minReplicas is set to 0.

lindgrenj6:
It became apparent during a DB upgrade that when we update `minReplicas` to 0 on our clowdapps that the deployment resources wouldn't be scaled to 0. I think this is due to the check that the deployment replicas is _less than_ the configured replicas - e.g. it scales up the deployments but refuses to scale them down.

This PR basically changes that check to do a few things:
1. If the deployment is nil -> set it to the configured value for initialization
2. Otherwise, (if autoscaling is disabled) sync the deployment + podSpec replica count

Let me know if this isn't the fix you're looking for, I was just playing around here to see if it was in this little check. cc @psav


psav:
This is gonna be addressed in another PR already open #686.

lindgrenj6:
Oh sweet! That's great.

PR #700, opened 2022-09-09
RHCLOUD-21170: Add cyndi ConfigMap when kafka provider is managed-ephem

bsquizz:
We need to tweak the connector topic, "deadletterqueue" topic, and topic replication factor for cyndi to work properly in managed-ephem kafka environments.

Related to: https://github.com/RedHatInsights/cyndi-operator/pull/72


controllers/cloud.redhat.com/providers/kafka/managed-ephem.go
```diff
@@ -326,16 +326,29 @@ func (mep *managedEphemProvider) configureBrokers() error {
...
 
-func (mep *managedEphemProvider) createCyndiPipeline(app *crd.ClowdApp) error {
+func (mep *managedEphemProvider) configCyndi(app *crd.ClowdApp) error {
 	if !app.Spec.Cyndi.Enabled {
 		return nil
 	}
-	return createCyndiPipeline(mep, app, getConnectNamespace(mep.Env), getConnectClusterName(mep.Env))
+
+	// TODO: as a followup, make these topic names configurable
+	eventsTopic := fmt.Sprintf("%s-platform.inventory.events", mep.Env.Name)
```
psav, line 335: Honestly, I'd move these two topic name strings into the other function, then the prototype becomes a lot simpler.

psav (changes requested):

psav:
Hey - a little confused in this one, I know the change to cyndi will affect both types of operator. The test seems to be called `managed-kafka` but the mode is set to `operator`. If it is an operator mode test, it shouldn't be called `managed kafka` and if it's managed kafka, it shouldn't have mode of `operator`.

bsquizz:
Yep the confusion is warranted -- the test should be using 'managed-ephem' mode, will change it.

controllers/cloud.redhat.com/providers/kafka/managed-ephem.go
bsquizz: Moved

bsquizz:
Gave up on writing a kuttl test -- the dependency on contacting managed kafka is too ingrained so at the moment, too much communication w/ managed kafka occurs before we get to the cyndi codepath.

psav (approved):

PR #705, opened 2022-09-23
Improve clowder performance refactor

psav:


adamrdrew (approved):
"64 files, 1282/1195 changes and no comments Adam? Really?" I didn't check hit "File Viewed" - I read every line. I just didn't see any problems. That's because if there are problems they are hiding in the implementation details that go over my head without pulling the code down and running it through a debugger. Most of the change by line count is simple stuff that all looked right: adding the EnvProvide to every provider, changing how config and context are accessed - and all that looks great. If there were errors there I would have caught them. But the deeper stuff I don't see anything obviously wrong - but I can't say I fully understand all of the ramifications of the logic. That said we're passing tests, and you know this code worlds better than me, so I say :+1: :shipit:

PR #712, opened 2022-10-18
[RHCLOUD-21108] - Add parallelism and completions to job specs

SteveHNH:
- feat: Add completions and parallelism to clowdjob
- feat: add completions parallelsim to job spec

This feature adds the ability to set parallelism and completion to jobs and cronjobs
using clowder.

RHCLOUD-21108


adamrdrew (commented):
All looks great to me! I've got no feedback except thanks for the feature!

PR #713, opened 2022-10-21
Set secretData on managed ephem provider after MK secret is fetched

bsquizz:
It doesn't appear that we are storing the data of the ephemeral managed kafka secret on the managed ephemeral provider


maknop (approved):

adamrdrew (approved):
This looks like we might have missed it. I don't see anywhere we store this on `mep`.

PR #714, opened 2022-10-21
Set mep.secretData on envProvide instead of app Provide

bsquizz:
We need to set the secretData in the env provider since we do not yet have a ClowdApp present at the point this code runs


maknop (approved):

adamrdrew (approved):
This could have been an oversight in the split to 2 different reconcile paths. I'm not sure this will fix our issue but it looks likle it could. At worst its a noop.

PR #732, opened 2022-12-12
Service annotations

SteveHNH:
I took a stab at putting annotations on services but can't seem to get the tests to pass. I followed the pattern that podSpec uses for doing the same thing but I may have missed something that I'm just not seeing. Throwing this out there to see if I can get some extra eyes.


SteveHNH:
Looks like this is working as expected now. Ready for review

psav:
Can I ask what's the use for this?

SteveHNH:
@psav Sure! I'm interested in using service-serving-certs which is an openshift feature that requires you to setup an annotation on a service. With the annotation in place, openshift will create certs for you that you can use for TLS encryption between pods. 

It's one of our options in FedRamp outside of getting service mesh going. 

I'm certainly open to different methods if there is a better way.

Here's the docs on it: https://docs.openshift.com/container-platform/4.11/security/certificates/service-serving-certificate.html

psav:
I know of it - I asked the question because I feel perhaps we should let Clowder manage this on the service, rather than allowing people to place arbitrary annotations on the service. Thoughts?

SteveHNH:
@psav That would be fine by me. Would that just be some keyword in the clowder config that would provide the proper annotation? I can rework the PR if that sounds like a viable thing.

something like:

```
webServices:
  tls: true
```

or break it down between private or public

```
webServices:
  public:
    enabled: true
    tls: true
  private:
    enabled: true
```

SteveHNH:
Closing. Replaced by #745

PR #745, opened 2023-01-03
feat: Add TLS annotation for intra-pod encryption

SteveHNH:
Adding a feature into clowder that allows a service to specify that it wants TLS encryption. This will add an annotation to the service that will create a secret containing a certificate that can be used to encrypt the traffic. [Documented Here](https://docs.openshift.com/container-platform/4.11/security/certificates/service-serving-certificate.html)

The secret will need to be accessed by the pod and then used by the app, but this will make it available. 

Example:
```yaml
webServices:
  tls: True
  public:
    enabled: True
  private:
    enabled: True
```

EDIT: moved to clowdenv spec

```yaml
spec:
  serviceConfig:
    tls: True
    type: ClusterIP
```


SteveHNH:
@kylape that's a good point. Thanks for the feedback

SteveHNH:
@kylape @psav First time messing with clowdenv updates so let me know if I missed anything. The tests pass locally which seems like a good start.

controllers/cloud.redhat.com/version.txt
```diff
@@ -1 +1 @@
-master
\ No newline at end of file
+v0.46.0-149-gea2bc060
```
kylape, line 1: it feels like this shouldn't be updated, but I'm not 100% sure.

psav: This accurate

psav (commented):
Putting it on the ENV means ALL apps would have to be TLS enabled, also, I'm pretty sure that we also need to inject the cert into a config map and mount it as part of the deployment?

kylape:
After thinking about this some more, we should probably add in some automation to make the certificate files available in the container for apps.  I think the best way would be to mount the certificate secret into the container and add a pointer to the cert dir in `cdappconfig.json`. wdyt?

adamrdrew:
FYI `Deployment template not updated. Please run make build-template and recommit`

Also, with TLS being required will this break anyone?

kylape:
> will this break anyone?

afaict there is no enforcement here, just making the certs available.  hmm but now that I'm thinking about it, should we also add a TLS port to the service?  and if we do that, should we _remove_ the non-TLS port? :thinking:

kylape:
Removing the non-TLS port is where we'd run into enforcement issues, so perhaps we don't do that yet?

controllers/cloud.redhat.com/version.txt
SteveHNH: I think you're right. It gets updated when you do a clowder install so I probably just shouldn't have committed it.

SteveHNH:
All this is going to do is setup an annotation that tells openshift to make a secret available for tls, but it doesn't enable anything. That requires changes within the app itself.

It probably would be a good idea to get it into cdappconfig.json so apps could use it like they do with rdsCA certs whenever they are ready to enable it.

Do we want to consider throwing a new tls port into this PR while we're at it? We won't remove the standard port, but at least we'll have a new one.

kylape:
There are a few factors going into this whole initiative:

* Consoledot is being pushed on several fronts to make all inter-service communication use TLS.  These various initiatives have different timelines, so different environments will need to be in different states of TLS-only modes
* While it is true that there is no enforcement by clowder when providing TLS certs to services, the enforcement implicitly comes from clients that need to connect to these services.  How do they know when to use TLS?
* This seems obvious, but: services will have to configure TLS before any clients can connect to them using TLS.

Given these factors, here is a sketch of a larger proposal:

* Just add the new annotation to every service resource; there is no need for a flag to "enable" TLS
* Either include the TLS certs directly in cdappconfig.json or mount them separately and add a pointer
* Add another port to every container for TLS (again no need for a flag)
* Add a flag to ClowdEnv to disable the plain-text port in the service.  This is just to make it clear during an audit when TLS is exclusively used.
* Clients will then need to connect with some extra logic during the transition: Try HTTPS; if it fails, then try HTTP.
* Communicate the change out to teams asap so they can start making required changes to both services and clients.

Perhaps this should be an ADR?  Sorry that this small change has ballooned into a Whole Thing!

psav:
Should we close this off in favour of the new ADR?

psav:
We implemented this using a different method

Issue #579, opened 2022-03-23
Broken links in Getting Started doc page


Fynardo:
I found what I think are a few broken links in the [getting started page](https://github.com/RedHatInsights/clowder/blob/master/docs/antora/modules/usage/pages/getting-started.adoc).

I tried to investigate where these links should point in order to try and help fix them, as follows:

- In [Install clowder in your local box](https://github.com/RedHatInsights/clowder/blob/master/docs/antora/modules/usage/pages/getting-started.adoc#install-clowder-on-your-local-box) section there is a link for the [developer guide](https://github.com/RedHatInsights/clowder/blob/master/docs/developer-guide.md) that I think should point here: https://github.com/RedHatInsights/clowder/blob/master/docs/antora/modules/ROOT/pages/developer-guide.adoc
- In [Create your first ClowdEnvironment](https://github.com/RedHatInsights/clowder/blob/master/docs/antora/modules/usage/pages/getting-started.adoc#create-your-first-clowdenvironment) there is a link for the [ephemeral environments getting started](https://clouddot.pages.redhat.com/docs/dev/getting-started/ephemeral/index.html) that I think should point here: https://consoledot.pages.redhat.com/docs/dev/getting-started/ephemeral/index.html
- Also in this section, there is a link pointing to the [default ephemeral template](https://github.com/RedHatInsights/bonfire/blob/master/bonfire/resources/ephemeral-clowdenvironment.yaml) which may point here instead: https://github.com/RedHatInsights/bonfire/blob/master/bonfire/resources/ephemeral-cluster-clowdenvironment.yaml
- In [Create your first ClowdApp](https://github.com/RedHatInsights/clowder/blob/master/docs/antora/modules/usage/pages/getting-started.adoc#create-your-first-clowdapp) section there is a link to [ClowdApps API](https://redhatinsights.github.io/clowder/api_reference.html#k8s-api-cloud-redhat-com-clowder-v2-apis-cloud-redhat-com-v1alpha1-clowdapp) which I think should point here: https://redhatinsights.github.io/clowder/clowder/dev/api_reference.html

Hope this helps

psav:
@gburges Can you fix these links?

gburges:
yeah, i actually have a half done PR about this ;)

PR #746, opened 2023-01-04
Fix broken links https://github.com/RedHatInsights/clowder/issues/579

gburges:


kylape:
Fixes #579

kylape:
New links work for me, though I don't know if it's preferable to point to internal docs rather than external ones that are available.

Victoremepunto (approved):
:+1:  thanks a lot, lets get this merged

Victoremepunto:
> New links work for me, though I don't know if it's preferable to point to internal docs rather than external ones that are available.

That is actually a good point. Though to be fair, https://redhatinsights.github.io/clowder was (and is) publicly accessible. Should it ? more important, should the consoledot pages be open as well ?

maknop (approved):

kylape:
> https://redhatinsights.github.io/clowder was (and is) publicly accessible. Should it ?

Since the github project is public, I see no reason to make the docs private.  Main question in my mind is if it's worth the maintenance effort.

> should the consoledot pages be open as well ?

Sounds like  PM question :)

gburges:
I made the link point to the internal docs, because the EE docs only live there. That said, i have 2 questions, does the Clowder-ENO-Bonfire work without our tooling (qontract-server and bonfire is public too) without a large overhead/extra coding to do, so would it make sense to give out a 'half-solution' publicly?
Should we publish our internal docs? or more like, Why shouldn't we make everything public?

kylape:
> Why shouldn't we make everything public?

There is likely internal infrastructure details in the internal docs repo.  Perhaps we could split that out and make part of it external at some point, but that is likely something that will have to be planned out.  So for now, let's maintain the status quo by keeping the links to the internal docs site.

PR #749, opened 2023-01-05
Do not run tests within pr_check if changes only affect doc files

Victoremepunto:


adamrdrew (approved):
Very good idea!

Victoremepunto:
> Very good idea!

Thanks :) - I actually wonder if it's worth to extend it to other areas of the repo - not only docs - for instance, github actions files - those are of no interest to pr_check, I believe

PR #750, opened 2023-01-06
Logging is now awesome

psav:


adamrdrew (approved):
Great revision. Much more readable. I like concept of "filter" much more than "predicate" for naming. Good DRY and elimination of those functionally identical methods.

PR #752, opened 2023-01-12
Fix bug where deployment figures no longer get updated if the entire …

psav:
…status does not change


psav:
@adamrdrew need a review on this please

PR #753, opened 2023-01-13
Add logLevel to CJI spec which sets IQE_LOG_LEVEL

bsquizz:
Currently we're unable to change the log level of IQE when a CJI is invoked. This PR gives us that ability. Default log level passed to the pod will be 'info' and the field allows 'omitempty' for backward compatibility.


mshriver (commented):
LGTM, consistent, don't know clowder well enough to 'approve' here.

Thanks @bsquizz!

adamrdrew (approved):
Looks good to me! A spec change with the related doc and related-test changes. The real heavy lifting will be on the pod, this just passes some info through. I'm good with this!

adamrdrew:
This is a small and clear enough change that I'm just gonna merge it - will allow us to soak in crcd and if we have a problem back-out will be super easy.

PR #758, opened 2023-01-23
Fix typo in docs

adamrdrew:


adamrdrew:
I have no idea why this pr_check is failing - I don't think we should even be running a pr_check on a doc change based on https://github.com/RedHatInsights/clowder/pull/749

And the reason it is failing isn't even accurate. Its saying the commit isn't signed - which is not true:

```
09:30:13 exec gpg --no-default-keyring --keyring=$PWD/git.gpg "$@"'
09:30:13 + chmod +x gitgpg
09:30:13 + git config gpg.program /srv/volume/jenkins/workspace/RedHatInsights-clowder-pr-check/gitgpg
09:30:13 + git verify-commit 9527ed776b7c752062b897b333ae195c3ef4d783
09:30:13 gpg: Signature made Mon Jan 23 14:27:10 2023 UTC
09:30:13 gpg:                using RSA key E4C357237CC4A94E
09:30:13 gpg: Can't check signature: No public key
```

What's the deal?

PR #761, opened 2023-01-25
Adding annotation to pods for autoscale

gburges:


psav:
I'm cared that just blindly enabling this will make things unstable, we should have the ability to turn this on and off. Else anytime OS tries to rebalance pod load, these pods could get evicted.

gburges:
> I'm cared that just blindly enabling this will make things unstable, we should have the ability to turn this on and off. Else anytime OS tries to rebalance pod load, these pods could get evicted.

yeah, that would be more ideal to have a switch that does it. But there is an option in autoscaler to turn off balancing, so we could 'solve' it from the other side

psav:
What I was trying to say earlier was that this PR doesn't apply any guard rails - the eviction is always turned on

psav (changes requested):
I think we should have a config item, or a ClowdEnvironment option to turn this on - else it's on everywhere all the time - I knwo there is a cluster level thing to turn it on, but......I still think we should make it conditional

gburges:
How about a ClowdEnv annotation? if autoscaleEvict = true, then it applies?

psav:
I'm less keen on this as it means we are adding a new API layer now

psav:
Closing as per @gburges request

PR #765, opened 2023-02-02
Support customizable rds ca bundle URL

bsquizz:
Currently when the DB provider is in 'app-interface' mode, it fetches an rds CA bundle .pem from a hard-coded URL. The CA bundle used in AWS govcloud is a completely different bundle, so we need a way to adjust the bundle URL.

This PR adds support for a new field under the ClowdEnvironment's database config named 'caBundleURL'. If it is not present, we continue to use the same URL we've been using all this time.

If it is present, Clowder will download the bundle from that URL instead.

Previously we were storing the downloaded bundle in a string to prevent re-downloading it every time the ClowdEnvironment was reconciled. Now since it is possible to have a different URL across different ClowdEnvironments, we store the downloaded bundle in a map where the keys are the 'bundle URL' and the values are the 'bundle content'. This should help prevent us from re-downloading a bundle, but we now will cache the fetched bundle for each separate URL.

### DIFF
--- a/controllers/cloud.redhat.com/providers/database/appinterface.go
+++ b/controllers/cloud.redhat.com/providers/database/appinterface.go
@@ -18,16 +18,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var rdsCa string
+var rdsCaBundles = make(map[string]string)
 
-const caURL string = "https://s3.amazonaws.com/rds-downloads/rds-combined-ca-bundle.pem"
+const defaultCaBundleURL string = "https://s3.amazonaws.com/rds-downloads/rds-combined-ca-bundle.pem"
 
 type appInterface struct {
 	providers.Provider
 }
 
-func fetchCa() (string, error) {
-	resp, err := http.Get(caURL)
+func fetchCa(caURL string) (string, error) {
+	resp, err := http.Get(caURL) // nolint
 
 	if err != nil {
 		return "", errors.Wrap("Error fetching CA bundle", err)
@@ -60,15 +60,21 @@ func NewAppInterfaceDBProvider(p *providers.Provider) (providers.ClowderProvider
 }
 
 func (a *appInterface) EnvProvide() error {
-	if rdsCa == "" {
-		_rdsCa, err := fetchCa()
+	caURL := a.Env.Spec.Providers.Database.CaBundleURL
+	if caURL == "" {
+		caURL = defaultCaBundleURL
+	}
+
+	if rdsCaBundles[caURL] == "" {
+		_rdsCa, err := fetchCa(caURL)
 
 		if err != nil {
 			return errors.Wrap("Failed to fetch RDS CA bundle", err)
 		}
 
-		rdsCa = _rdsCa
+		rdsCaBundles[caURL] = _rdsCa
 	}
+
 	return nil
 }
 
@@ -106,7 +112,8 @@ func (a *appInterface) Provide(app *crd.ClowdApp) error {
 		searchAppName = refApp.Name
 	}
 
-	matched, err := GetDbConfig(a.Ctx, a.Client, namespace, app.Name, searchAppName, dbSpec)
+	rdsCaBundleURL := a.Env.Spec.Providers.Database.CaBundleURL
+	matched, err := GetDbConfig(a.Ctx, a.Client, namespace, app.Name, searchAppName, dbSpec, rdsCaBundleURL)
 
 	if err != nil {
 		return err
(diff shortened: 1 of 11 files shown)


controllers/cloud.redhat.com/providers/database/appinterface.go
```diff
@@ -60,15 +60,26 @@ func NewAppInterfaceDBProvider(p *providers.Provider) (providers.ClowderProvider
...
+	if caURL == "" {
+		caURL = defaultCaBundleURL
+	}
+
+	_, ok := rdsCaBundles[caURL]
+	if !ok {
+		rdsCaBundles[caURL] = ""
+	}
+
+	if rdsCaBundles[caURL] == "" {
```
psav, line 73: why not put this inside the above conditional?

controllers/cloud.redhat.com/providers/database/appinterface.go
```diff
@@ -162,10 +176,13 @@ func GetDbConfig(ctx context.Context, pClient client.Client, namespace, appName,
 	// The creds given by app-interface have elevated privileges
 	matched.Config.AdminPassword = matched.Config.Password
 	matched.Config.AdminUsername = matched.Config.Username
-	matched.Config.RdsCa = &rdsCa
+	bundle, ok := rdsCaBundles[rdsCaBundleURL]
```
psav, line 179: Slightly less complex
```
	b := map[string]string{"a": "b"}
	a := ""
	a, _ = b["c"]
```

thoughts? - removes `if` reduces complexity

psav (changes requested):

controllers/cloud.redhat.com/providers/database/appinterface.go
bsquizz: I am checking for 2 separate conditions:
1. the key may not yet exist at all
2. the key may exist, but be and empty string because the last time an HTTP GET was attempted it failed

tests/kuttl/test-multi-app-interface-db/02-json-asserts.yaml
```diff
@@ -3,20 +3,28 @@ apiVersion: kuttl.dev/v1beta1
...
...GEwJVUzETMBEGA1UE\nCAwKV2FzaGluZ3RvbjEQMA4GA1UEBwwHU2VhdHRsZTEiMCAGA1UECgwZQW1hem9u\nIFdlYiBTZXJ2aWNlcywgSW5jLjETMBEGA1UECwwKQW1hem9uIFJEUzEhMB8GA1UE\nAwwYQW1hem9uIFJEUyBldS1zb3V0aC0xIENBMIIBIjANBgkqhkiG9w0BAQEFAAOC\nAQ8AMIIBCgKCAQEAtEyjYcajx6xImJn8Vz1zjdmL4ANPgQXwF7+tF7xccmNAZETb\nbzb3I9i5fZlmrRaVznX+9biXVaGxYzIUIR3huQ3Q283KsDYnVuGa3mk690vhvJbB\nQIPgKa5mVwJppnuJm78KqaSpi0vxyCPe3h8h6LLFawVyWrYNZ4okli1/U582eef8\nRzJp/Ear3KgHOLIiCdPDF0rjOdCG1MOlDLixVnPn9IYOciqO+VivXBg+jtfc5J+L\nAaPm0/Yx4uELt1tkbWkm4BvTU/gBOODnYziITZM0l6Fgwvbwgq5duAtKW+h031lC\n37rEvrclqcp4wrsUYcLAWX79ZyKIlRxcAdvEhQIDAQABo2YwZDAOBgNVHQ8BAf8E\nBAMCAQYwEgYDVR0TAQH/BAgwBgEB/wIBADAdBgNVHQ4EFgQU7zPyc0azQxnBCe7D\nb9KAadH1QSEwHwYDVR0jBBgwFoAUFBAFcgJe/BBuZiGeZ8STfpkgRYQwDQYJKoZI\nhvcNAQELBQADggEBAFGaNiYxg7yC/xauXPlaqLCtwbm2dKyK9nIFbF/7be8mk7Q3\nMOA0of1vGHPLVQLr6bJJpD9MAbUcm4cPAwWaxwcNpxOjYOFDaq10PCK4eRAxZWwF\nNJRIRmGsl8NEsMNTMCy8X+Kyw5EzH4vWFl5Uf2bGKOeFg0zt43jWQVOX6C+aL3Cd\npRS5MhmYpxMG8irrNOxf4NVFE2zpJOCm3bn0STLhkDcV/ww4zMzObTJhiIb5wSWn\nEXKKWhUXuRt7A2y1KJtXpTbSRHQxE++69Go1tWhXtRiULCJtf7wF2Ksm0RR/AdXT\n1uR1vKyH5KBJPX3ppYkQDukoHTFR0CpB+G84NLo=\n-----END CERTIFICATE-----\n"' -e < /tmp/test-multi-app-interface-db-default-ca-json
```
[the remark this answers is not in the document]
bsquizz: I thought about that, just wasn't sure what dir to put files like that in, but I'll experiment.

controllers/cloud.redhat.com/providers/database/appinterface.go
bsquizz: I was trying to avoid assigning a nil pointer but actually I don't think I need any of this. Looks like the value of `b["c"]` is already `""`

controllers/cloud.redhat.com/providers/database/appinterface.go
bsquizz: OK no need to check for the key presence, I can just combine this into 1 'if' ... `if rdsCaBundles[caURL] == "" {`

bsquizz:
OK, addressed the feedback and kuttl is passing locally

psav (approved):

PR #769, opened 2023-02-03
Support customizable sasl mechanism in 'managed' kafka mode

bsquizz:
Allows the SASL mechanism to optionally be set in the secret that clowder looks up when running the kafka provider in 'managed' mode. If the key is absent, `PLAIN` is used as the mechanism.


psav:
lgtm

psav (approved):

PR #788, opened 2023-03-14
Makes clowder install properly on local minikube

gburges:


config/manager/kustomization.yaml
```diff
@@ -17,5 +17,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: 127.0.0.1:5000/clowder
-  newTag: b7db066
+  newName: quay.io/cloudservices/clowder
```
psav, line 20: This shouldn't be updated - `make pre-push` should clean it up

controllers/cloud.redhat.com/version.txt
```diff
@@ -1 +1 @@
-master
\ No newline at end of file
+v0.52.0-122-g312373b9
```
psav, line 1: This also shouldn't be updated

controllers/cloud.redhat.com/clowderconfig/config.go
```diff
@@ -51,7 +51,7 @@ type ClowderConfig struct {
 }
 
 func getConfig() ClowderConfig {
-	configPath := "/config/clowder_config.json"
+	configPath := "./config/clowder_config.json"
```
psav, line 54: This could break the container in OpenShift

config/clowder_config.json
```diff
@@ -0,0 +1,24 @@
+    {
```
psav, line 1: We have a sample config......was there something missing from it?

psav (changes requested):
Let's talk through the problems you were solving with this one, my install works fine - I don't want to do a `works for me`, but I think we should discuss this one, I maybe have something local that you don't, but I don't think it's this.

PR #793, opened 2023-03-23
[RHCLOUD-24811] Adds ability for jobs to be disabled

psav:
Currently there is no way to disable jobs. This PR gives the ability for jobs to be disabled if they are not required to run or need to be disabled for some time.

Also includes documentation and test updates.

### DIFF
--- a/controllers/cloud.redhat.com/providers/cronjob/default.go
+++ b/controllers/cloud.redhat.com/providers/cronjob/default.go
@@ -27,7 +27,7 @@ func (j *cronjobProvider) Provide(app *crd.ClowdApp) error {
 
 	for _, cronjob := range app.Spec.Jobs {
 		innerCronjob := cronjob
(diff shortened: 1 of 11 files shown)


controllers/cloud.redhat.com/providers/cronjob/default.go
```diff
@@ -27,7 +27,7 @@ func (j *cronjobProvider) Provide(app *crd.ClowdApp) error {
 
 	for _, cronjob := range app.Spec.Jobs {
 		innerCronjob := cronjob
-		if innerCronjob.Schedule != "" {
+		if innerCronjob.Schedule != "" && !cronjob.Disabled {
```
bsquizz, line 30: Just curious -- I do not know the reasoning behind the variable re-assignment to start with -- but is there a reason you chose to check `cronjob.Disabled` instead of `innerCronjob.Disabled` here?

PR #799, opened 2023-04-04
Migrate to appsre unleash

Victoremepunto:


Victoremepunto:
Unsure at this point if this is worth it at all. 

- Can't use (or don't know how to use) public routes for Keycloak (so Unleash's GUI is disabled to all effects)
- Having AppSRE's Unleash adds little to no value compared to Vanilla Unleash with GUI disabled (in fact we loose that compared to vanilla Unleash)
- To fix the kuttl test, it REALLY needs to do better than rely on `wget` command present in the image, extending the scope of this PR.

May be superseded (again) by simply #912 

@adamrdrew @psav  I'd appreciate your thoughts here

PR #802, opened 2023-04-17
Add parallelEnabled and parallelWorkerCount to CJI spec which sets corresponding IQE vars.

san7ket:
There are two variables those were added to iqecore to handle the way we trigger tests in parallel, this would benefit the most with resource and time consuming UI tests.

### DIFF
--- a/controllers/cloud.redhat.com/providers/iqe/impl.go
+++ b/controllers/cloud.redhat.com/providers/iqe/impl.go
@@ -61,6 +61,8 @@ func createIqeContainer(j *batchv1.Job, nn types.NamespacedName, cji *crd.ClowdJ
 		{Name: "IQE_REQUIREMENTS", Value: joinNullableSlice(cji.Spec.Testing.Iqe.Requirements)},
 		{Name: "IQE_REQUIREMENTS_PRIORITY", Value: joinNullableSlice(cji.Spec.Testing.Iqe.RequirementsPriority)},
 		{Name: "IQE_TEST_IMPORTANCE", Value: joinNullableSlice(cji.Spec.Testing.Iqe.TestImportance)},
+		{Name: "IQE_PARALLEL_ENABLED", Value: cji.Spec.Testing.Iqe.ParallelEnabled},
+		{Name: "IQE_PARALLEL_WORKER_COUNT", Value: cji.Spec.Testing.Iqe.ParallelWorkerCount},
 	}
 
 	// set image tag
--- a/apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
+++ b/apis/cloud.redhat.com/v1alpha1/clowdjobinvocation_types.go
@@ -83,6 +83,12 @@ type IqeJobSpec struct {
 	// sets value for IQE_LOG_LEVEL (default if empty: "info")
 	//+kubebuilder:validation:Enum={"", "critical", "error", "warning", "info", "debug", "notset"}
 	LogLevel string `json:"logLevel,omitempty"`
+
+	// sets value passed to IQE 'IQE_PARALLEL_ENABLED' arg
+	ParallelEnabled string `json:"parallelEnabled,omitempty"`
+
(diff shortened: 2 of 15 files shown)


san7ket:
@bsquizz added the two variables from IQE hopefully it works, do we also need updates in bonfire ?

san7ket:
Any pointers on why the check is failing ?

bsquizz (changes requested):
This is looking ok so far -- you are just missing one key piece which is to set the env var values onto the pod at this section of code:

https://github.com/RedHatInsights/clowder/blob/master/controllers/cloud.redhat.com/providers/iqe/impl.go#L52

You'll want 2 new entries:
```
{Name: "IQE_PARALLEL_ENABLED", Value: cji.Spec.Testing.Iqe.ParallelEnabled},
{Name: "IQE_PARALLEL_WORKER_COUNT", Value: cji.Spec.Testing.Iqe.ParallelWorkerCount},
```

If empty strings get passed in for these env vars it is fine, the IQE container entrypoint will just use the defaults set in `run.py` in that case.

bsquizz:
> do we also need updates in bonfire ?

Yes, we would need an update in bonfire as well for the `bonfire process-iqe-cji` and `bonfire deploy-iqe-cji` commands to be able to set these 2 new options.

1. We'd need to update the IQE CJI template here to add the 2 new options:
https://github.com/RedHatInsights/bonfire/blob/master/bonfire/resources/default-iqe-cji.yaml#L23

2. Then add new `click` CLI arguments that allow you to change the values passed to the CJI template if desired:
https://github.com/RedHatInsights/bonfire/blob/6c639bec0716aae0594f0fb51f76c213219200cc/bonfire/bonfire.py#L547

3. Then add the new click args to these 2 functions:
* https://github.com/RedHatInsights/bonfire/blob/6c639bec0716aae0594f0fb51f76c213219200cc/bonfire/bonfire.py#L1272
* https://github.com/RedHatInsights/bonfire/blob/6c639bec0716aae0594f0fb51f76c213219200cc/bonfire/bonfire.py#L1314

4. Finally, you'll notice those 2 functions above call `process_iqe_cji` -- you'd want to pass the new click args on to that function. That is the function that handles running `oc process` on the template: https://github.com/RedHatInsights/bonfire/blob/6c639bec0716aae0594f0fb51f76c213219200cc/bonfire/processor.py#L128

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -61,6 +61,8 @@ func createIqeContainer(j *batchv1.Job, nn types.NamespacedName, cji *crd.ClowdJ
 		{Name: "IQE_REQUIREMENTS", Value: joinNullableSlice(cji.Spec.Testing.Iqe.Requirements)},
 		{Name: "IQE_REQUIREMENTS_PRIORITY", Value: joinNullableSlice(cji.Spec.Testing.Iqe.RequirementsPriority)},
 		{Name: "IQE_TEST_IMPORTANCE", Value: joinNullableSlice(cji.Spec.Testing.Iqe.TestImportance)},
+		{Name: "IQE_PARALLEL_ENABLED", Value: cji.Spec.Testing.Iqe.ParallelEnabled},
+		{Name: "IQE_PARALLEL_WORKER_COUNT", Value: cji.Spec.Testing.Iqe.ParallelWorkerCount},
```
san7ket, line 65: @bsquizz have already added these

san7ket:
This helps I will work on bonfire changes early next week

controllers/cloud.redhat.com/providers/iqe/impl.go
bsquizz: thanks, not sure how I missed this :)

bsquizz (approved):

adamrdrew:
Small surface area and kuttl tests Looks good to me!

adamrdrew (approved):
This all looks good to me!

PR #805, opened 2023-04-18
Added envoy config for stats

psav:
Creating the admin endpoint and hooking up a route to interface with metrics


psav:
Closing as we moved to caddy

PR #817, opened 2023-05-23
Commneting out KEDA stuff

gburges:


psav:
There is a config item to turn off caring about KEDA - we don't need to remove code for this

PR #826, opened 2023-06-09
Nokedali

gburges:


psav:
I'm not sure we want to make this the default - can you give me the context again for this?

gburges:
because we added enableKedaResources": ${ENABLE_KEDA_RESOURCES} to the template, we can't just delete the values. The manager won't be using it by default (it's set to false), just the template default will be true. But with the CMA conundrum still in progress, this can stay on hold for a while

gburges:
closing this as we're going to test and explicitly enable on a per cluster basis first

PR #829, opened 2023-06-14
Add api path

psav:


controllers/cloud.redhat.com/providers/dependencies/impl.go
```diff
@@ -130,6 +137,7 @@ func processAppEndpoints(
 					Name:     innerDeployment.Name,
 					App:      depApp.Name,
 					TlsPort:  utils.IntPtr(int(tlsPort)),
+					ApiPath:  fmt.Sprintf("/api/%s/", apiPath),
```
Victoremepunto, line 140: Not sure I like the root of the path being hardcoded like this ( I mean, forcing the endpoint to always be `/api/your-path` ) 

I understand it's heavily unlikely that this changes on Stage and Prod - however unlikely, I think this part (the ROOT api) should be something configurable. I anything, could this be set at a "higher" level (ClowdEnvironment level) and picked up from there ? I don't know if this makes sense, ultimately I'd like to avoid having to change the source code if we change the root path.

docs/appconfig/schema-definitions-dependencyendpoint.md
```diff
@@ -104,3 +105,19 @@ The TLS port of the dependent service.
...
+
+## apiPath
+
+The top level api path that the app should serve from /api/<apiPath>
+
+
+`apiPath`
+
+-   is required
+-   Type: unknown
```
Victoremepunto, line 117: shouldn't the type be `String` ?

Victoremepunto (approved):
I've left a couple of comments, I'll approve, you decide whether to address them or not.

controllers/cloud.redhat.com/providers/dependencies/impl.go
psav: I think we can look at this further down the line - I'm not afraid of putting a configuration item in later. But making it a mandatory field right now is a no-go and then we need yet more discussions on how to implement this. Unfortunately `/api` is already hardcoded in other places. Yes it's non-ideal but I think if we fix it, we fix it everywhere of course, and I don't want to do that in this PR.

docs/appconfig/schema-definitions-dependencyendpoint.md
psav: Auto generated - I'm not going to fix this rabbit hole now

docs/appconfig/schema.md
```diff
@@ -1475,6 +1476,22 @@ The TLS port of the dependent service.
...
 
+### apiPath
+
+The top level api path that the app should serve from /api/<apiPath>
+
+
+`apiPath`
+
+-   is required
+-   Type: unknown
```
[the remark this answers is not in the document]
psav: Same as before

Victoremepunto (approved):

PR #830, opened 2023-06-14
[RHCLOUD-27862] add Pod Lifecyle hooks

maskarb:
https://issues.redhat.com/browse/RHCLOUD-27862

minimal example (use the example clowdenv.yml):
```
kind: ClowdApp
metadata:
  name: hello
  namespace: jumpstart
spec:
  envName: env-jumpstart
  deployments:
  - name: app
    podSpec:
      image: nginx
      terminationGracePeriodSeconds: 500
      lifecycle:
        postStart:
          exec:
            command:
            - "/bin/sh"
            - "-c"
            - echo "$(date -Iseconds) ...MESSAGE FROM THE POSTSTART HOOK..." > /proc/1/fd/1
        preStop:
          exec:
            command:
              - "/bin/sh"
              - "-c"
              - echo "$(date -Iseconds) ...MESSAGE FROM THE PRESTOP HOOK..." > /proc/1/fd/1
    webServices:
      private:
        enabled: true
      public:
        enabled: true
```
(These hooks are a variation from: https://thecloudblog.net/lab/kubernetes-container-lifecycle-events-and-hooks/. The `echo "message" > /proc/1/fd/1` prints that message to the pod logs so we can see the hooks in action)

The deployment generated by this clowdapp contains:
```
...
          lifecycle:
            postStart:
              exec:
                command:
                  - /bin/sh
                  - '-c'
                  - echo "$(date -Iseconds) ...MESSAGE FROM THE POSTSTART HOOK..." > /proc/1/fd/1
            preStop:
              exec:
                command:
                  - /bin/sh
                  - '-c'
                  - echo "$(date -Iseconds) ...MESSAGE FROM THE PRESTOP HOOK..." > /proc/1/fd/1
          ...
      terminationGracePeriodSeconds: 500
...
```

Using this image and lifecycle hooks, open the pod terminal and see the printed log messages. The readiness/liveness probes will fail for this container. So when the pod restarts, we can see the preStop hook just before the sigquit was received:
```
2023/11/01 21:44:08 [notice] 1#1: start worker process 34
2023-11-01T21:44:08+00:00 ...MESSAGE FROM THE POSTSTART HOOK...
2023-11-01T21:45:37+00:00 ...MESSAGE FROM THE PRESTOP HOOK...
2023/11/01 21:45:37 [notice] 1#1: signal 3 (SIGQUIT) received, shutting down
```


psav:
Can you run `make pre-push` on this please?

maskarb:
> Can you run `make pre-push` on this please?

Done!

I will note, I have not done any testing of these changes at all yet. I was going to see if I can get this working, and then see if yall had any thoughts on these changes to the Spec.

psav:
I have no problem with the spec changes

docs/antora/modules/ROOT/pages/api_reference.adoc
```diff
@@ -23,24 +23,8 @@ Package v1alpha1 contains API Schema definitions for the cloud.redhat.com v1alph
 
 
 
-[id="{anchor_prefix}-github-com-redhatinsights-clowder-apis-cloud-redhat-com-v1alpha1-apipath"]
-==== APIPath
-
-_Underlying type:_ _string_
```
bsquizz, line 29: Things look good but I think some of these api doc formatting changes are related to version mismatches on some of the generation tools. Try this:

1.  `rm -fr docs/build/crd-ref-docs`
2. Make sure you're using asciidoctor v2.0.20 (`asciidoctor -v`) if not, `sudo dnf update -y asciidoctor` or similar
3. `make pre-push` (crd-ref-docs will be downloaded and installed again at this point)


The doc changes to `apiPaths` we are seeing generated here will go in via #888

bsquizz (changes requested):

tests/kuttl/test-lifecycle-hook/01-pods.yaml
```diff
@@ -0,0 +1,61 @@
...
+  name: lifecycle
+  namespace: test-lifecycle-hook
+spec:
+  envName: test-lifecycle-hook
+  deployments:
+  - name: processor
+    podSpec:
+      image: quay.io/psav/clowder-hello
+      lifecycle:
+        # postStart:
```
adamrdrew, line 46: If this is commented-out can we remove it?

controllers/cloud.redhat.com/providers/deployment/impl.go
```diff
@@ -270,6 +271,8 @@ func initDeployment(app *crd.ClowdApp, env *crd.ClowdEnvironment, d *apps.Deploy
 
 	d.Spec.Template.Spec.InitContainers = ics
 
+	d.Spec.Template.Spec.TerminationGracePeriodSeconds = pod.TerminationGracePeriodSeconds
```
adamrdrew, line 274: Why is this added in this PR? This doesn't look related to the lifecycle hooks at first glance.

adamrdrew (commented):
LGTM really, but a couple of questions before approval

controllers/cloud.redhat.com/providers/deployment/impl.go
maskarb: This is needed for hooks that can take a long time. We need this for Trino which will likely take about 10-20 minutes to fully shutdown because of the work it needs to do before it can say it is done with work. The default termination is 30 seconds, so it delete the pod before the pod was actually done.

tests/kuttl/test-lifecycle-hook/01-pods.yaml
maskarb: I really left this here to show how else it could be used, but I'll remove it since there is an example in this PR description.

docs/appconfig/schema-definitions-dependencyendpoint-properties-apipaths-items.md
```diff
@@ -0,0 +1,16 @@
+# Untitled string in AppConfig Schema
```
bsquizz, line 1: Still looks like an unrelated API docs change, probably fixed now that #888 went in... I think you can delete this file and if you have revised then "make pre-push" should not add it again...

docs/appconfig/schema-definitions-dependencyendpoint-properties-apipaths.md
```diff
@@ -0,0 +1,16 @@
+# Untitled array in AppConfig Schema
```
bsquizz, line 1: Same for this file

bsquizz (changes requested):
Sorry just a couple minor doc generator things again

bsquizz (approved):
I'll sort out those doc generation shenanigans later, it doesn't need to hold up this PR

controllers/cloud.redhat.com/providers/deployment/impl.go
psav, line 274: I _think_ we need to set the default here to 30, because otherwise we will get 0 which would change the default behaviour on the pods. We should have a test that tests the non-setting option too to ensure the default is what we expect.

maskarb: I'll add a test and update if we need to!

maskarb: I pushed a couple more tests, one of which completely leaves off the `terminationGracePeriodSeconds` in the clowdapp, and the assertion checks that it is correctly set to 30.

psav (approved):

psav:
Pending test outcome I'm happy to merge this

PR #839, opened 2023-07-07
fix(dev): install Floorist CRD

vkrizan:
Install Floorist operator CRDs in kube_setup.sh for minikube. This is to prevent errors when applying templates and resources that include FloorPlan CRs.

cc @fijshion

### DIFF
--- a/build/kube_setup.sh
+++ b/build/kube_setup.sh
@@ -302,6 +302,13 @@ function install_subscription_crd {
     cd "$ROOT_DIR"
 }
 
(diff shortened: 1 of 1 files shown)


vkrizan:
Needs testing...

bsquizz:
I believe the PR check is currently failing because you need to gpg sign the commit

vkrizan:
@bsquizz oh, thx

vkrizan:
via #840

PR #841, opened 2023-07-14
add RunOnNotReady feature to cji

gburges:


controllers/cloud.redhat.com/clowdjobinvocation_controller.go
```diff
@@ -142,15 +142,17 @@ func (r *ClowdJobInvocationReconciler) Reconcile(ctx context.Context, req ctrl.R
 
 	// Determine if the ClowdApp containing the Job is ready
 	if !app.IsReady() {
-		r.Recorder.Eventf(&app, "Warning", "ClowdAppNotReady", "ClowdApp [%s] is not ready", cji.Spec.AppName)
-		r.Log.Info("App not yet ready, requeue", "jobinvocation", cji.Spec.AppName, "namespace", app.Namespace)
-		readyErr := errors.NewClowderError(fmt.Sprintf("The %s app must be ready for CJI to start", cji.Spec.AppName))
-		if condErr := SetClowdJobInvocationConditions(ctx, r.Client, &cji, crd.ReconciliationFailed, readyErr); condErr != nil {
-			return ctrl.Result{}, condErr
-		}
+		if !cji.Spec.RunOnNotReady {
```
adamrdrew, line 145: The negative conditional nested in a negative conditional is hard to read and reason about. And this makes a large method larger. I think you could improve this code by factoring this logic out into a method (reduce method complexity) and recast the conditionals as positive non-nested guards. Here's a sketch (didn't test this):

Recast as a method
```golang
func (r *ClowdJobInvocationReconciler) HandleNotReady(... fill out correct args here ...) {
	if app.IsReady() {
		return nil
	}
	if cji.Spec.RunOnNotReady {
		return nil
	}
	r.Recorder.Eventf(&app, "Warning", "ClowdAppNotReady", "ClowdApp [%s] is not ready", cji.Spec.AppName)
	r.Log.Info("App not yet ready, requeue", "jobinvocation", cji.Spec.AppName, "namespace", app.Namespace)
	readyErr := errors.NewClowderError(fmt.Sprintf("The %s app must be ready for CJI to start", cji.Spec.AppName))
	if condErr := SetClowdJobInvocationConditions(ctx, r.Client, &cji, crd.ReconciliationFailed, readyErr); condErr != nil {
		return condErr
	}

	// requeue with a buffer to let the app come up
	return readyErr
}
```

And calling it within your main method
```golang
notReadyError := HandleNotReady(... fill out correct args here ...)
if notReadyError != nil {
	return ctrl.Result{Requeue: true}, notReadyError
}
```suggestion
		if !cji.Spec.RunOnNotReady {
```
```

File: tests/kuttl/test-runonnotready-cji/01-assert.yaml [hunk 2]
```diff
@@ -0,0 +1,61 @@
+---
+apiVersion: cloud.redhat.com/v1alpha1
+kind: ClowdApp
+metadata:
+  name: puptoo
+  namespace: test-runonnotready-jobs
+status:
+  ready: false
```
adamrdrew, line 8: This assert is failing because your clowdapp is ready. You can slice this one of two ways. You can remove this assert and it should pass (or at least, not fail because of this specific problem) or you can try to force your clowdapp to not be ready, which if I'm understanding the feature correctly is probably what you want. Problem is I'm not sure how you'd force that. Maybe force the running imaghe to fail healthchecks?

adamrdrew (changes requested):

tests/kuttl/test-runonnotready-cji/01-assert.yaml
gburges: yeah, i made it a bad image instead so it fails to pull :)

adamrdrew:
LGTM!

psav (dismissed):

tests/kuttl/test-runonnotready-cji/01-pods.yaml
```diff
@@ -0,0 +1,65 @@
...
+kind: ClowdApp
+metadata:
+  name: puptoo
+  namespace: test-runonnotready-jobs
+spec:
+  envName: test-runonnotready-jobs
+  deployments:
+    - name: processor
+      podSpec:
+        image: quay.io/psav/clowder-hellou
```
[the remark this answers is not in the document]
gburges: yes, ty for pointing it out

PR #843, opened 2023-07-31
fix: make envoy tls cert optional

SteveHNH:
This should help us overcome a race condition where the secret is not available when the deployment is attempted. The envoy TLS secret is created after the service is defined and this causes a race condition on deployment.

We'll still need to reconcile the clowdapp after everything is stood up and available, but this at least allows the clowdapp to roll out successfully.


psav:
This seemed to fail on the jobs test

PR #844, opened 2023-08-01
SLO panel update, minor additional changes

maknop:
The following changes are in the PR:
- Added panel to reflect Clowder's SLO changes
- Plugin version updates
- SLO panel is now above the "App Reconciliation Requests" panel since we have one panel instead of three for SLOs
- "App Reconciliation Requests" panel did not have a transparent background. Made it transparent to match all other panels.

### DIFF
--- a/dashboards/clowder-dashboard.yml
+++ b/dashboards/clowder-dashboard.yml
@@ -1768,7 +1768,7 @@ data:
               "refId": "A"
             }
           ],
-          "title": "Reconciliation Time SLI",
+          "title": "Clowder Reconciliation Time Per Controller SLOs",
           "transparent": true,
           "type": "timeseries"
         },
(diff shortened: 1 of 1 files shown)


dashboards/clowder-dashboard.yml
```diff
@@ -194,7 +194,7 @@ data:
           "steppedLine": false,
           "targets": [
             {
-              "expr": "sum by (method) (rate(rest_client_requests_total{namespace=\"clowder-system\"}[1m]))",
+              "expr": "sum by (method) (rate(rest_client_requests_total[1m]))",
```
psav, line 197: Why dropping the namespace - this doesn't seem right? Without it we will be pulling metrics from any container that has the same metric exposed.

psav (changes requested):
Questions to answer please

PR #847, opened 2023-08-08
Adding PlatSec gh workflow

gburges:


psav:
This just validates th base image - is that enough?

gburges:
> This just validates th base image - is that enough?

it builds the whole image, those args are just needed because of multi stage build. I had @xxlhacker and @casey-williams-rh  check it out, they said :+1:

PR #848, opened 2023-08-22
Add gojsonschema command to makefile

adamrdrew:
This patch looks for the gojsonschema command to be installed and if it isn't installs it


adamrdrew:
Note I pinned the version to the latest current available, but didn't use the latest tag. The choice of version was arbitrary because I don't think we have on opinion on a "correct" version, but we don't want to use the latest tag.

psav (approved):

PR #849, opened 2023-08-25
use the latest tag of mbop

dagbay-rh:
## Summary

Use the `latest` tag of mbop when deploying default image.

Update to MBOP here to tag `latest`: https://github.com/RedHatInsights/mbop/pull/96

### DIFF
--- a/controllers/cloud.redhat.com/providers/utils/utils.go
+++ b/controllers/cloud.redhat.com/providers/utils/utils.go
@@ -23,7 +23,7 @@ import (
 )
 
 var DefaultImageCaddySideCar = "quay.io/cloudservices/crc-caddy-plugin:1c4882e"
-var DefaultImageMBOP = "quay.io/cloudservices/mbop:bb071db"
+var DefaultImageMBOP = "quay.io/cloudservices/mbop:latest"
 var DefaultImageMocktitlements = "quay.io/cloudservices/mocktitlements:e24820c"
 var DefaultKeyCloakVersion = "15.0.2"
 var DefaultImageKeyCloak = fmt.Sprintf("quay.io/keycloak/keycloak:%s", DefaultKeyCloakVersion)


dagbay-rh:
@gburges @adamrdrew can either of yall take a look at this when you have a sec? Not sure who to ping so I just looked at most recent contributors

dagbay-rh:
not sure whats actually failing the PR build, I see that no output being recorded:
```
13:13:03 ERROR: Step ?Publish JUnit test result report? failed: No test report files were found. Configuration error?
``` 
but its unclear whats actually causing the error

bsquizz (changes requested):
We normally do not use 'latest' because the recommened best practice is that pods use an `imagePullPolicy` of 'IfNotPresent' -- under that condition, if a pod runs on a node in the cluster that has already pulled an older version of `latest` it would not re-pull again.

bsquizz:
There is also 1 kuttl test assertion you'll have to update when you change the tag in the `.go` file:

```
tests/kuttl/test-ephemeral-gateway/01-assert.yaml:      - image: quay.io/cloudservices/mbop:bb071db
```

bsquizz:
Looks like latest tag is `a4aaa4c` -- if you bump the code to that (and update the kuttl test assertion) I think this will be good to go.

dagbay-rh:
@bsquizz gotcha! Thanks for the feedback :cowboy_hat_face:

Victoremepunto:
@dagbay-rh  please, mind the GPG commit signature, otherwise the `build` pipeline will fail.

dagbay-rh:
> @dagbay-rh please, mind the GPG commit signature, otherwise the `build` pipeline will fail.

cool thanks for the info :cowboy_hat_face: updated with signed commits

Victoremepunto:
@vbelchio this needs some more work as Basic Auth is not supported yet I think (CC @psav )

vbelchio:
@Victoremepunto can we have this PR merged once it's just bumping the MBOP version?

psav:
Working on something related - gonna need another day or so to get this working

PR #851, opened 2023-09-06
Cert auth

psav:
Still to do

- [x] permissions
- [x] caddy config change should reload container
- [x] openshift mode
- [x] tests
- [x] modify ingress controller ssl mode


controllers/cloud.redhat.com/providers/providers.go
```diff
@@ -157,16 +157,28 @@ func createResource(cache *rc.ObjectCache, resourceIdent rc.ResourceIdent, nn ty
...
 
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error creating object: %s%s %w", nn.Name, nn.Namespace, err)
 	}
 	return nobj, nil
 }
 
 func updateResource(cache *rc.ObjectCache, resourceIdent rc.ResourceIdent, object client.Object) error {
 	err := cache.Update(resourceIdent, object)
```
adamrdrew, line 166: If `err` is `nil` here can we just return? I'm not sure I totally understand but it seems like everything we do after updating the cache is only to contextualize an error we may get. If the cache update succeeds do we really need to do the rest of this stuff?

controllers/cloud.redhat.com/providers/providers.go
```diff
@@ -157,16 +157,28 @@ func createResource(cache *rc.ObjectCache, resourceIdent rc.ResourceIdent, nn ty
...
 
+	gvks, nok, werr := cache.GetScheme().ObjectKinds(resourceIdent.GetType())
+
+	if werr != nil {
```
adamrdrew, line 170: Could we have a more descriptive name or a comment? I don't know what this failure condition is.

adamrdrew (approved):
I think this is a good change set. The new code all makes sense to me. Going through it together helped which is why I have no questions, but it all makes sense. It follows what's idiomatic for clowder. The refactors here are good too. I don't really have anything to add except for the minor suggestions in the cache code and errors. LGTM!!!!

PR #852, opened 2023-09-07
Add AdditionalFilters field to cindy spec [RHCLOUD-28098]

gburges:
DO NOT MERGE UNTIL CYNDI RELEASES NEW VERSION


skateman (approved):
The Seal of Approval

apis/cloud.redhat.com/v1alpha1/clowdapp_types.go
```diff
@@ -369,6 +369,9 @@ type CyndiSpec struct {
 	// +kubebuilder:validation:Pattern:="[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
 	AppName string `json:"appName,omitempty"`
 
+	// AdditionalFilters
+	AdditionalFilters []map[string]string `json:"additionalFilters,omitempty"`
```
skateman, line 373:
```
// +optional
// +kubebuilder:default:={}
```
Maybe?

gburges: thought about it, prolly not need it :)

bsquizz: Since the field on the ClowdApp already allows empty via the `omitempty` syntax, we decided we'd just let cyndi-operator be in charge of stating what the default value should be if the field is empty.

psav (changes requested):
I'm expecting to see a `cyndi.go` change here, without it nothing will happen, the new spec will be accepted, but it won't go anywhere. We should also add something to the tests - I think there is a test for cyndipipeline that ensures this field ends up in the right place.

psav (changes requested):
The test here only tests that the new field is acceptable - it doesn't actually assert that those fields land in the cyndipipeline resource

tests/kuttl/test-cyndi-strimzi/02-json-asserts.yaml
```diff
@@ -19,3 +19,5 @@ commands:
 - script: EXPECTED=$(jq '.data["name"]' /tmp/myapp-db-secret); jq -e '.data["db.name"] == '$EXPECTED /tmp/myapp-db-cyndi-secret
 - script: USER=$(jq -r '.data["db.user"]' /tmp/myapp-db-cyndi-secret | base64 -d); [ "$USER" = "cyndi" ]
 - script: PW=$(jq -r '.data["db.password"]' /tmp/myapp-db-cyndi-secret | base64 -d); [ "$PW" = "cyndi" ]
+- script: kubectl get cyndipipeline -n test-cyndi-strimzi -o json > /tmp/cyndipipeline
+- script: EXPECTED=$(jq '."spec".additionalFilters' /tmp/cyndipipeline); jq -e '."spec".additionalFilters == '$EXPECTED /tmp/cyndipipeline
```
psav, line 23: New line at the end please!

tests/kuttl/test-cyndi-strimzi/01-install.yaml
```diff
@@ -43,6 +43,8 @@ metadata:
 spec:
   cyndi:
     enabled: true
+    additionalFilters:
```
psav, line 46: Did you check that this doesn't break if additionalFilters is missing?

psav (changes requested):

tests/kuttl/test-cyndi-strimzi/01-install.yaml
skateman: We have a default value set to `[]˙ in our app, if that helps

gburges: doesn't break if it's empty or missing either (tested both on my local minikube)

psav:
Need a rebase @gburges but, where are we on the cyndi release?

psav:
Can you rebase and fix conflicts please?

PR #856, opened 2023-09-08
Bumping Prometheus Operator version

gburges:


psav (changes requested):
This won't affect the prometheus that is deployed - but should be done in tandem with updating the prometheus on the cluster right?

gburges:
there's no prometheus deployed on the cluster - we removed it a while ago because it wasn't cleaning up properly

psav:
> there's no prometheus deployed on the cluster - we removed it a while ago because it wasn't cleaning up properly

While we disabled the prometheus instances by default, the "operator" is still installed and at a much earlier version `0.56.3` _this_ is what needs updating first.....or in tandem with the update to the CRD spec in Clowder

psav (changes requested):
Commented

psav:
Is prometheus updated in the clusters now?

psav (commented):
Is prometheus updated in the clusters now?

We should take a look at the version we deploy in the kubebuild script too

adamrdrew:
We need to see that we have the correct operator versions deployed. Also, "We should take a look at the version we deploy in the kubebuild script too" from @psav can we get this looked at too?

gburges:
> We need to see that we have the correct operator versions deployed. Also, "We should take a look at the version we deploy in the kubebuild script too" from @psav can we get this looked at too?

Looking. I would also propose to research user workload monitoring. This all could be moot if it works well for us

adamrdrew (approved):
LGTM

PR #857, opened 2023-09-11
Fix typo "that are will" -> "that will" 

juandspy:


controllers/cloud.redhat.com/version.txt
```diff
@@ -1 +1 @@
-master
\ No newline at end of file
+v0.61.0-16-gb20e0094
```
bsquizz, line 1: Need to run `make pre-push` here to avoid changing this file.

bsquizz (changes requested):

juandspy:
@bsquizz @psav not sure if the changed files are related to this fix :thinking:

bsquizz:
You're right, the doc changes looked a little larger than you probably intended. Something is probably different on your environment that was causing the docs to be generated with slight formatting differences. I re-generated the API docs.

bsquizz (approved):

psav (approved):

PR #866, opened 2023-09-21
[RHCLOUD-28457] - Fix bucket issue

psav:
The problem here is that secrets that were being "considered" as useful were able to set the hostname, so I made two tweaks to this.
1) I moved the logic for setting the hostname so that it happens during the step where we validate the hostname. It's still not great because it means that you could get one or other of two hostnames, assuming that there are multiple secrets with differing endpoints.
2) As part of this, I needed a way to send the endpoint through to that stage of the process, so I now added the ability for the endpoint to be exposed at the bucket level too. This means that buckets can have their own configuration of differing endpoints. It will be on applications to change to adopt this and I expect us to put out an ADR to deprecate the old approach very soon.

### DIFF
--- a/controllers/cloud.redhat.com/config/schema.json
+++ b/controllers/cloud.redhat.com/config/schema.json
@@ -342,6 +342,14 @@
                 "name": {
                     "description": "The actual name of the bucket being accessed.",
                     "type": "string"
+                },
+                "tls": {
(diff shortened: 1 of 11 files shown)


psav:
I also got rid of `providers.StrPtr` as we have that function in the rhc-osdk-utils now

PR #872, opened 2023-10-09
Allow empty dynaconfEnvName now that 'env' is supported

bsquizz:
Related to #855 -- 'omitempty' should have been set on this field


psav (commented):
I don't mind the extra changes in this - but I'd like to check with you to ensure that you are running a newer version than I, then I can update!

psav (approved):

PR #873, opened 2023-10-10
Added new ephem-msk mode

psav:


adamrdrew (approved):
A hearty LGTM from me. This looks great. Test coverage so I feel safe. The code itself is short, sweet, and imminently readable and I have no questions.

wcmitchell (approved):
Lots to cover here, but overall looks good! Definitely think this is something we'll have to push out and let soak to see how it behaves live.

PR #875, opened 2023-10-11
[RHCLOUD-28745] Add a new disable flag for deployments

adamrdrew:
I want the ability to disable reconciliation for a single deployment but not disable the entire clowdapp.


adamrdrew:
I am going to defer this. We're going to try it at the clowdapp level.

psav:
Can we closed this out?

PR #882, opened 2023-10-25
Update schema.json to add apiPaths

bsquizz:
When #869 went in, I did not update `schema.json`, I modified the config types file directly. Here I updated the schema and then ran `make genconfig`


adamrdrew (approved):
LGTM

PR #883, opened 2023-10-26
Add pyyaml to build scripts

adamrdrew:
We switched from ruamel to pyyaml but forgot to add it to the install script


maknop (approved):

adamrdrew:
Force merging this because there's nothing to test here

PR #884, opened 2023-10-26
Example of a refactored kuttl tests with dynamic namespace names and …

adamrdrew:
…test step templating


adamrdrew:
Doesn't work

PR #885, opened 2023-10-31
Updating kafka schema and docs to provide better details

bsquizz:
The kafka docs are a bit out of date, especially since ADR-20. Some people were confused on what values they can expect to see in the cdappconfig, so the new examples here hopefully clear it up. Also I updated the schema to provide better descriptions. Tried to use enums but that ended up resulting in the code generator creating new types for the fields instead of sticking with `*string` ... I want to avoid code breakage so just updating the `description` to indicate what values people can expect the fields to be set to by Clowder.

### DIFF
--- a/docs/appconfig/schema-definitions-dependencyendpoint-properties-apipath.md
+++ b/docs/appconfig/schema-definitions-dependencyendpoint-properties-apipath.md
@@ -1,10 +1,10 @@
-# Untitled undefined type in AppConfig Schema
+# Untitled string in AppConfig Schema
 
 ```txt
 https://cloud.redhat.com/schemas/clowder-appconfig#/definitions/DependencyEndpoint/properties/apiPath
 ```
 
-The top level api path that the app should serve from /api/<apiPath>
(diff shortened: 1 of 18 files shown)


docs/appconfig/schema-definitions-dependencyendpoint-properties-apipath.md
```diff
@@ -1,10 +1,10 @@
-# Untitled undefined type in AppConfig Schema
+# Untitled string in AppConfig Schema
 
 ```txt
 https://cloud.redhat.com/schemas/clowder-appconfig#/definitions/DependencyEndpoint/properties/apiPath
 ```
 
-The top level api path that the app should serve from /api/<apiPath>
+The top level api path that the app should serve from /api/<apiPath> (deprecated, use apiPaths)
```
bsquizz, line 7: I guess the api docs did not get updated in the prior apiPaths MR's ... anything related to 'apiPaths' changing in the docs here is due to that.

bsquizz: #888 will fix this.

psav (approved):

PR #889, opened 2023-11-10
Update Build-Deploy Script to support SC Manifest Builds

casey-williams-rh:
This PR updates the `build_deploy` script to checks if the current Git branch `$GIT_BRANCH` is `origin/security-compliance`. If the condition is true, meaning the current branch is `origin/security-compliance`, it sets the `IMAGE_TAG` variable to the value of `SECURITY_COMPLIANCE_TAG`, and will tag the built container images with in the associated generated manifest appropriately to support the creation of Security-Compliance Images.


maknop (approved):
LGTM!

PR #896, opened 2023-11-15
Fixed up multi host mode

psav:


adamrdrew (approved):
I don't have anything to add, this makes sense to me. Nothing stands out as needing correction or changing. LGTM!

PR #898, opened 2023-11-15
Fix ca cert location in web 'local' mode

bsquizz:
Although the ConfigMap is named 'openshift-service-ca.crt', the ca service cert is actually mounted at `/cdapp/certs/service-ca.crt`:

```
oc get configmap openshift-service-ca.crt -o yaml
apiVersion: v1
data:
  service-ca.crt: |+
    -----BEGIN CERTIFICATE-----
```

The 'default' web provider mode has this correct, the `local` web provider mode path needs to be changed.


tests/kuttl/test-clowder-jobs/01-pods.yaml
```diff
@@ -177,10 +177,10 @@ spec:
 ---
 apiVersion: v1
 data:
-  service-ca.crt: test-ca
+  service-ca: test-ca
```
bsquizz, line 180: Fixing this ...

psav (approved):

PR #899, opened 2023-11-28
Upgrade Keycloak

psav:
Keycloak has been upgraded elsewhere and we need to respond
* Updates the default Keycloak image
* Changes the database environment variable to be dev-mem based
* Changes the env var for initial admin credentials
* Changes the mount path for the initial realm import

### DIFF
--- a/controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
+++ b/controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
@@ -29,6 +29,9 @@ var LocalFFDeployment = rc.NewSingleResourceIdent(ProvName, "ff_deployment", &ap
 // LocalFFService is the ident referring to the local Feature Flags service object.
 var LocalFFService = rc.NewSingleResourceIdent(ProvName, "ff_service", &core.Service{})
 
+// LocalFFSecret is the ident referring to the local Feature Flags secret object.
+var LocalFFSecret = rc.NewSingleResourceIdent(ProvName, "ff_secret", &core.Secret{})
+
 // LocalFFDBDeployment is the ident referring to the local Feature Flags DB deployment object.
 var LocalFFDBDeployment = rc.NewSingleResourceIdent(ProvName, "ff_db_deployment", &apps.Deployment{})
 
@@ -41,9 +44,6 @@ var LocalFFDBPVC = rc.NewSingleResourceIdent(ProvName, "ff_db_pvc", &core.Persis
 // LocalFFDBSecret is the ident referring to the local Feature Flags DB secret object.
 var LocalFFDBSecret = rc.NewSingleResourceIdent(ProvName, "ff_db_secret", &core.Secret{})
 
-// LocalFFSecret is the ident referring to the local Feature Flags secret object.
-var LocalFFSecret = rc.NewSingleResourceIdent(ProvName, "ff_secret", &core.Secret{})
-
 type localFeatureFlagsProvider struct {
 	providers.Provider
 }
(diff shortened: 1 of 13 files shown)


controllers/cloud.redhat.com/providers/web/resources_keycloak.go
```diff
@@ -213,9 +370,19 @@ func makeKeycloak(o obj.ClowdObject, objMap providers.ObjectMap, _ bool, nodePor
...
+		Env:   envVars,
+		Args: []string{
+			"start",
+			"--import-realm",
+			"--hostname-strict",
+			"false",
+			"--http-enabled",
+			"true",
+			"--http-relative-path",
+			"/auth",
```
Victoremepunto, line 384: nitpick: it may be more readable to limit the amount of args added to the container. If you agree, can we switch the params to environment variables instead ?
 
According to the [Keycloak docs](https://www.keycloak.org/server/all-config), these should be:

|CLI param| Environment varialbe|
| -------------- | ------------------------------ |
| --hostname-strict | `KC_HOSTNAME_STRICT` |
| --http-enabled | `KC_HTTP_ENABLED` |
| --http-relative-path | `KC_HTTP_RELATIVE_PATH` |

We would still need `start --import-realm` but that'd be it. Again, take this as an optional change.

controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
```diff
@@ -41,18 +44,16 @@ var LocalFFDBPVC = rc.NewSingleResourceIdent(ProvName, "ff_db_pvc", &core.Persis
...
 }
 
 // NewLocalFeatureFlagsProvider returns a new local featureflags provider object.
 func NewLocalFeatureFlagsProvider(p *providers.Provider) (providers.ClowderProvider, error) {
 	p.Cache.AddPossibleGVKFromIdent(
-		LocalFFDBDeployment,
+		LocalFFDeployment,
 		LocalFFService,
+		LocalFFSecret,
 		LocalFFDBDeployment,
```
Victoremepunto, line 57: Is this supposed to be here ? why is it necessary to update the featureflags service ?

Victoremepunto (commented):
I have some questions about the changes made, left my comments. Great work overall!

tests/kuttl/test-ephemeral-gateway/test_creds.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 export EPHEM_BASE64=`kubectl get secret test-ephemeral-gateway-keycloak -n test-ephemeral-gateway -o json | jq '.data | map_values(@base64d)' | jq -r -j '"\(.defaultUsername):\(.defaultPassword)" | @base64'`
-export PODNAME=`kubectl get pod -l env-app=test-ephemeral-gateway-keycloak -n test-ephemeral-gateway -o json | jq -r '.items[0].metadata.name'`
```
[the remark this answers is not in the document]
psav: Because before we were running it inkeycloak which was _ok_ but not the best.....now keycloak dropped support for curl, so.......they forced me to do a better thing

controllers/cloud.redhat.com/providers/web/resources_keycloak.go
psav: I'd like to separate out the ones that need to be parametrized and the ones that are static. I agree it's a little ugly and if it got bigger maybe we can do it the other way.

controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
psav: It was missing - as I went through fixing other stuff I noticed that it was missing and added it

Victoremepunto: OK, thanks for clarifying and for fixing the bug!

controllers/cloud.redhat.com/providers/web/resources_keycloak.go
Victoremepunto: Sure. it is an implementation detail and I have no strong opinions about it. We can leave it as is.

Victoremepunto (approved):
We could set this to merge and start testing on the DEV cluster.

PR #905, opened 2023-12-06
Switch out the envoy reverse proxy for caddy

SteveHNH:
Envoy dropped support for the only version that will run on RHEL8 and no longer does security patches. This impacts secure environments since we can't use later images for envoy. Instead, we can swap out for caddy which effectively does the same thing

[RHCLOUD-29123](https://issues.redhat.com/browse/RHCLOUD-29123)


SteveHNH:
looks like this failing test is due to a `wget` to feature flags getting a 404. Not sure the solution to that just yet

psav (changes requested):
Happy to make these changes if you're too busy on return - just let me know

PR #910, opened 2023-12-20
Fixed Keycloak reliability issue and added helper for env vars

psav:


Victoremepunto (approved):
Looks good, thanks for adding the Keycloak metrics and health probes :+1:

PR #919, opened 2024-01-15
[RHCLOUD-29809] - Update caddy dependency

adamrdrew:
This updates the caddy go dependency (and unfortunately its dependencies as well)


adamrdrew (commented):
LGTM!

PR #929, opened 2024-02-05
[RHCLOUD-30810] update(db): Bump PostgreSQL images to add HLL support

Victoremepunto:


psav:
looks good to me

psav (approved):
I'm not a massive fan of moving the database definitions, I know they are used outside of the database provider, but it just feels a little counter intuitive to move them else where. I would perhaps prefer to have a separate KC one track it separately, but I'm fine with this going in and we'll see what happens.

psav:
This needs a JIRA link

Victoremepunto:
> This needs a JIRA link

Added to the title

Victoremepunto:
> I'm not a massive fan of moving the database definitions, I know they are used outside of the database provider, but it just feels a little counter intuitive to move them else where. I would perhaps prefer to have a separate KC one track it separately, but I'm fine with this going in and we'll see what happens.

This was my best attempt at trying to "fix" hardcoded configuration in the code, moving it to a common place, I don't have strong opinions on where to, "provider utils" seemed as good as any other place - I'd love to have some better configuration management so we avoid having config embedded between code

adamrdrew (approved):

PR #934, opened 2024-02-19
replaced listener.type with name

maknop:
Recent Kafka update has deprecated the `type` value to just use `name`.


adamrdrew (approved):
LGTM!

PR #936, opened 2024-02-21
Clowdapp Dependency metrics

mjholder:
Ticket: https://issues.redhat.com/browse/RHCLOUD-30912

This adds a new metric, clowder_dependency_availability, that reports the availability of each clowdapp's clowdapp dependencies + optional dependencies. 1 being UP and 0 being DOWN.
This metric gets recorded during ClowdApp reconciliation.

### DIFF
--- a/controllers/cloud.redhat.com/clowdapp_reconciliation.go
+++ b/controllers/cloud.redhat.com/clowdapp_reconciliation.go
@@ -89,13 +89,13 @@ func (r *ClowdAppReconciliation) stopMetrics() (ctrl.Result, error) {
 	return ctrl.Result{}, nil
 }
 
-func reportDependencies(ctx context.Context, pClient client.Client, o *crd.ClowdApp) (ctrl.Result, error) {
+func ReportDependencies(ctx context.Context, pClient client.Client, o *crd.ClowdApp) error {
 	appName := o.Name
 	applist := crd.ClowdAppList{}
 	dependencies := append(o.Spec.Dependencies, o.Spec.OptionalDependencies...)
 
 	if err := pClient.List(ctx, &applist, client.MatchingFields{"spec.envName": o.Spec.EnvName}); err != nil {
-		return ctrl.Result{}, err
+		return err
 	}
 
 	for _, dependency := range dependencies {
@@ -114,7 +114,7 @@ func reportDependencies(ctx context.Context, pClient client.Client, o *crd.Clowd
 		}
 	}
 
-	return ctrl.Result{}, nil
+	return nil
 }
 
 func (r *ClowdAppReconciliation) setPresentAndManagedApps() (ctrl.Result, error) {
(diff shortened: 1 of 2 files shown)


controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -89,6 +89,41 @@ func (r *ClowdAppReconciliation) stopMetrics() (ctrl.Result, error) {
 	return ctrl.Result{}, nil
 }
 
+func reportDependencies(ctx context.Context, pClient client.Client, o *crd.ClowdApp) (ctrl.Result, error) {
```
adamrdrew, line 92: This method could be slimmed down a bit and made more idiomatic.

```
func reportDependencies(ctx context.Context, pClient client.Client, o *crd.ClowdApp) (ctrl.Result, error) {
	appName := o.Name
	applist := crd.ClowdAppList{}
	dependencies := append(o.Spec.Dependencies, o.Spec.OptionalDependencies...)
	
	if err := pClient.List(ctx, &applist, client.MatchingFields{"spec.envName": o.Spec.EnvName}) ; err != nil {
		return ctrl.Result{}, err
	}

	for _, dependency := range dependencies {
		for _, app := range applist.Items {
			if app.Name != dependency {
				continue
			}

			observedReady := 0.0
			if app.Status.Ready {
				observedReady = 1.0
			} 
			
			dependencyMetrics.With(prometheus.Labels{"app": appName, "dependency": dependency}).Set(observedReady)

		}
	}

	return ctrl.Result{}, nil
}
```

We remove unneeded variable declarations, slim down the dependency array creation, do the idiomatic golang `if` error handler, and make the nested loops less nested which makes it easier to read and reason about later.

There's no logic change here, and this method doesn't need to be decomposed, so I'd approve this PR the way it is, but I wanted to present this if you wanted to refactor it to be just that little bit more idiomatic, efficient, and easy to read. That said, this is an "icing on the cake" type deal, not a deal breaker.

controllers/cloud.redhat.com/clowdapp_reconciliation.go
```diff
@@ -358,6 +393,11 @@ func (r *ClowdAppReconciliation) deletedUnusedResources() (ctrl.Result, error) {
 }
 
 func (r *ClowdAppReconciliation) setReconciliationSuccessful() (ctrl.Result, error) {
+	_, err := reportDependencies(r.ctx, r.client, r.app)
+	if err != nil {
```
adamrdrew, line 397: You could do the idiomatic go `if` error thing here if you wanted.

adamrdrew (commented):
Looks good. I'd approve as-is, but I do have some minor suggestions for improving readability, reducing waste, and being a bit more idiomatic. Sending this back to give you the decision to refactor or not. It's good as it, but there is minor room for improvement.

controllers/cloud.redhat.com/clowdapp_reconciliation.go
mjholder: Thanks for this input! I needed a second set of eyes to help clean things up. I'll test this out locally and make sure it still works.

adamrdrew (approved):
LETS GET THIS MERGED!

PR #937, opened 2024-02-22
refactor(metrics): Refactor to use built-in library function

psav:
* Replacing custom List call with a pre-built function
* Passes in ClowdEnv now


mjholder (approved):
LGTM! I like the improvements and refactoring to use existing functions.

PR #943, opened 2024-02-27
Control dependency metrics via a config or env

mjholder:
This adds a flag for enabling dependency metrics.
The env var is `ENABLE_TELEMETRY_METRICS`


maknop (approved):
LGTM!

PR #953, opened 2024-03-07
Update db provider pgsql16 image

bsquizz:
The image tag we currently specify has a permissions issue with /var/run/postgresql, updating to an image tag that has the issue resolved.


adamrdrew (approved):
LGTM

PR #957, opened 2024-03-13
Add saslMechanism and top-level securityProtocol

wcmitchell:
These configuration entries are both present in current cdappconfigs but were previously unlisted. This clarifies their existence and location for clients configuring authed Kafka clients

### DIFF
--- a/docsmd/providers/kafka.md
+++ b/docsmd/providers/kafka.md
@@ -124,7 +124,7 @@ A User auth enabled Kafka will look like this
                   "saslMechanism": "SCRAM-SHA-512"
                   "securityProtocol": "SASL_SSL"
               },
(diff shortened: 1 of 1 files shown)


docsmd/providers/kafka.md
```diff
@@ -121,8 +121,10 @@ A User auth enabled Kafka will look like this
               "sasl":{
                   "username": "kafkausername",
                   "password": "kafkapassword",
+                  "saslMechanism": "SCRAM-SHA-512"
                   "securityProtocol": "SASL_SSL"
-              }
+              },
+              "securityProtocol": "SASL_SSL
```
psav, line 127: Looks like a quote is missing here?

psav (changes requested):

PR #960, opened 2024-03-18
fix: align cloudwatch provider with documentation

avisiedo:
We observed the value for the `Type` field is an empty
string when we are using the cloudwatch provider.

We think this could fix the root cause by setting
the value for the clouderwatch logging provider, and
this value would be propagated to the configuration
injected in the final deployed pods.

https://github.com/RedHatInsights/clowder/blob/master/docsmd/providers/logging.md


avisiedo:
Closing PR as duplicated.

See: https://github.com/RedHatInsights/clowder/pull/627

PR #984, opened 2024-06-05
feat: share Downloads between iqe and selenium container

jrusz:
We want to be able to access the selenium Downloads directory from the iqe container for testing.

### DIFF
--- a/controllers/cloud.redhat.com/providers/iqe/impl.go
+++ b/controllers/cloud.redhat.com/providers/iqe/impl.go
@@ -10,6 +10,7 @@ import (
 	"github.com/go-logr/logr"
 	batchv1 "k8s.io/api/batch/v1"
 	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (approved):
lovely starter hack - we ought to standardize this for iqe so we can make it work in more circumstances (like local selenium)

we might need extra considerations if we ever go down the road of using a grid

controllers/cloud.redhat.com/providers/iqe/impl.go
```diff
@@ -129,6 +129,17 @@ func createIqeContainer(j *batchv1.Job, nn types.NamespacedName, cji *crd.ClowdJ
 		TerminationMessagePolicy: core.TerminationMessageReadFile,
 	}
 
+	// attach sel-downalods volume to access Downloads from selenim container
+	j.Spec.Template.Spec.Volumes = append(j.Spec.Template.Spec.Volumes, core.Volume{
+		Name:         "sel-downloads",
+		VolumeSource: core.VolumeSource{EmptyDir: &core.EmptyDirVolumeSource{Medium: "Memory"}},
```
bsquizz, line 135: Can we try to set a `SizeLimit` on this empty dir? Otherwise the memory-backed volume is apparently allowed to grow to 1/2 of the node's available RAM

bsquizz (changes requested):

bsquizz (dismissed):
Looks good!

bsquizz (dismissed):

PR #988, opened 2024-06-10
fix: revert condition handling to cluster-api

psav:


wcmitchell:
Changes look good, read through things and it all matches what we were talking about on the call. Only concern I have is that I see some version updates in go.mod that seem unrelated to the other changes going on. Any concerns that these version updates will introduce any additional issues that would make them worth moving to their own PR?

psav:
They shouldn't affect anything

PR #1014, opened 2024-08-05
[FEAT] Disable CJI

gburges:


psav:
Looks good - but we should have a test for it also

bsquizz (approved):
LGTM, the 'errors' assertion in kuttl makes sure that the Job doesn't appear

PR #1018, opened 2024-08-09
Fixing pr check running on new jenkins node that enforces SE Linux

maknop:


psav:
@adamrdrew is this one still needed?

PR #1024, opened 2024-08-14
Fix RHEL 8 Node PR Checks

adamrdrew:
This PR fixes PR checks running on RHEL 8 nodes. It contains 2 small fixes:

- The volume mount of the workspace in the test container has the Z flag on it to change the SE Linux context within the container to allow executing the inner test script. We aren't disabling or working around SE Linux. The context stays the same on the Jenkins node, it is simply that the volume as it shows up in the container has the context set such that we can execute a script.
- We download, verify, and install kubectl in the inner test script. The kubectl command is used in the script but not installed by the script. I am not sure how it was using it before, but seems like it was relying on a side effect of another script or part of the image build process. This fix ensures that the binaries required by the script are installed by the script, just as the script installs krew a little further down. Just like with krew we install from a tarball. I also added MD5 sum verification to ensure we are actually running the expected bits. We then place the binary in the path.


adamrdrew:
This should be ready for sqash and merge. There are 2 checks failing. One is a go linter that doesn't pertain to this change. The other is the commit lint. If you want the commits in a specific format please provide the command(s) to do so. However, if we squash and merge it shouldn't matter.

PR #1034, opened 2024-08-27
feat: Add prepush precommit hook so if we forget to make pre-push we get wa…

gburges:
…rned

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,6 +34,28 @@ pipeline {
     }
 
     stages {
+        stage('Did you run pre-push?') {
+            environment {
+                TEST_CONTAINER="clowder-ci-prepush-test-${IMAGE_TAG}-${CURR_TIME}"
+            }
+            steps {
+                withVault([configuration: configuration, vaultSecrets: secrets]) {
+                    sh '''
+                        docker login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
+                        docker build -f Dockerfile.test --build-arg BASE_IMAGE=${BASE_IMG} -t $TEST_CONTAINER .
+                        docker run -i $TEST_CONTAINER \
+                            ./githooks/prepush.sh
+                    '''
+                    }
+                }
+            
+            post {
+                always {
+                    sh 'docker rm -f $TEST_CONTAINER'
+                    }
+                }
+            }
+
         stage('Build and Push Base Image') {
             steps {
                 withVault([configuration: configuration, vaultSecrets: secrets]) {
--- a/githooks/prepush.sh
+++ b/githooks/prepush.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+function prepush() {
+make pre-push
+
+# If the make command failed, prevent the push
+if [[ -n "$(git status -s)" ]]; then
+  echo "ERROR: Please run 'make pre-push', then 'git add' any changes, and then 'git push' a new commit."
+  exit 1
+fi
+}
+
+prepush
--- a/install.sh
+++ b/install.sh
@@ -25,6 +25,13 @@ if [ $? != "0" ]; then
     pip3 install --user --upgrade crc-bonfire
 fi
 
+# Install git precommit hook
+HOOKDIR=".git/hooks"
+mkdir -p "$HOOKDIR"
+cp -r githooks/* "$HOOKDIR/"
+chmod +x "$HOOKDIR/"*
+echo "Githooks installed"
+
 # Start minikube based on specs
 minikube status 2>/dev/null >/dev/null
 if [ $? != "0" ]; then


gburges:
13:25:53 [2024-08-27T13:25:53.963Z]         --- FAIL: kuttl/harness/test-annotations-job (506.47s)

/retest

githooks/prepush
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+make pre-push
+
+status=$?
+
+# If the make command failed, prevent the push
+if [ $status -ne 0 ]; then
+  echo "Pre-push hook: make pre-push failed, aborting push."
```
bsquizz, line 8:
```suggestion
  echo "ERROR: Please run 'make pre-push', then 'git add' any changes, and then 'git push' a new commit."
```

bsquizz (changes requested):
Just a couple picky suggestions:

1. See the suggested change
2. Can we rename `prepush` to `prepush.sh` and add a newline at the end of file?

githooks/prepush
gburges: sure, though it gets copied into .git/hooks to be a pre-commit hook, and those usually are not extensioned 

The commit message, def

bsquizz (dismissed):

Victoremepunto (changes requested):
Please, can we add the lint step to the `build` Makefile goal instead of recurring to git hooks?  

See [here]()

Jenkinsfile
```diff
@@ -34,6 +34,18 @@ pipeline {
     }
 
     stages {
+        stage('Did you run pre-push?') {
+            steps {
+                sh '''
+                    make pre-push
+                    if [[ -z $(git status -s) ]]; then
```
bsquizz, line 41: This returns true if the length of `git status -s` is 0 which would be the "good" state and you would not want to fail on that. Use `-n` which will be true if the length of git status is longer than 0

```suggestion
                    if [[ -n "$(git status -s)" ]]; then
```

bsquizz (changes requested):

githooks/prepush.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+make pre-push
+
+# If the make command failed, prevent the push
+if [[ -z "$(git status -s)" ]]; then
```
bsquizz, line 5: This returns true if the length of `git status -s` is 0 which would be the "good" state and you would not want to fail on that. Use `-n` which will be true if the length of git status is longer than 0

```suggestion
                    if [[ -n "$(git status -s)" ]]; then
```

bsquizz (changes requested):

Jenkinsfile
```diff
@@ -32,8 +32,16 @@ pipeline {
         BOOTSTRAP_FUNCTIONS="${CICD_URL}/src/bootstrap.sh"
         CURR_TIME=sh(script: "date +%s", returnStdout: true).trim()
     }
-
+ 
     stages {
+        stage('Did you run pre-push?') {
+            steps {
+                withVault([configuration: configuration, vaultSecrets: secrets]) {
```
bsquizz, line 39: You can remove the `withVault` block here because this requires no creds right? So this stage will look more like https://github.com/RedHatInsights/clowder/pull/1034/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R55

Jenkinsfile
```diff
@@ -32,8 +32,16 @@ pipeline {
         BOOTSTRAP_FUNCTIONS="${CICD_URL}/src/bootstrap.sh"
         CURR_TIME=sh(script: "date +%s", returnStdout: true).trim()
     }
-
+
```
bsquizz, line 35: Extra whitespace here I think?

bsquizz (dismissed):
2 more things, then looks good to me ...

Victoremepunto (dismissed):

Jenkinsfile
```diff
@@ -34,6 +34,16 @@ pipeline {
     }
 
     stages {
+        stage('Did you run pre-push?') {
+            steps {
+                sh '''
+                    export PATH=$PATH:$(go env GOPATH)/bin
+                    ./githooks/prepush.sh
+                    cd -
```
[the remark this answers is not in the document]
gburges: the script calls pre-push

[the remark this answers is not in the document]
Victoremepunto: what I meant, is to remove the export, call make pre-push directly, and call the script from inside the Makefile, not the other way round.

That way your step looks like

```
sh '''
    make pre-push
''' 
```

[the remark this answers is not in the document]
Victoremepunto: inside the script you handle the "change dir" and "path setting" if you end up needing to, but not in the step,I found it way more straightforward to encapsulate all the shell operations in the actual shell script.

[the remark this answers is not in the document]
Victoremepunto: Also , why the "cd -" ? one should not change the working dir in scripts, that is an antipattern, please try to avoid it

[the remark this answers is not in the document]
gburges: the make  changes the directory, not the script - that 'cd -' and the path is not needed if you run it locally

putting the check in the make makes no sense when you run it manually (which you should)

doing it the other way around makes the pre-commit hook not working

PR #1036, opened 2024-08-28
feat: remove finalizer on topics in eph

maknop:
In ephemeral environments, when a namespace is cleaned up, the topics operator gets removed before handling cleanup of the topics within the namespace. These topics contain a finalizer on them and are forcing namespaces to get stuck in `deleting` status. 

Per these [docs](https://strimzi.io/blog/2023/11/02/unidirectional-topic-operator/), we need to add the following to the topics operator deployment template to ensure finalizers aren't added to the topics:
```
spec:
  template:
    spec:
      containers:
        - name: STRIMZI_USE_FINALIZERS
          value: "false"
```


psav:
I'm happy to approve this but we need to watch it carefully and ensure that the test cluster does the right thing when no config is set, before changing the operation flag.

PR #1043, opened 2024-09-03
setting default value for strimzi finalizer to false

maknop:
By default we want strimzi finalizers to be enabled for Clowder.


gburges (approved):

gburges (approved):

psav:
The naming around this is very confusing - USE_STRIMZI_FINALIZSER - setting to false, ....enables them?

PR #1044, opened 2024-09-04
change strimzi finalizer variable name

maknop:
`STRIMZI_USE_FINALIZER` is confusing for the variable name. We are updating to `DISABLE_STRIMZI_FINALIZER`.


deploy.yml
```diff
@@ -8732,6 +8732,6 @@ parameters:
   value: quay.io/cloudservices/minio:RELEASE.2020-11-19T23-48-16Z-amd64
 - name: FEATUREFLAGS_IMAGE
   value: quay.io/cloudservices/unleash-docker:5.6.9
-- name: STRIMZI_USE_FINALIZERS
+- name: DISABLE_STRIMZI_FINALIZER
```
psav, line 8735: This should be set to "false"

psav (changes requested):

PR #1056, opened 2024-09-25
Konfluxing

maknop:
Cachi2 will allow us to cache Golang dependencies during the image build process. The impact here is that the time it takes to get through the build process will be reduced enough that we can remove the base image process we are currently using to cut the time down.

Unit tests and E2E tests will also be added to the pipeline.

### DIFF
--- a/ci/konflux_minikube_e2e_tests.sh
+++ b/ci/konflux_minikube_e2e_tests.sh
@@ -1,0 +1,117 @@
+#!/bin/bash
+
+set -exv
+
+mkdir -p /container_workspace/bin
+cp /opt/app-root/src/go/bin/* /container_workspace/bin
+
+export KUBEBUILDER_ASSETS=/container_workspace/testbin/bin
+
+(
+  set -x; cd "$(mktemp -d)" &&
+  OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
+  ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" &&
+  KREW="krew-${OS}_${ARCH}" &&
+  curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREW}.tar.gz" &&
+  tar zxvf "${KREW}.tar.gz" &&
+  ./"${KREW}" install krew
+)
+
+source build/template_check.sh
+
+export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
+export PATH="/bins:$PATH"
+
+chmod 600 minikube-ssh-ident
+
+ssh -o StrictHostKeyChecking=no $MINIKUBE_USER@$MINIKUBE_HOST -i minikube-ssh-ident "minikube delete"
+ssh -o StrictHostKeyChecking=no $MINIKUBE_USER@$MINIKUBE_HOST -i minikube-ssh-ident "minikube start --cpus 6 --disk-size 10GB --memory 16000MB --kubernetes-version=1.26.3 --addons=metrics-server --disable-optimizations"
+
+export MINIKUBE_IP=`ssh -o StrictHostKeyChecking=no $MINIKUBE_USER@$MINIKUBE_HOST -i minikube-ssh-ident "minikube ip"`
+
+scp -o StrictHostKeyChecking=no -i minikube-ssh-ident $MINIKUBE_USER@$MINIKUBE_HOST:$MINIKUBE_ROOTDIR/.minikube/profiles/minikube/client.key ./
+scp -i minikube-ssh-ident $MINIKUBE_USER@$MINIKUBE_HOST:$MINIKUBE_ROOTDIR/.minikube/profiles/minikube/client.crt ./
+scp -i minikube-ssh-ident $MINIKUBE_USER@$MINIKUBE_HOST:$MINIKUBE_ROOTDIR/.minikube/ca.crt ./
+
+ssh -o ExitOnForwardFailure=yes -f -N -L 127.0.0.1:8444:$MINIKUBE_IP:8443 -i minikube-ssh-ident $MINIKUBE_USER@$MINIKUBE_HOST
+
+cat > kube-config <<- EOM
+apiVersion: v1
+clusters:
+- cluster:
+    certificate-authority: $PWD/ca.crt
+    server: https://127.0.0.1:8444
+  name: 127-0-0-1:8444
+contexts:
+- context:
+    cluster: 127-0-0-1:8444
+    user: remote-minikube
+  name: remote-minikube
+users:
+- name: remote-minikube
+  user:
+    client-certificate: $PWD/client.crt
+    client-key: $PWD/client.key
+current-context: remote-minikube
+kind: Config
+preferences: {}
+EOM
+
+export PATH="$KUBEBUILDER_ASSETS:$PATH"
+export PATH="/root/go/bin:$PATH"
+
+export KUBECONFIG=$PWD/kube-config
+export KUBECTL_CMD="kubectl "
+$KUBECTL_CMD config use-context remote-minikube
+$KUBECTL_CMD get pods --all-namespaces=true
+
+source build/kube_setup.sh
+
+export IMAGE_TAG=`git rev-parse --short=8 HEAD`
+
+$KUBECTL_CMD create namespace clowder-system
+
+mkdir artifacts
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,6 +1,46 @@
...
+COPY Makefile Makefile
+
+# RUN make controller-gen kustomize
+
+# RUN GO111MODULE=on go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0 \
+#     && GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v4@v4.5.2
+
+COPY hack/boilerplate.go.txt hack/boilerplate.go.txt
+
+COPY main.go main.go
```
psav, line 29: Why are you copying them all in, deleting them, and then copying in again?

Victoremepunto (changes requested):
After reviewing the Tekton tasks here, I can't understand why the on-push one is running on every commit push, and why the on-pull-request one is actually not running at all. 

I would suggest if possible, to, on a separate PR, simply onboard Clowder as is, and merge those tasks directly (even if they don't build successfully) - and then on this PR, we can rebase on top of it, so that we can see the changes you are adding on this PR (as opposed to, the full Tekton task definition).

.tekton/clowder-pull-request.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
     pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
-      == "master"
```
Victoremepunto, line 11: this needs to be `master` in order to be triggered, as it defines the target branch against the PR is going to be merged

Victoremepunto:
@maknop  I've reviewed your PR, thank you very much for doing the actual Konflux onboarding first, as a separate PR, and rebased this PR on top, it is now easier to see what you are adding on top in this PR. 

However, I've noticed the pipeline you used was incorrect (it was the `docker-build` one, and I can see you're using trusted artifacts, hence your chosen pipeline should've been the `docker-build-oci-ta` one instead.

I've fixed that with https://github.com/RedHatInsights/clowder/pull/1074, please rebase your changes once again, we should see even less LOC changed from your PR. My early comment about the target branch for the trigger is still valid, please do not change it, it needs to be set to master 

CC @psav

.tekton/clowder-pull-request.yaml
```diff
@@ -46,7 +48,7 @@ spec:
         - name: name
           value: show-sbom
         - name: bundle
-          value: quay.io/konflux-ci/tekton-catalog/task-show-sbom:0.1@sha256:52f8b96b96ce4203d4b74d850a85f963125bf8eef0683ea5acdd80818d335a28
+          value: quay.io/konflux-ci/tekton-catalog/task-show-sbom:0.1@sha256:9bfc6b99ef038800fe131d7b45ff3cd4da3a415dd536f7c657b3527b01c4a13b
```
Victoremepunto, line 51: rollback this please, this is an older hash

https://quay.io/repository/konflux-ci/tekton-catalog/task-show-sbom?tab=tags

.tekton/clowder-pull-request.yaml
```diff
@@ -184,7 +189,7 @@ spec:
         - name: name
           value: prefetch-dependencies-oci-ta
         - name: bundle
-          value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta:0.1@sha256:3c11f5de6a0281bf93857f0c85bbbdfeda4cc118337da273fef0c138bda5eebb
+          value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta:0.1@sha256:621b13ab4a01a366a2b1d8403cf06b2b7418afd926d13678c4432858514407d3
```
Victoremepunto, line 192: same as before, this is older than upstream, please don't apply this change

https://quay.io/repository/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta?tab=tags&tag=0.1

.tekton/clowder-pull-request.yaml
```diff
@@ -225,7 +230,7 @@ spec:
         - name: name
           value: buildah-oci-ta
         - name: bundle
```
Victoremepunto, line 232: same...

https://quay.io/repository/konflux-ci/tekton-catalog/task-buildah-oci-ta?tab=tags&tag=0.2

Victoremepunto (changes requested):
I've noticed the tasks are using older SHAs than what's upstream, please revert the ones that are using older than what's upstream

Victoremepunto (commented):
Good job, thanks for adding the changes to the Tekton tasks

@psav this still needs your approval, I have only reviewed the Tekton bits, they look good, we'll consider adding a follow up to add an ITS to run the EC in PRs

Makefile
```diff
@@ -3,8 +3,8 @@
 #
 # For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
 # cloud.redhat.com/clowder-bundle:$VERSION and cloud.redhat.com/clowder-catalog:$VERSION.
-IMAGE_TAG_BASE ?= quay.io/cloudservices/clowder
-CLOWDER_BUILD_TAG ?= $(shell git rev-parse --short=8 HEAD)
+IMAGE_TAG_BASE ?= quay.io/redhat-user-workloads/hcm-eng-prod-tenant/clowder/clowder
```
psav, line 6: There are changes that need to be made below - we should consider removing some older `make targets` - we are deprecating the base image here, and that is used as part of the build below, we should also consider completely removing the base dockerfile, as it's now useless.

ci/minikube_e2e_tests.sh
psav, line 1: Can we delete these two older files please as they are no longer used in the process at all.

deps/update_e2e_deps.sh
psav, line 1: This is going to be superseded by some changes that are in a future PR - but OK to leave for now

ci/konflux_minikube_e2e_tests.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+
+set -exv
+
+mkdir -p /container_workspace/bin
+
+export KUBEBUILDER_ASSETS=/container_workspace/testbin/bin
+
+curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
psav, line 9: How is this working when hermetic builds are set to true? We should clarify with konflux team because, either we are not enforcing hermetic builds and we think we are, or there is a glaring bug/issue with the hermetic build system. Either way the future of this needs to have kubectl as a trusted artifact.

ci/konflux_minikube_e2e_tests.sh
```diff
@@ -0,0 +1,128 @@
...
+chmod +x kubectl
+mv kubectl /container_workspace/bin
+export PATH="/container_workspace/bin:$PATH"
+
+(
+  cd "$(mktemp -d)" &&
+  OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
+  ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" &&
+  KREW="krew-${OS}_${ARCH}" &&
+  curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREW}.tar.gz" &&
```
psav, line 21: Same comment above, but for krew

ci/konflux_minikube_e2e_tests.sh
```diff
@@ -0,0 +1,128 @@
...
+
+$KUBECTL_CMD create namespace clowder-system
+
+mkdir artifacts
+
+make release
+
+cat manifest.yaml > artifacts/manifest.yaml
+
+sed -i "s/clowder:latest/clowder:on-pr-$IMAGE_TAG/g" manifest.yaml
```
psav, line 91: This shouldn't work? because `on-pr` is removed. Can we investigate how this is working? Maybe this line is just silently doing nothing in the script now? We need to figure this out.

psav (changes requested):
@Victoremepunto and I went through this today for a final review - Regrettably there are some issues that we need to address/fix/explain before this can be merged, however these should be relatively quick to remediate with the right guidance from the team/konflux.

psav (changes requested):

PR #1067, opened 2024-10-23
(feat):Add configurable configmap config to otelcollector config

gburges:


adamrdrew:
```
08:29:27 [2024-10-24T12:29:27.172Z] === NAME  kuttl/harness/test-sidecars
08:29:27 [2024-10-24T12:29:27.172Z]     logger.go:42: 12:29:25 | test-sidecars/1-pods | test step failed 1-pods
08:29:27 [2024-10-24T12:29:27.172Z]     case.go:399: failed in step 1-pods
08:29:27 [2024-10-24T12:29:27.172Z]     case.go:401: deployments.apps "puptoo-processor" not found
08:29:27 [2024-10-24T12:29:27.172Z]     case.go:401: cronjobs.batch "puptoo-cron" not found
```
Now that is weird. There's no assertion failure, it just straight up didn't create the requested resources? I *think* we'd need to see what was going on in k8s to get a handle on this.

adamrdrew:
I'm running re-test so we can count out it just being a fluke. Kuttl tests can fail sometimes just because reasons. If it fails in the same way though you'll need to run locally and debug.

controllers/cloud.redhat.com/providers/sidecar/default.go
```diff
@@ -164,5 +176,9 @@ func getOtelCollector() *core.Container {
 			"memory": resource.MustParse("1024Mi"),
 		},
 	}
+	cont.VolumeMounts = []core.VolumeMount{{
+		Name:      fmt.Sprintf("%s-otel-config", appName),
+		MountPath: "/etc/otelcol/config",
```
dkliban, line 181: This path should be `/etc/otelcol/config.yaml`.

gburges: thx, updated :)

dkliban:
@gburges we would like to make use of this feature. is there any way i can help in getting this work finished?

adamrdrew:
LGTM

PR #1085, opened 2024-11-26
added versioning step for tekton pipelines

maknop:


adamrdrew:
I have a couple of concerns with this. They may actually be just misunderstandings, but I want to ASK

* Are we allowed to install deps from dnf at build time when using konflux? I thought everything had to be there already.
* We set the clowder version as the last tag, but we don't update the tags anywhere? so would this just produce the same version over and over again?

maknop:
We decided to find a new way to handle this since it wasn't working perfectly in our Jenkins builds anyway. Going to close this out.

PR #1087, opened 2024-11-27
Upgrade go version

psav:
Upgrade go version


.tekton/clowder-pull-request.yaml
```diff
@@ -244,6 +244,97 @@ spec:
...
+        - name: ociStorage
+          value: $(params.output-image).make-template-kustomize
+        - name: IMAGE_EXPIRES_AFTER
+          value: $(params.image-expires-after)
+      runAfter:
+        - prefetch-dependencies
+      taskSpec:
+        results:
+        - name: TEMPLATE_ARTIFACT
+          type: array
```
ernesgonzalez33, line 261: Why is this an array?

.tekton/clowder-pull-request.yaml
```diff
@@ -244,6 +244,97 @@ spec:
...
+      runAfter:
+        - make-template
+      taskSpec:
+        params:
+          - description: The Trusted Artifact URI pointing to the artifact with the application source code.
+            name: SOURCE_ARTIFACT
+            type: string
+          - description: The Trusted Artifact URI pointing to the artifact with the template.
+            name: TEMPLATE_ARTIFACT
+            type: string
```
ernesgonzalez33, line 311: And here a string?

ernesgonzalez33 (changes requested):

.github/workflows/lint.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/setup-go@v4
         name: Set up Go 1.x
         with:
-          go-version: "1.21.9"
+          go-version: "1.22.7"
```
maskarb, line 25: You could use `go-version-file: 'path/to/go.mod'` instead. It's one less thing to remember to update in a future go version update.

We use it [here](https://github.com/project-koku/koku-metrics-operator/blob/main/.github/workflows/ci.yaml#L20-L23).

psav: Thanks for that

gburges (approved):

PR #1089, opened 2024-11-28
docs(examples): fix yaml examples and use pg 16

vkrizan:


bsquizz (approved):

bsquizz:
Rebased so that the commits will be signed/verified

PR #1101, opened 2024-12-09
feat(providers): single database mode

vkrizan:
Creates a new database provider mode named `single`.
The aim of the mode is to create a single database instance and share it
between the apps by separating them by Posgres schemas.
Each app gets a schema created by the name of the requested database.
Schemas are created under environment's database, with the same name as
the ClowdEnv.

Adds the following refactorings:
* `provutils.DefaultPGPort` to set port constant
* `provutils.DefaultPGAdminUsername` to set default PG admin user
* `provutils.PGAdminConnectionStr` that builds sanitized connection string to PG from a `DatabaseConfig` using admin creds
* `provutils.ReadDbConfigFromSecret` to reuse read of `DatabaseConfig`s from secrets, optionally using a cache

RHINENG-14526

### DIFF
--- a/tests/kuttl/test-multi-app-single-db/00-install.yaml
+++ b/tests/kuttl/test-multi-app-single-db/00-install.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: test-single-db
+spec:
+  finalizers:
+  - kubernetes
--- a/tests/kuttl/test-multi-app-single-db/01-assert.yaml
+++ b/tests/kuttl/test-multi-app-single-db/01-assert.yaml
@@ -1,0 +1,29 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test-single-db-db-single
+  namespace: test-single-db
+spec:
+  template:
+    spec:
+      containers:
+        - image: quay.io/cloudservices/postgresql-rds:16-759c25d
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: test-single-db-db-single
+  namespace: test-single-db
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: app-a-db
+  namespace: test-single-db
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: app-b-db
+  namespace: test-single-db
--- a/tests/kuttl/test-multi-app-single-db/01-pods.yaml
+++ b/tests/kuttl/test-multi-app-single-db/01-pods.yaml
@@ -1,0 +1,79 @@
+---
+apiVersion: cloud.redhat.com/v1alpha1
+kind: ClowdEnvironment
+metadata:
+  name: test-single-db
+  namespace: test-single-db
+spec:
+  targetNamespace: test-single-db
+  providers:
+    web:
+      port: 8000
+      mode: operator
+    metrics:
+      port: 9000
+      mode: operator
+      path: "/metrics"
+    kafka:
+      mode: none
+    db:
+      mode: single
+    logging:
+      mode: none
+    objectStore:
+      mode: none
+    inMemoryDb:
+      mode: none
+    featureFlags:
+      mode: none
+  resourceDefaults:
+    limits:
+      cpu: 400m
+      memory: 1024Mi
+    requests:
+      cpu: 30m
+      memory: 512Mi
+---
+apiVersion: cloud.redhat.com/v1alpha1
+kind: ClowdApp
+metadata:
+  name: app-a
+  namespace: test-single-db
+spec:
+  envName: test-single-db
+  deployments:
+  - name: processor
+    podSpec:
+      image: quay.io/psav/clowder-hello
+  database:
+    name: app-a
+---
+apiVersion: cloud.redhat.com/v1alpha1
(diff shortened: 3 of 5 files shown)


vkrizan:
@psav I've opted for "single" as the name of the db mode. Suggestions and reviews are welcome. Thank you.

controllers/cloud.redhat.com/providers/database/single.go
```diff
@@ -0,0 +1,302 @@
...
+func (db *singleDbProvider) createOrReadDbConfig(obj obj.ClowdObject, nn types.NamespacedName, username string, setAdmin bool) (cfg *config.DatabaseConfig, err error) {
+	cfg = &config.DatabaseConfig{}
+	password, err := utils.RandPassword(16, provutils.RCharSet)
+	if err != nil {
+		return cfg, errors.Wrap("password generate failed", err)
+	}
+
+	var pgPassword string
+	if setAdmin {
+		pgPassword, err = utils.RandPassword(16, provutils.RCharSet)
```
vkrizan, line 218: My initial thought was to disable admin user for apps, however some apps might still require this to do some setups (i.e. create additional users). @psav any suggestions how we should approach this situation? I guess, one of the option is to update those services needing such privileges, or look at the privileges case-by-case.

vkrizan:
I'm opening this up from a draft state for review. There still might be some small issues in certain conditions. I prefer to get this out sooner to do a bunch of testing on a test cluster. Thank you.

controllers/cloud.redhat.com/providers/database/single.go
psav: It's tricky because ephemeral usage often relies on them needing to do more stuff than usual, but in prod they will just get the credentials which are appropriate

psav:
> I'm opening this up from a draft state for review. There still might be some small issues in certain conditions. I prefer to get this out sooner to do a bunch of testing on a test cluster. Thank you.

We should talk about the expectation here and I was thinking we can deploy this for testing in the cluster - without merging it and only merge when we are ready - allows us to not goof up the cluster too much. @bsquizz @wcmitchell

docs/providers/database.md
```diff
@@ -64,6 +64,17 @@ ClowdEnv Config options available:
...
+#### single
+
+(*EXPERIMENTAL*)
+
+In the single mode, the the **Database Provider** will provision a single node
+PostgreSQL instance, of latest supported version, for the whole ClowdEnvironment.
+Each ClowdApp requesting a database would get a separate user and
+a [PostgreSQL schema](https://www.postgresql.org/docs/current/ddl-schemas.html)
+with the same name as the user. The user has only access to create objects
+only within it's schema. Admin credentials are (currently) not provided.
```
PaulWay, line 76: "it's" (abbreviation) -> "its" (possessive)  (nitpick :-) )

PaulWay (commented):
I don't know Go or Clowder well enough to know whether it works or not but just contributing a few suggestions for the documentation.

docs/providers/database.md
vkrizan: fixed

config/crd/bases/cloud.redhat.com_clowdenvironments.yaml
```diff
@@ -81,12 +81,15 @@ spec:
...
-                          Provider. Valid options are: (*_app-interface_*) where the
-                          provider will pass through database credentials found in
-                          the secret defined by the database name in the ClowdApp,
-                          and (*_local_*) where the provider will spin up a local
-                          instance of the database.'
+                          Provider. Valid options are: (*_app-interface_*) where
+                          the provider will pass through database credentials found
+                          in the secret defined by the database name in the
+                          ClowdApp, (*_local_*) where the provider will spin up a
+                          local instance of the database, and (*_single_*) that is
```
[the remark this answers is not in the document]
vkrizan: hopefully fixed

bsquizz (changes requested):
The approach looks good and I like the consolidation of common operations into provutils. Aside from my nitpicks on the DB mode descriptions, I think the biggest thing this needs is a kuttl test to ensure this works as expected. Let me know if you need help to get started on creating one.

controllers/cloud.redhat.com/providers/database/single.go
vkrizan: Yeah. I want to test those apps to see how they would work with this limited access... AFAIK Advisor, has dropped or is dropping the HLL extension.

config/crd/bases/cloud.redhat.com_clowdenvironments.yaml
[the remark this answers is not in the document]
vkrizan: I've called it again, and there is no visible change.

I've used this as the `no-update` target is git setup dependent:
```
make manifests generate genconfig build-template api-docs
```

controllers/cloud.redhat.com/providers/database/single.go
PaulWay: We plan to remove use of the HLL extension next week (17th Feb 2025).

PaulWay: That's the only extension I'm aware of in Advisor.

PR #1115, opened 2025-01-03
Deploy unleash-edge alongside local featureFlags

ezr-ondrej:
Deploys unleash-edge with local feature flags.

### DIFF
--- a/controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
+++ b/controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
@@ -306,15 +306,83 @@ func makeLocalFeatureFlags(o obj.ClowdObject, objMap providers.ObjectMap, _ bool
 		},
 	}
 
-	dd.Spec.Template.Spec.Containers = []core.Container{c}
+	portEdge := int32(3063)
+
+	envVarsEdge := []core.EnvVar{
+		{
+			// communication with the main featureflags service on localhost
+			Name:  "UPSTREAM_URL",
+			Value: fmt.Sprintf("http://127.0.0.1:%d", port),
+		},
+	}
+	envVarsEdge = provutils.AppendEnvVarsFromSecret(envVarsEdge, nn.Name,
+		provutils.NewSecretEnvVar("TOKENS", "clientAccessToken"))
+
+	portsEdge := []core.ContainerPort{{
+		Name:          "service",
+		ContainerPort: portEdge,
+		Protocol:      "TCP",
+	}}
+
+	readinessProbeEdge := core.Probe{
+		ProbeHandler: core.ProbeHandler{
+			Exec: &core.ExecAction{
+				Command: []string{"/unleash-edge", "ready"},
+			},
+		},
+		InitialDelaySeconds: 1,
+		TimeoutSeconds:      5,
+		PeriodSeconds:       30,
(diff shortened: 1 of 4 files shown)


psav:
Can you give me a little more about if this is necessary, or optional, and the use-cases? Perhaps updating the featureflags provider with some docs?

ezr-ondrej:
/retest
pending for 2 days, it might not finish I assume xD

bsquizz:
There's currently some issues with the pipeline that we are working on

ezr-ondrej:
It's VERY green :)
@bsquizz anything you're missing here?
I'd like to add Ingress, but if you'd agree, I'd do it as a follow-up ;)

psav (dismissed):
I'm happy with this, looking good to me - my only concern is that when we start using it, it will temporarily increase the load on ephemeral as everyone will have both edge and the old proxy and we can't remove the old one until the ingress is there -  @gburges something to be aware of.

ezr-ondrej:
I promise to do my best to contribute the ingres as well. I'm very commited to move to the unleash-edge.
It might unblock many frontends running IQE tests on Konflux, so it's high on my prio list :stuck_out_tongue:

tests/kuttl/test-ff-local/test_feature_flags.sh
```diff
@@ -59,3 +68,12 @@ if [ 'true' != "$(get_request "$CLIENT_TOKEN" "/api/client/features/$FEATURE_TOG
...
     exit 1
 fi
+
+# Unleash seems to refresh the cache every 5 seconds, didn't find a way to force it
+# see https://github.com/Unleash/unleash-edge/blob/12cf9e3f87099d3c0dce1884bcd305604c1e68ff/server/src/http/refresher/feature_refresher.rs#L443
+echo "Waiting for edge to sync"
+sleep 6
+
+if ! get_request_edge "$CLIENT_TOKEN" "/api/client/features/$FEATURE_TOGGLE_NAME"; then
+    echo "Feature toggle '$FEATURE_TOGGLE_NAME' should be available through edge"
```
bsquizz, line 78: This would be a test failure if the toggle is not available, correct? If so, I think you need an `exit 1` within this block.

bsquizz (changes requested):
Looks good, very similar to the existing logic for this provider. I spotted 2 small things.

controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
```diff
@@ -319,3 +338,106 @@ func makeLocalFeatureFlags(o obj.ClowdObject, objMap providers.ObjectMap, _ bool
...
+	labeler(dd)
+
+	replicas := int32(1)
+
+	dd.Spec.Replicas = &replicas
+	dd.Spec.Selector = &metav1.LabelSelector{MatchLabels: labels}
+
+	dd.Spec.Template.ObjectMeta.Labels = labels
+
+	portEdge := int32(3063)
```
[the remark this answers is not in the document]
ezr-ondrej: Done :)

tests/kuttl/test-ff-local/test_feature_flags.sh
ezr-ondrej: Nice catch, I've addressed it in the place I copy pasted from too xD

PR #1122, opened 2025-01-13
Update Clowder to reconcile on changes to non-app secrets/configmaps

psav:
* Add hashCache to the Environment reconciler for usage later
* Remove all Secrets/ConfigMaps from Environment at start of reconciliation cycle
* Embue HashObject with the always flag
* Change updateHashCacheForConfigMapAndSecret to try to read in the sec/config and return true if always set
* Slight concern of erroring if not in the cache, but
* Make app_interface add the secret to the cache at the start of reconciliation so it should be there, and also link the objects.


psav:
These changes have been incorporated into #1124 so I'll close this PR

PR #1132, opened 2025-01-23
Update buildah task

bsquizz:


maknop (approved):

bsquizz:
We don't need the full kuttl test run on this one, just needed to make sure the build task looked good. Merging early.

PR #1137, opened 2025-01-27
fix: Add support for multiple APIPaths in caddygateway

ezr-ondrej:


psav:
Good fix - is it possible to get a test in here too?

ezr-ondrej:
I've refactored the method a bit to separate the logic for preparation of the paths.
That allowed me to cover it by unit test. It feels good enough to me, WDYT @psav ?

psav:
@wcmitchell a good one to review ;)

PR #1152, opened 2025-02-20
Add override for inmemoryDB images for ephemeral testing ov valkey

psav:
* Adds test for inmemoryDB image switch


adamrdrew (approved):
I've got no feedback. The change is small, you've got a test around it, I don't see any problems in the code.

PR #1159, opened 2025-02-24
Fixes deps to be more go-like

psav:


psav:
Doesn't work because controller-runtime/kustomize want different versions of core dependencies

PR #1164, opened 2025-02-26
Fix redis valkey transition

psav:
* Redis was special in that it functions at the app level and the env level
* Most cached components will only ever be deployed at the env level and contain a function call to check for the presence of the env and bomb out if the o object is not an env
* In the case of Redis, it needed to be called at app  level with the problem being that the functions for getting default images required access to the env spec
* One option was to try to smuggle the env with every app, but this seemed counter intuitive
* The next option was to try to grab the env from the app during the makeLocalRedis() function
* This fails because it's necessary to also pass in context and a client object
* A third option was to make the functions that grab the default images hang off of the ClowdApp/ClowdEnvironment, but the problem here is that the ClowdApp would need the ClowdEnvironment, meaning it would also need a client/context.
* Final option landed upon is to always pass the Env to the CachedMakeComponent() call, as well as a ClowdObject, which can be a ClowdEnvironment, or a ClowdApp as before
* This does yield some redundancy because the Env is passed in twice in somecases

So....

* Too many args being passed in
* Several of those were objects that already hung off of the provider object
* So now we just pass the provider object in to allow it to provide the Cache, the NodePort option and the Env

### DIFF
--- a/controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
+++ b/controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
@@ -91,7 +91,7 @@ func (ff *localFeatureFlagsProvider) EnvProvide() error {
 		LocalFFService,
 	}
 
-	if err := providers.CachedMakeComponent(ff.Cache, objList, ff.Env, "featureflags", makeLocalFeatureFlags, false, ff.Env.IsNodePort()); err != nil {
+	if err := providers.CachedMakeComponent(ff.Cache, objList, ff.Env, ff.Env, "featureflags", makeLocalFeatureFlags, false, ff.Env.IsNodePort()); err != nil {
 		return err
 	}
 
@@ -100,7 +100,7 @@ func (ff *localFeatureFlagsProvider) EnvProvide() error {
 		LocalFFEdgeService,
 	}
 
-	if err := providers.CachedMakeComponent(ff.Cache, objList2, ff.Env, "featureflags-edge", makeLocalFeatureFlagsEdge, false, ff.Env.IsNodePort()); err != nil {
+	if err := providers.CachedMakeComponent(ff.Cache, objList2, ff.Env, ff.Env, "featureflags-edge", makeLocalFeatureFlagsEdge, false, ff.Env.IsNodePort()); err != nil {
 		return err
 	}
 
@@ -229,7 +229,7 @@ func (ff *localFeatureFlagsProvider) Provide(_ *crd.ClowdApp) error {
 	return nil
 }
 
-func makeLocalFeatureFlags(o obj.ClowdObject, objMap providers.ObjectMap, _ bool, nodePort bool) error {
+func makeLocalFeatureFlags(_ *crd.ClowdEnvironment, o obj.ClowdObject, objMap providers.ObjectMap, _ bool, nodePort bool) error {
 	nn := providers.GetNamespacedName(o, "featureflags")
 
 	dd := objMap[LocalFFDeployment].(*apps.Deployment)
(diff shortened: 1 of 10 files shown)


controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
```diff
@@ -91,7 +91,7 @@ func (ff *localFeatureFlagsProvider) EnvProvide() error {
 		LocalFFService,
 	}
 
-	if err := providers.CachedMakeComponent(ff.Cache, objList, ff.Env, "featureflags", makeLocalFeatureFlags, false, ff.Env.IsNodePort()); err != nil {
+	if err := providers.CachedMakeComponent(ff, objList, ff.Env, "featureflags", makeLocalFeatureFlags, false); err != nil {
```
maknop, line 94: I know you mentioned the redundancy here in your description but since you're passing in the whole object as the first parameter, do you still need to pass in `ff.Env` as another param? I do see this in various places.

controllers/cloud.redhat.com/providers/inmemorydb/redis_test.go
```diff
@@ -37,7 +37,7 @@ func TestLocalRedis(t *testing.T) {
 		RedisDeployment: &dd,
 		RedisService:    &svc,
 	}
-	_ = makeLocalRedis(&env, objMap, true, false)
+	_ = makeLocalRedis(&env, &env, objMap, true, false)
```
adamrdrew, line 40: We pass in env twice, but the second param is supposed to be a ClowdObject. I'm sure I just don't understand but was this on purpose?

controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
```diff
@@ -229,7 +229,7 @@ func (ff *localFeatureFlagsProvider) Provide(_ *crd.ClowdApp) error {
 	return nil
 }
 
-func makeLocalFeatureFlags(o obj.ClowdObject, objMap providers.ObjectMap, _ bool, nodePort bool) error {
+func makeLocalFeatureFlags(_ *crd.ClowdEnvironment, o obj.ClowdObject, objMap providers.ObjectMap, _ bool, nodePort bool) error {
```
adamrdrew, line 232: Why did we add an unused param for ClowdEnvironment? I don't understand.

controllers/cloud.redhat.com/providers/inmemorydb/redis_test.go
psav: Yes this was on purpose - the first time is the "I'll always give you an env", the second is - and "the redis you create will be for this env" - as opposed to ..... for this particular "app"

psav: A ClowdObject is either a ClowdApp or a ClowdEnvironment - interface FTW

adamrdrew: Gotcha thanks!

controllers/cloud.redhat.com/providers/providers.go
```diff
@@ -184,13 +184,13 @@ type ObjectMap map[rc.ResourceIdent]client.Object
...
 // deployment and PVC, based on the makeFn that is passed in.
-func CachedMakeComponent(cache *rc.ObjectCache, objList []rc.ResourceIdent, o obj.ClowdObject, suffix string, fn makeFnCache, usePVC bool, nodePort bool) error {
+func CachedMakeComponent(prov RootProvider, objList []rc.ResourceIdent, o obj.ClowdObject, suffix string, fn makeFnCache, usePVC bool) error {
 	nn := GetNamespacedName(o, suffix)
 
 	makeFnMap := make(map[rc.ResourceIdent]client.Object)
 
 	for _, v := range objList {
-		obj, err := createResource(cache, v, nn)
+		obj, err := createResource(prov.GetCache(), v, nn)
```
adamrdrew, line 193: Do we have any concern that an incorrectly implemented provider might not have this method? Like, do we need to check for it first or something before calling it? I was worried about a nil pointer.

adamrdrew (approved):
I have a few questions that I'm sure are no blocker to merging this.

controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
psav: This is to enable it to be used in one particular provider only (redis)

maknop (approved):

controllers/cloud.redhat.com/providers/featureflags/localfeatureflags.go
psav: And to satisfy the interface ;)

PR #1177, opened 2025-03-12
Make proxy image configurable and update default image

adamrdrew:
This adds the following

* CaddyProxy Image to the ClowdEnvironment spec
* CaddyProxy Image to the Clowder config
* New getter to get the correct image version from either spec, clowder config, or default
* Add getter to where we construct the proxy resource
* Bumped to the latest caddy-ubi


adamrdrew:
Something worth considering with this. Our default image will now come from a Konflux repo rather than cloudservices. That means a pull secret is required. Of course it can be overridden in config, but our default wont work anywhere but our clusters. Maybe not great?

psav:
legit failure in the image in the tests

adamrdrew:
Marking this as draft until I can figure out what's wrong. `make pre-push` is bombing out. I don't know what's up.

wcmitchell (approved):
Tests passing, everything looks reasonable overall. Willing to give this a merge and see what happens

PR #1180, opened 2025-03-14
Add ingress for local unleash edge

ezr-ondrej:
Adds ingress for local unleash-edge for it to be available from outside of the cluster.


ezr-ondrej:
Ah, the test cluster doesn't have ingress 🤔 I'm not sure how to test that it's properly exposed, but it works locally :)

psav (approved):

psav (approved):

PR #1189, opened 2025-03-18
Adds otel sidecar as an initContainer with the restartPolicy to always

psav:
* Previously otel was deployed as a normal container but this had a negative impact when pods were restarted
* As a result we have moved this to an init container to ensure that it is running before the main service and shutdown AFTER


wcmitchell (approved):
Looking good and seeing passing tests, a-ok with me! Went over the code and functionality with @psav earlier.

PR #1204, opened 2025-04-08
Search for exact match on object store bucket name

bsquizz:
We ran into an issue where an app team had two bucket configurations for s3 object storage, similar to this:

`my-secret-temp`
`my-secret`

The requested bucket name was `my-secret`, but they were provided the credentials of `my-secret-temp`.

I'm not aware of a need to use `strings.Prefix` when searching for the bucket name, so we can instead just do a simple `==` comparison.


juandspy (approved):

gburges:
This is good, but can potentially break every/any service that relies on partial match

PR #1208, opened 2025-04-22
task: Update go version and modules

SteveHNH:
CVE remediations


wcmitchell (approved):
Looks all good to me, still a couple of vulns present, but this is a great step forward

wcmitchell:
Everything but test-ff-local passing, that's due to an issue with the test script rather than the test itself, gonna accept this

PR #1212, opened 2025-04-29
chore(kafka): Overal strimzi/kafka version updates

wcmitchell:
We've had our kafka/strimzi tests failing for some time now, so updating our versions of kafka/strimzi to at least get closer to modern times.

One suspicion we've had in investigating the failing tests is slow KafkaConnect startup times, so this commit also includes an update to the KafkaConnect image and the default resource requests made by it.


wcmitchell:
Only test-ff-local failing, which is expected currently. merging this in so we can unblock other PRs awaiting testing

PR #1220, opened 2025-05-28
Shared inmemorydb

wcmitchell:
I know it's giant, and a mess, but it's working! At least in testing...


adamrdrew:
We're not going to be able to do anything about PlatSec scan failing. It wants an update to go 1.23.10 or above and that isn't available on UBI9 go-toolset yet :(

juandspy:
> UBI9 go-toolset

will this block the PR or are we ok merging this with that CVE?

bsquizz:
The only failing test is 'test-ff-local' which is known to be flaky

PR #1223, opened 2025-05-30
Fixing konflux pipelines

maknop:
# Purpose

Not all deps were properly downloaded to the `go.mod` file. Running `go mod tidy` to correct the issue. Additionally we updated the UBI8 go toolset version to UBI9 go toolset 1.23.


adamrdrew (approved):
LGTM!

PR #1227, opened 2025-06-03
updates default kafka naming convention to use just env name

tonytheleg:
By default, Kafka-related objects derived their naming convention from ClowdEnvironment Name + UID to ensure uniqueness. In ephemeral, this makes it impossible to predict the name of Kafka resources without first creating them which impacts those using bonfire to deploy objects like KafkaConnectors or Users that need to know the Kafka/Connect cluster names for proper labeling.

Since the ClowdEnv names are unique enough, in that they are cluster scoped and there can only be one for a desired name, this change removes the UID portion to allow for Clowder/Bonfire users to predict the naming of their resources using the env name alone.

Included in the change:
* API/Docs or any references to the previous default value to ensure consistent info throughout


controllers/cloud.redhat.com/providers/kafka/provider.go
```diff
@@ -86,14 +86,11 @@ func getKafkaUsername(env *crd.ClowdEnvironment, app *crd.ClowdApp) string {
...
-		// convert e.UID (which is a apimachinery types.UID) to string
-		// types.UID is a string alias so this should not fail...
-		uidString := string(e.UID)
-
-		// append the initial portion of the UUID onto the kafka cluster's name
-		return fmt.Sprintf("%s-%s", e.Name, strings.Split(uidString, "-")[0])
+		// historically this function returned <ClowdEnvironment Name>-<UID> for uniqueness
+		// but affects ClowdApp consumers that need a predictable naming convention to create
+		// other Kafka related resources (Users, Connectors, etc)
+		// Instead, we return the env name which is unique enough since objects a namespaced
```
bsquizz, line 92:
```suggestion
		// Instead, we return the env name which is unique enough since objects are namespaced
```

bsquizz (approved):

bsquizz (approved):

PR #1266, opened 2025-07-07
RHCLOUD-41032: Create otel collector sidecar as container

Sgitario:
The OpenTelemetry collector sidecar is currently implemented as an `InitContainer` instead of a regular `Container`, causing it to terminate before the main application starts. This prevents the collector from receiving telemetry data from the application at runtime.


Sgitario:
Close as won't do.

PR #1284, opened 2025-07-15
Clowder local development for MacOS users

maknop:
# Purpose

This PR will add a new make target for pushing the built clowder image to the local registry in Minikube. At the time of writing this, Docker is the recommended way of handling this on MacOS. 

Additionally, this PR adds documentation regarding how to get Clowder working in Minikube on MacOS. 

Also removing the roadmap section of the docs.


adamrdrew (approved):
LGTM! I checked it and see nothing wrong, and AI agrees so go ahead and merge it!

PR #1287, opened 2025-07-16
Creating new CRD for referencing app dependencies across multiple clusters

maknop:
## Description

This PR adds support for application dependencies to be referenced from external clusters. A new CRD called `ClowdAppRef` has been added which allows users to define these dependencies in their ClowdApps. 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

Assisted-By: Cursor v1.2.4 / claude-4-sonnet


tests/kuttl/test-clowdappref/03-config-assert.yaml
```diff
@@ -0,0 +1,36 @@
...
+kind: TestStep
+commands:
+- script: |
+    # Get the consumer-app config and verify it contains the ClowdAppRef endpoints
+    kubectl get secret consumer-app -n test-clowdappref -o jsonpath='{.data.cdappconfig\.json}' | base64 -d > /tmp/consumer-app-config.json
+    echo "Consumer app config:"
+    cat /tmp/consumer-app-config.json
+    
+    # Check that remote-service endpoints are present
+    grep -q "auth.remote-cluster.example.com" /tmp/consumer-app-config.json
```
bsquizz, line 12: The greps are good to check the presence of these strings ... but I think it would be better to assert that the JSON structure for this app in the cdappconfig.json looks like you would expect it to. For that, I'd probably use some jq asserts similar to https://github.com/RedHatInsights/clowder/blob/master/tests/kuttl/test-basic-app/02-json-asserts.yaml#L5-L23

bsquizz (changes requested):
Looking really good! I notice a bunch of lint errors, other than that just a few observations.

bsquizz (approved):

PR #1293, opened 2025-07-21
Add prometheus gateway

adamrdrew:
Add an optional prometheus gateway that clowder can deploy


bsquizz (commented):
This looks pretty solid to me ... definitely good enough to ship to crcd and give it a test. The main thing that I'm "concerned" about is all the changes to deploy.yml and deploy-mutate.yml -- seems like a lot of unrelated things changed there. Taking a closer look ...

adamrdrew:
Totally different test failing this time. I'm re-running PR checks on https://github.com/RedHatInsights/clowder/pull/1294 - if that fails in another seemingly random way then I say we merge both of these

PR #1306, opened 2025-07-29
Add .golangci.yml and update lint workflow

bsquizz:
We want to ignore certain lint errors, like the static check error for "deprecated".

This PR attempts to set up a more modern config for golangci-lint so that we can add checks to the exclusion list.

We also fix a LOT of lint issues via this PR.


bsquizz:
The only test failing is our old friend 'test-ff-local'

PR #1343, opened 2025-08-25
Set default topic replication factor

bsquizz:
This sets the default topic replication factor to match the number of replicas when kafka is provisioned using the strimzi provider.


wcmitchell (approved):
Reasonable and looks all good to me!

PR #1344, opened 2025-08-28
Updated default cpu limit

ffeghali:
The default cpu limit is now null. This is to remove the cpu limit when a limit is not specified.

### DIFF
--- a/controllers/cloud.redhat.com/providers/deployment/impl.go
+++ b/controllers/cloud.redhat.com/providers/deployment/impl.go
@@ -441,7 +441,7 @@ func ProcessResources(pod *crd.PodSpec, env *crd.ClowdEnvironment) core.Resource
 	if *pod.Resources.Limits.Cpu() != nullCPU {
 		lcpu = pod.Resources.Limits["cpu"]
 	} else {
-		lcpu = env.Spec.ResourceDefaults.Limits["cpu"]
+		lcpu = nullCPU
 	}
 
 	if *pod.Resources.Limits.Memory() != nullMemory {


bsquizz (approved):

bsquizz:
Ok, I think what needs to happen is the 'cpu' key needs to be absent instead of being set to '0':

What I'd recommend doing is:
* At [this point](https://github.com/RedHatInsights/clowder/pull/1344/files#diff-4cf0b90fba826deeff1fe02f04ab8a067d9ba5ad2975a3610ce9131cb024fe14L465) instead of returning, define the resourceRequirements like this:

```go
    resourceRequirements = core.ResourceRequirements{
		Limits: core.ResourceList{
			"memory": lmemory,
		},
		Requests: core.ResourceList{
			"cpu":    rcpu,
			"memory": rmemory,
		},
	}
```

Move the logic [here](https://github.com/RedHatInsights/clowder/pull/1344/files#diff-4cf0b90fba826deeff1fe02f04ab8a067d9ba5ad2975a3610ce9131cb024fe14L447-L451) below that, and check for these conditions:

* Condition 1: If `*pod.Resources.Limits.Cpu() != nullCPU`, then `resourceRequirements.Limits["cpu"] = pod.Resources.Limits["cpu"]`
* Condition 2: If the config for `env.Spec.ResourceDefaults.Limits["cpu"]` is defined (in other words, if the key is present in that config), then `resourceRequirements.Limits["cpu"] = env.Spec.ResourceDefaults.Limits["cpu"]`
* If the cpu limit is not defined on the `.pod.Resources.Limits` and it is also not defined at `env.Spec.ResourceDefaults.Limits["cpu"]`, then the 'cpu' key should not be set at all.

bsquizz (dismissed):

PR #1387, opened 2025-09-29
Support for configuring web provider TLS per-app

bsquizz:
Currently the web provider's TLS mode is controlled at the ClowdEnvironment level. This PR would add configuration options on a ClowdApp's deployments to enable (or disable) TLS for specific deployments if desired. If the configuration for this on the deployment is absent, then we will default to the ClowdEnvironment setting.

For example, if the ClowdEnvironment is configured like this:
```yaml
  providers:
    web:
      port: 8000
      privatePort: 10000
      mode: operator
      tls:
        enabled: false  # TLS disabled by default in environment
        port: 8800
        privatePort: 18800
```

A deployment under a ClowdApp could turn TLS support on using:
```yaml
    webServices:
      private:
        enabled: true
        tls: true  # Override: Enable TLS for private service
      public:
        enabled: true
        tls: true  # Override: Enable TLS for public service
```


controllers/cloud.redhat.com/providers/featureflags/provider.go
```diff
@@ -10,10 +10,10 @@ import (
 )
 
 // DefaultImageFeatureFlagsUnleash defines the default Unleash server image for feature flags
-var DefaultImageFeatureFlagsUnleash = "unleashorg/unleash-server:5.6.9"
+var DefaultImageFeatureFlagsUnleash = "quay.io/app-sre/unleash-server:5.6.9"
```
bsquizz, line 13: Woops this accidentally snuck in, I'll remove it

adamrdrew (commented):
LGTM

I had both ChatGPT and Gemini review the PR by giving them the diff and your description and both said it was excellent, and neither pointed out any problems.

I also went through the changes and don't see any problems or avenues for improvement.

PR #1398, opened 2025-10-03
fix: Avoid setting tlsPort/tlsPrivatePort in dependency endpoint when TLS is not enabled for web service

JuanmaBM:
When TLS ports are set on the ClowdEnv but TLS is disabled for an app, the ports are still advertised in the cdappconfig and this causes issues for some teams (some apps do not like seeing 'tlsPort' defined without the ca certificate provided)

The TLS ports will be set to '0' in cdappconfigs if the deployment does not have TLS enabled.

Updates made by @bsquizz --
* Using a shared function for 'IsPublicTLSEnabled' / 'IsPrivateTLSEnabled' (located in provider utils).
* The logic for the ClowdAppRef needed some tweaking... since a ClowdAppRef is referring to a ClowdApp / ClowdEnvironment in a remote cluster... if we don't find config values in the "remote environment config" then we will default to using the "local environment config"
* I changed the ClowdAppRefRemoteEnvironment to now have a 'tls' field so we could pass type *TLS to the utils functions... whether we are analyzing a ClowdEnvironment or a ClowdAppRefEnvironment
* kuttl test changes

  ### ⚠️ Breaking Change - ClowdAppRef API Update
  
  This PR makes a change to the ClowdAppRef, which should be low impact because the use of the feature by app teams is currently minimal (perhaps even none)

  #### What Changed

  The `ClowdAppRefRemoteEnvironment` spec has been updated to use a unified `tls` configuration object instead of separate port fields:

  **Old structure (deprecated):**
  ```yaml
  remoteEnvironment:
    port: 8080
    tlsPort: 8443              # ❌ Removed
    privatePort: 10000
    tlsPrivatePort: 10443      # ❌ Removed
    ```

  New structure:
  ```yaml
  remoteEnvironment:
    port: 8080
    privatePort: 10000
    tls:                       # ✅ New unified TLS config
      port: 8443
      privatePort: 10443
      enabled: true
    ```
    
  #### Migration Guide

  If you have existing ClowdAppRef resources, update them as follows:

  Before:
  ```yaml
  kind: ClowdAppRef
  metadata:
    name: my-remote-app
  spec:
    envName: production
    remoteEnvironment:
      name: remote-env
      port: 8000
      tlsPort: 8443
      privatePort: 10000
      tlsPrivatePort: 10443
    # ... rest of spec
    ```

  After:
  ```yaml
  kind: ClowdAppRef
  metadata:
    name: my-remote-app
  spec:
    envName: production
    remoteEnvironment:
      name: remote-env
      port: 8000
      privatePort: 10000
      tls:
        port: 8443
        privatePort: 10443
    # ... rest of spec
    ```
    
  #### Default Behavior

  If you omit TLS port values, the system will default to using the TLS configuration from your local ClowdEnvironment. The following configuration maintains backward compatibility for most use cases:

  ```yaml
  remoteEnvironment:
    name: remote-env
    port: 8000
    # tls section omitted - will use local ClowdEnvironment TLS settings
  ```


controllers/cloud.redhat.com/providers/dependencies/impl.go
```diff
@@ -24,16 +24,25 @@ func (dep *dependenciesProvider) makeDependencies(app *crd.ClowdApp) error {
 	// Process self endpoints
 	appMap := map[string]crd.ClowdApp{app.Name: *app}
 	appRefMap := map[string]crd.ClowdAppRef{} // empty since we're only processing self
+
+	// Avoid mapping ports when TLS is false
+	tlsPort := int32(0)
+	tlsPrivPort := int32(0)
+	if dep.Env.Spec.Providers.Web.TLS.Enabled {
```
bsquizz, line 31: TLS can be turned on/off for a ClowdApp in more than 1 place now (see the function `isPublicTLSEnabled` and `isPrivateTLSEnabled`, you probably want to use those here now instead of only checking the ClowdEnvironment)

bsquizz (changes requested):

controllers/cloud.redhat.com/providers/dependencies/impl.go
JuanmaBM: Yeah, I saw that, but I don't want to overengineering so much this fix. Also I didn't sure the order of this, I mean, if you have the tls enable in env but false in clowdapp, it should be enable or disable? I can add something more robust in term of reliability but need to know the order to apply the different conditions

bsquizz: If TLS is enabled in env but false in ClowdApp, it will be disabled on the app

The `isPublicTLSEnabled` and `isPrivateTLSEnabled` checks both the env and the app

maknop (approved):
LGTM!

PR #1411, opened 2025-10-09
Always populate tlsCAPath if TLS ports are defined in environment

bsquizz:
As a follow-up to #1387 -- we need to always provide apps with the TlsCAPath if we have the ClowdEnvironment TLS ports configured.

As an example, if 'app B' depends on 'app A' and 'app A' has TLS enabled, then app B needs the CA certificate. Rather than iterating through app B's dependencies to see if any have TLS enabled, we'll just go ahead and set the TLS CA path whether it is going to use TLS to connect to app A or not.

We will assume that if a ClowdEnvironment has the TLS port/privatePort configured then the cluster has been set up with the Service CA Operator, and so it is safe to assume the `openshift-service-ca.crt` ConfigMap is available in every namespace.


maknop (approved):
LGTM!

PR #1419, opened 2025-10-22
chore(IQE-2762): Update IQE selenium image

akhil-jha:


mshriver:
https://github.com/RedHatInsights/clowder/pull/1772

handling along with playwright support, way to go @mmusil

PR #1421, opened 2025-10-23
updating quic-go dependency

maknop:
## Purpose

Updating the `quic-go` dep from version `0.52.0` to `0.54.1`.


bsquizz:
Handled via #1551

PR #1581, opened 2025-12-17
feat: add configuration to use CodeBuild to release testing process

JuanmaBM:
Add configuration to use CodeBuild to release testing process


JuanmaBM:
> Very cool! Great work on this. I think it is ready to go in. The only thing I hope we can improve in a follow-up PR is:
> 
> Instead of having all the install/pre_build/build shell commands defined directly in `buildspec.yml`, I wonder if we can just store all that logic in .sh files and call it from the YAML? That would make buildspec.yml look a lot cleaner.
> 
> But, I don't think that needs to block this PR.

Thank you! Noted, I'll create a task to move all configuration into `buildspec.yaml`

PR #1707, opened 2026-02-23
fix: add missing Type field to LoggingConfig

romanblanco:
Adding type missing on LoggingConfig in attempt to revert commit https://github.com/RedHatInsights/compliance-backend/commit/482546a652656722545ef2c681669c6bb132302a and correct cloudwatch logging

clowder-common-ruby seems to provide it in
https://github.com/RedHatInsights/clowder-common-ruby/blob/c10476027cd07b7d89b83f5826aeff84c7fc1cfd/lib/clowder-common-ruby/rails_config.rb#L110


bsquizz (dismissed):
I've checked advisor, inventory, and ingress apps and none of them seem to check the value of 'type' in the LoggingConfig -- so it seems like setting this value should be ok.

PR #1759, opened 2026-04-02
Support of volumeMounts in sidecars

Sgitario:
## Summary

The Clowder `Sidecar` API currently does not support custom `volumeMounts`, which prevents use cases like sharing a log directory between the main container and the otel-collector sidecar. This is needed for file-based log collection (`filelog` receiver), where the application writes logs to a shared volume and the otel-collector tails them for forwarding to Splunk/SumoLogic.

The `filelog` approach is preferred over alternatives like `syslog` or `tcplog` because it requires zero coupling with the application — the app just writes to a file (which frameworks like Spring Boot, Quarkus, and NGINX already support natively), and the collector picks them up independently. It's also resilient to collector restarts since logs persist in the shared volume.

### Changes

- **API**: Add optional `volumeMounts` field (`[]v1.VolumeMount`) to the `Sidecar` struct in `clowdapp_types.go`. Fully backward compatible (`omitempty`).
- **Provider**: Append custom volume mounts after the default ConfigMap mount in `getOtelCollector()`.
- **Unit tests**: 4 new test cases covering no mounts, single mount, multiple mounts (with `readOnly`), nil sidecar, and empty slice.
- **Kuttl integration test**: New `test-sidecars-volume-mounts` suite with 3 scenarios: baseline (no custom mounts), single shared volume, and multiple volumes.

## Test plan

- [x] Unit tests pass (`go test ./controllers/cloud.redhat.com/providers/sidecar/... -v`)
- [ ] Kuttl integration tests pass in CI (`make kuttl KUTTL_TEST="--test=test-sidecars-volume-mounts"`)
- [ ] Existing sidecar kuttl tests still pass (no regressions)


bsquizz (dismissed):

bsquizz:
One of the new kuttl tests failed if you're able to take a look:

` --- FAIL: kuttl/harness/test-sidecars-volume-mounts (610.44s)`

Sgitario:
@bsquizz I don't have permission to run the e2e tests. Can you trigger it please?

controllers/cloud.redhat.com/providers/sidecar/default.go
bsquizz: Claude spotted this error:

  Silent no-op for token-refresher (medium concern)
                                                                                                                                                                                               
  The VolumeMounts field is on the shared Sidecar struct, but getTokenRefresher never receives the Sidecar pointer — it only takes appName. If a user adds volumeMounts to a token-refresher   
  sidecar entry, it will be silently ignored. Either:                                                                                                                                          
  - Document this limitation explicitly (in the API comment and/or docs), or                                                                                                                   
  - Extend getTokenRefresher to accept *crd.Sidecar for consistency                                                                                                                            
                                                                   
  This is important because the field appears in the API for all sidecar types, but only works for otel-collector.

bsquizz (changes requested):

bsquizz (approved):

ffeghali (approved):
/lgtm

PR #1760, opened 2026-04-08
fix: bypass informer cache for ClowdApp read in CJI controller

rodrigonull:
## Problem

PR #1742 introduced a generation check to prevent `runOnNotReady` CJI jobs from using stale ClowdApp specs. However, the fix was insufficient because the CJI controller reads the ClowdApp from the **informer cache**, which can serve an entirely stale but internally consistent object — where both `metadata.generation` and `status.generation` match at old values. The generation check passes, and the job is created with the old image.

We confirmed this in stage: CJI `run-db-migrations-90cb9b7` was created for image `90cb9b7`, but the resulting Job used old image `1f8810c`. Both the CJI and Job share the same creation timestamp (`22:01:45Z`) — the controller never requeued, proving the cache served stale data that passed the generation check.

## Solution

Replace `r.Get()` (informer cache) with `r.APIReader.Get()` (direct API server / etcd read) when fetching the ClowdApp in the CJI controller. This guarantees the controller sees the latest spec.

Combined with the existing generation check, the flow is now:

1. APIReader reads ClowdApp from etcd → gets new spec (`generation=N`, `status.generation=N-1`)
2. Generation check detects mismatch → requeue
3. ClowdApp controller reconciles → `status.generation=N`
4. CJI retries → generation matches → job created with correct image

## Changes

- **`clowdjobinvocation_controller.go`**: Add `APIReader client.Reader` field; use `r.APIReader.Get()` for the ClowdApp fetch
- **`run.go`**: Wire `mgr.GetAPIReader()` into the CJI reconciler
- **`clowdjobinvocation_controller_test.go`**: New unit tests covering generation check (match, mismatch, backward compat), APIReader vs cache behavior, missing app, and early-exit paths
- **`02-json-asserts.sh`**: Fix namespace and path bugs introduced in #1742

## Test plan

- [x] `go vet ./controllers/cloud.redhat.com/...` passes
- [x] `go test -run TestCJI ./controllers/cloud.redhat.com/` — all 5 tests pass
- [x] `make test` — full suite passes (pre-existing `covdata` tool error only)

### DIFF
--- a/controllers/cloud.redhat.com/clowdjobinvocation_controller.go
+++ b/controllers/cloud.redhat.com/clowdjobinvocation_controller.go
@@ -48,9 +48,10 @@ import (
 // ClowdJobInvocationReconciler reconciles a ClowdJobInvocation object
 type ClowdJobInvocationReconciler struct {
 	client.Client
-	Log      logr.Logger
-	Scheme   *runtime.Scheme
-	Recorder record.EventRecorder
+	APIReader client.Reader
+	Log       logr.Logger
+	Scheme    *runtime.Scheme
+	Recorder  record.EventRecorder
 }
 
 // +kubebuilder:rbac:groups=cloud.redhat.com,resources=clowdjobinvocations,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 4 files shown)


bsquizz:
e2e test 'test-sidecars' failed but that should not be related to this change

bsquizz (approved):

PR #1761, opened 2026-04-13
ServerAddress Set by ClowdApp

mjholder:
Ticket: https://redhat.atlassian.net/browse/ENGPROD-9733

App-level trigger metadata now takes precedence over auto-injected env-level values. If the ClowdApp doesn't set a field, the env value fills in as a default. This applies to all trigger types (prometheus, kafka, etc.).

Kuttl test added: tests/kuttl/test-autoscaler-prometheus-app-serveraddress/ — verifies app-specified serverAddress is preserved and updates propagate to the ScaledObject.

This change came from a bug reported by the notifications team when trying to test auto scaling based on a different prometheus server. When they updated the clowdapp Notifications Engine's serverAddress, clowder reconciled the resources but did not actually apply the new address. So we are in a weird situation where the resource doesn't match the definition but clowder says everything is reconciled correctly.

Thread: https://redhat-internal.slack.com/archives/C07FQ00BJV8/p1774452941621759

Verification:
Deployed fixed Clowder to minikube, manual check confirmed ScaledObject now uses app-level serverAddress
New kuttl test: PASS
Existing tests (app-interface, status-only-update): PASS — no regressions


JGiola (dismissed):

JGiola (approved):

wcmitchell (approved):
Looks good from a quick readover, well explained, small changeset with tests. 👍

PR #1767, opened 2026-04-28
Feat/add v2 cdappconfig endpoints

JuanmaBM:
## Summary

Add V2 dependency endpoint format to cdappconfig.json with URI-based structure for improved usability. V2 coexists with V1 for backward compatibility.

## Changes

- **Schema**: Add `dependencyEndpoints.v2` and `privateDependencyEndpoints.v2` to app config schema
- **Structure**: Organize endpoints by app name → service name (e.g., `puptoo.processor`)
- **Format**: Use complete URIs (`protocol://hostname:port`) instead of separate hostname/port fields
- **TLS**: Add `ca_certificate` field with cert path for HTTPS endpoints
- **Port Types**: Create separate entries for base, TLS, H2C, and H2C TLS ports with suffixes (`_tls`, `_h2c`, `_h2c_tls`)
- **Backward Compatibility**: V1 `endpoints[]` and `privateEndpoints[]` arrays remain unchanged

### Example Output

```json
{
  "endpoints": [...],  // V1 unchanged
  "dependencyEndpoints": {
    "v2": {
      "puptoo": {
        "processor": {
          "uri": "http://puptoo-processor.namespace.svc:8000"
        },
        "processor_tls": {
          "uri": "https://puptoo-processor.namespace.svc:8443",
          "ca_certificate": "/cdapp/certs/service-ca.crt"
        },
        "processor_h2c": {
          "uri": "http://puptoo-processor.namespace.svc:9800"
        },
        "processor_h2c_tls": {
          "uri": "https://puptoo-processor.namespace.svc:9443",
          "ca_certificate": "/cdapp/certs/service-ca.crt"
        }
      }
    }
  },
  "privateDependencyEndpoints": {
    "v2": {
      "puptoo": {
        "processor": {
          "uri": "http://puptoo-processor.namespace.svc:10000"
        },
        "processor_tls": {
          "uri": "https://puptoo-processor.namespace.svc:10443",
          "ca_certificate": "/cdapp/certs/service-ca.crt"
        }
      }
    }
  }
}
```

## Testing

- ✅ Unit tests: 21 tests covering URI construction, endpoint transformation, TLS handling
- ✅ KUTTL E2E test: Validates V2 structure in cdappconfig.json with TLS and H2C enabled
- ✅ Backward compatibility: Existing V1 tests continue passing
- ✅ All port types tested: base, TLS, H2C, H2C TLS

## Backward Compatibility

Fully backward compatible. V1 `endpoints[]` and `privateEndpoints[]` arrays are unchanged. V2 is additive only - applications can continue using V1 or migrate to V2 at their own pace.

### DIFF
--- a/controllers/cloud.redhat.com/providers/dependencies/impl.go
+++ b/controllers/cloud.redhat.com/providers/dependencies/impl.go
@@ -2,6 +2,8 @@ package dependencies
 
 import (
 	"fmt"
+	"net"
+	"net/url"
 
 	"github.com/RedHatInsights/rhc-osdk-utils/utils"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -311,7 +313,11 @@ func processAppAndAppRefEndpoints(
 
 // constructEndpointURI builds a complete URI from protocol, hostname, and port
 func constructEndpointURI(protocol string, hostname string, port int) string {
-	return fmt.Sprintf("%s://%s:%d", protocol, hostname, port)
+	u := &url.URL{
+		Scheme: protocol,
+		Host:   net.JoinHostPort(hostname, fmt.Sprintf("%d", port)),
+	}
+	return u.String()
 }
 
 // buildV2EndpointMap transforms V1 endpoint data into V2 format
--- a/controllers/cloud.redhat.com/providers/dependencies/impl.go
+++ b/controllers/cloud.redhat.com/providers/dependencies/impl.go
@@ -92,6 +92,11 @@ func (dep *dependenciesProvider) makeDependencies(app *crd.ClowdApp) error {
 
 	dep.Config.Endpoints = depConfig
 	dep.Config.PrivateEndpoints = privDepConfig
+
+	// Populate V2 endpoint structures
+	dep.makeV2DependencyEndpoints()
+	dep.makeV2PrivateDependencyEndpoints()
+
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


controllers/cloud.redhat.com/providers/dependencies/impl.go
```diff
@@ -303,3 +308,120 @@ func processAppAndAppRefEndpoints(
...
+		// Add H2C TLS port endpoint (https)
+		if ep.H2CTLSPort != nil && *ep.H2CTLSPort > 0 {
+			endpoint := config.DependencyEndpointV2{
+				Uri: constructEndpointURI("https", ep.Hostname, *ep.H2CTLSPort),
+			}
+			if ep.TlsCAPath != nil {
+				endpoint.CaCertificate = ep.TlsCAPath
+			}
+			v2Map[ep.App][ep.Name+"_h2c_tls"] = endpoint
+		}
```
JGiola, line 372: In this section we always assume that a custom port greater than 0 is passed to the configuration. Is safe to assume that? User can leave the port parameter empty to assume the default port? (80 for http, 443 for https, etc)

JuanmaBM: Good catch! 

Yes, it's safe to assume this. Port can be 0 when a service only exposes HTTPS (or H2C). The `anyNotZero` check at line 168 ensures at least one port is configured before creating the endpoint.                                           
   
 In practice:                                                                                                                                                                                                                                
  - If `Port=0` but `TlsPort>0` → only the `_tls` endpoint is created (expected behavior)
  - If all ports are 0 → no endpoint is created
                                                                                                                                                                                                                                              
Ports come from the `ClowdEnvironment` config, we don't use default ports (80/443) since in Kubernetes each service has its own configured ports.                                                                                           
                                                                                                                                                                                                                                              
Personally, I prefer to keep it explicit rather than adding "magic" logic that makes assumptions the user isn't aware of.

controllers/cloud.redhat.com/providers/dependencies/impl.go
```diff
@@ -303,3 +308,120 @@ func processAppAndAppRefEndpoints(
 
 	return missingDeps
 }
+
+// constructEndpointURI builds a complete URI from protocol, hostname, and port
+func constructEndpointURI(protocol string, hostname string, port int) string {
+	return fmt.Sprintf("%s://%s:%d", protocol, hostname, port)
```
wcmitchell: Would it be safer to import `net/url` here and use that to build the URI? IIRC it has some safeguards so we could do something like:

`func constructEndpointURI(protocol string, hostname string, port int) string {
    u := &url.URL{
        Scheme: protocol,
        Host: net.JoinHostPort(hostname, Sprintf("%d", port))
    }
}`

It seems below like we have overall control over the inputs, so the guards in url.URL may not be necessary, but wanted to mention just in case.

wcmitchell (dismissed):

controllers/cloud.redhat.com/providers/dependencies/impl.go
JuanmaBM: Good recommendation! Updated to use net/url and net.JoinHostPort for better robustness

JGiola (approved):

PR #1773, opened 2026-05-12
Multicluster Support for Clowder

JuanmaBM:
This PR implements comprehensive multicluster support for Clowder, enabling applications to seamlessly connect to dependencies across different clusters with proper authentication and TLS configuration.

### The Multicluster Challenge

Modern cloud-native applications need to connect to services across multiple clusters:
- **In-cluster services** (ClowdApp) - Same cluster, direct k8s Service access, network isolation provides security
- **Cross-cluster services** (ClowdAppRef) - Different clusters, routed through API gateways, require authentication

The existing V1 endpoint system doesn't distinguish between these scenarios, forcing applications to implement custom logic for:
- Determining if authentication is needed
- Selecting the appropriate protocol and port
- Configuring TLS certificates correctly
- Handling different connection patterns

### V1 Limitations

**Example V1 configuration:**
```json
{
  "endpoints": [
    {
      "app": "rbac",
      "hostname": "rbac-service.env.svc",
      "port": 8000,
      "tlsPort": 8443,
      "h2cPort": 0,
      "h2cTLSPort": 0
    }
  ]
}
```

**Problems:**
- Applications must decide which port/protocol to use
- No indication if authentication is required
- Inconsistent choices across different apps
- CA certificate configuration is global, not per-endpoint
- Error-prone (easy to use wrong port or forget TLS)

### Clean, Simple Configuration

**Example V2 configuration (in-cluster):**
```json
{
  "dependencyEndpoints": {
    "v2": {
      "rbac": {
        "service": {
          "uri": "https://rbac-service.env.svc:8443",
          "ca_certificate": "/cdapp/certs/service-ca.crt",
          "authenticated": false
        }
      }
    }
  }
}
```

**Example V2 configuration (cross-cluster):**
```json
{
  "dependencyEndpoints": {
    "v2": {
      "external-service": {
        "api": {
          "uri": "https://external-api.gateway.example.com:443",
          "authenticated": true
        }
      }
    }
  }
}
```

### Key Features

✅ **Single URI** - Applications use one field, no selection logic needed  
✅ **Automatic Protocol Selection** - TLS > H2C TLS > H2C > plaintext priority  
✅ **Per-Endpoint CA Certificates** - Only present when needed for in-cluster TLS  
✅ **Authenticated Field** - Explicit signal for authentication requirements  
✅ **Backward Compatible** - V1 endpoints still generated alongside V2

## Endpoint Selection Priority

V2 endpoints follow this priority order (highest to lowest):

1. **TLS Port** (`tlsPort`) - HTTPS with TLS encryption
2. **H2C TLS Port** (`h2cTLSPort`) - HTTP/2 cleartext over TLS
3. **H2C Port** (`h2cPort`) - HTTP/2 cleartext without TLS
4. **Standard Port** (`port`) - HTTP plaintext

**Example:** If a service exposes port 8000 (HTTP), tlsPort 8443 (HTTPS), and h2cPort 8080, V2 will always select the HTTPS endpoint at port 8443.

### Complete V2 Configuration Example

```json
{
  "dependencyEndpoints": {
    "v2": {
      "rbac": {
        "service": {
          "uri": "https://rbac-service.env.svc:8443",
          "ca_certificate": "/cdapp/certs/service-ca.crt",
          "authenticated": false
        }
      },
      "external-api": {
        "gateway": {
          "uri": "https://api.external.example.com:443",
          "authenticated": true
        }
      }
    }
  },
  "privateDependencyEndpoints": {
    "v2": {
      "internal-service": {
        "admin": {
          "uri": "https://internal-admin.env.svc:10443",
          "ca_certificate": "/cdapp/certs/service-ca.crt",
          "authenticated": false
        }
      }
    }
  },
  "tlsCAPath": "/cdapp/certs/service-ca.crt",  // Root-level for backward compatibility
  "endpoints": [...],  // V1 format still present
  "privateEndpoints": [...]  // V1 private format still present
}
```

## Backward Compatibility

✅ **Fully backward compatible**

- **V1 endpoints** (`endpoints[]` and `privateEndpoints[]`) continue to be generated
- **Root-level `tlsCAPath`** maintained for existing applications
- **V2 is additive** - both formats coexist in the same configuration
- **No CRD changes** required for basic usage
- **Existing applications** are completely unaffected

### Migration Path

1. **Deploy this Clowder version** - V2 endpoints appear in config alongside V1
2. **Update applications incrementally** - Read from `dependencyEndpoints.v2`
3. **Remove custom logic** - Delete endpoint selection and CA configuration code
4. **Use `authenticated` field** - Add authentication only when `authenticated: true`
5. **Simplify applications** - One line of code instead of complex selection logic

## Related Issues

Closes: https://redhat.atlassian.net/browse/ENGPROD-9704
Closes: https://redhat.atlassian.net/browse/ENGPROD-9960

---

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>
Signed-off-by: JuanmaBM <juanmabareamartinez@gmail.com>


controllers/cloud.redhat.com/providers/dependencies/impl.go
```diff
@@ -269,22 +282,30 @@ func configureAppRefDependencyEndpoints(innerDeployment *crd.ClowdAppRefDeployme
...
+		tlsPrivatePort = coalesceInt32(depAppRef.Spec.RemoteEnvironment.TLS.Port, envWebConfig.TLS.Port)
+	}
+
 	webConfig := crd.WebConfig{
 		Port:           coalesceInt32(depAppRef.Spec.RemoteEnvironment.Port, envWebConfig.Port),
 		PrivatePort:    coalesceInt32(depAppRef.Spec.RemoteEnvironment.PrivatePort, envWebConfig.PrivatePort),
 		H2CPort:        coalesceInt32(depAppRef.Spec.RemoteEnvironment.H2CPort, envWebConfig.H2CPort),
 		H2CPrivatePort: coalesceInt32(depAppRef.Spec.RemoteEnvironment.H2CPrivatePort, envWebConfig.H2CPrivatePort),
 		TLS: crd.TLS{
+			Enabled:        depAppRef.Spec.RemoteEnvironment.TLS.Enabled,
```
bsquizz: Should this be `depAppRef.Spec.RemoteEnvironment.TLS.Enabled || envWebConfig.TLS.Enabled` to fall back to the environment default?

controllers/cloud.redhat.com/providers/web/default.go
```diff
@@ -86,19 +87,14 @@ func (web *webProvider) Provide(app *crd.ClowdApp) error {
...
 			if err := web.Cache.Update(provCronjob.CoreCronJob, &innerItem); err != nil {
 				return err
-
 			}
 		}
 	}
 
-	if envTLSConfig.Enabled {
-		// if TLS is enabled environment-wide, set 'tlsCAPath' in the root level of cdappconfig
-		web.Config.TlsCAPath = provutils.GetServiceCACertPath()
```
bsquizz, line 100: Will applications consuming this root-level field will silently get `null` now?. Any app that reads the root-level field (as a default/fallback) will break. Is this removal intentional?

controllers/cloud.redhat.com/providers/web/local.go
```diff
@@ -182,11 +182,6 @@ func (web *localWebProvider) Provide(app *crd.ClowdApp) error {
 		}
 	}
 
-	if envTLSConfig.Enabled {
```
bsquizz, line 185: Same question here

controllers/cloud.redhat.com/providers/dependencies/impl.go
```diff
@@ -362,41 +383,41 @@ func buildV2EndpointMap(endpoints []config.DependencyEndpoint) map[string]map[st
...
-		if ep.TlsPort != nil && *ep.TlsPort > 0 {
-			endpoint := config.DependencyEndpointV2{
-				Uri: constructEndpointURI("https", ep.Hostname, *ep.TlsPort),
-			}
-			if ep.TlsCAPath != nil {
-				endpoint.CaCertificate = ep.TlsCAPath
-			}
-			v2Map[ep.App][ep.Name+"_tls"] = endpoint
+		switch {
+		case ep.TlsPort != nil && *ep.TlsPort > 0:
```
bsquizz: What happens if an app provides both REST API and a H2C service (like grpc) from the same application? We won't broadcast both of them anymore right? If I recall, some apps wanted to offer both of these protocols from the same app/container.

bsquizz (changes requested):

controllers/cloud.redhat.com/providers/dependencies/impl.go
JuanmaBM: Yes, if you take a look at the document written by Jozef [here](https://docs.google.com/document/d/1g-QTW1_Lo3dLwfMKFx2M8T5H82RHFNuS9f7HW4ZTqsA/edit?tab=t.0#bookmark=id.bkla2gmowsnn), one of the key points is: "One entry per dependency endpoint, not per protocol"

Of course, you can keep using V1 endpoints to define multiple protocols. Although it's a good point that you mentioned, if there are apps that have different protocols, we definitely should provide a mechanism to do that.

jharting: @bsquizz I think that will still be possible - a service can expose REST as the public endpoint and GRPC as the private endpoint. The configuration (uri, ca_certificate) is specific for each endpoint.

As for which port to use, the H2C port is very specific (only needed by Gateways). GRPC clients initiate HTTP 2 connections right away and don't need the hint from Kubernetes API. I.e. I think this can be simplified to just
- TLS if TLS port is set
- plaintext port otherwise

tests/kuttl/test-v2-endpoints-clowdapp-plaintext/02-json-asserts.sh
```diff
@@ -0,0 +1,32 @@
...
+for i in {1..10}; do
+    kubectl get secret --namespace="${NAMESPACE}" "${APP_NAME}" && break || sleep 1
+done
+
+# Extract config from secret
+kubectl get secret --namespace="${NAMESPACE}" "${APP_NAME}" -o json > "${TMP_DIR}/${TEST_NAME}"
+jq -r '.data["cdappconfig.json"]' < "${TMP_DIR}/${TEST_NAME}" | base64 -d > "${TMP_DIR}/${TEST_NAME}-json"
+
+# Run assertions - verify v2 endpoint structure (plaintext, in-cluster)
+jq -r '.dependencyEndpoints.v2.rbac.service.uri == "http://rbac-service.test-v2-clowdapp-plaintext.svc:8000"' -e < "${TMP_DIR}/${TEST_NAME}-json"
```
jharting, line 30: consider also validating that ca_certificate is undefined in this case

tests/kuttl/test-v2-endpoints-clowdappref-plaintext/02-json-asserts.sh
```diff
@@ -0,0 +1,32 @@
...
+for i in {1..10}; do
+    kubectl get secret --namespace="${NAMESPACE}" "${APP_NAME}" && break || sleep 1
+done
+
+# Extract config from secret
+kubectl get secret --namespace="${NAMESPACE}" "${APP_NAME}" -o json > "${TMP_DIR}/${TEST_NAME}"
+jq -r '.data["cdappconfig.json"]' < "${TMP_DIR}/${TEST_NAME}" | base64 -d > "${TMP_DIR}/${TEST_NAME}-json"
+
+# Run assertions - verify v2 endpoint structure
+jq -r '.dependencyEndpoints.v2.rbac.service.uri == "http://rbac-dev.local:8000"' -e < "${TMP_DIR}/${TEST_NAME}-json"
```
jharting, line 30: consider also validating that ca_certificate is undefined in this case

tests/kuttl/test-v2-endpoints-clowdappref-tls/02-json-asserts.sh
```diff
@@ -0,0 +1,32 @@
...
+for i in {1..10}; do
+    kubectl get secret --namespace="${NAMESPACE}" "${APP_NAME}" && break || sleep 1
+done
+
+# Extract config from secret
+kubectl get secret --namespace="${NAMESPACE}" "${APP_NAME}" -o json > "${TMP_DIR}/${TEST_NAME}"
+jq -r '.data["cdappconfig.json"]' < "${TMP_DIR}/${TEST_NAME}" | base64 -d > "${TMP_DIR}/${TEST_NAME}-json"
+
+# Run assertions - verify v2 endpoint structure with TLS (external, no CA)
+jq -r '.dependencyEndpoints.v2.rbac.service.uri == "https://rbac.external-cluster.example.com:8443"' -e < "${TMP_DIR}/${TEST_NAME}-json"
```
jharting, line 30: consider also validating that ca_certificate is undefined in this case

controllers/cloud.redhat.com/providers/dependencies/impl.go
JuanmaBM: @jharting  mentioned simplifying the V2 endpoint selection logic to just:
  - TLS if TLS port is set
  - plaintext port otherwise                                                                                            
                                                                                                                        
However, I'm concerned about breaking apps that expose both REST and gRPC on different ports within the same endpoint (public or private). For example:

  Scenario:
```
  # ClowdEnvironment
  spec:                                                     
    providers:
      web:                                                                                                              
        port: 8000      # REST HTTP                                                                                     
        h2cPort: 9800   # gRPC HTTP/2                                                                                   
                                                            
  # ClowdApp
  spec:                                                                                                                 
    deployments:
      - name: service                                                                                                   
        webServices:                                        
          public:
            enabled: true
            h2cEnabled: true  # Exposes BOTH protocols on same endpoint
```
                                                            
  Current implementation (TLS > H2C TLS > H2C > plaintext): 
  - V2 would expose: http://service:9800 (H2C has priority)
  - gRPC client connects to port 9800 (OK)                                                                           
 
  Simplified version (TLS > plaintext):
  - V2 would expose: http://service:8000 (plaintext)
  - gRPC client would try to connect to port 8000 (BROKEN)                                                            
  - Port 9800 wouldn't be exposed in V2 at all
                                                                                                                        
Questions:
  1. Are there real use cases where apps use h2cPort != port for gRPC within the same endpoint?
  2. Or do gRPC apps typically:                                                                                         
    - Use the same port for both REST and gRPC (gRPC clients initiate HTTP/2 on any port)?                              
    - Or use public for REST and private for gRPC (separate endpoints)?
                                                                                                                        
If there are no real cases where h2cPort needs to be different from port, then the simplification makes sense. Otherwise, I should keep the full priority logic.

bsquizz: Correct me if I am wrong, but I think that kessel enables both from the same deployment ... https://github.com/project-kessel/inventory-api/blob/main/deploy/kessel-inventory.yaml#L101-L105

controllers/cloud.redhat.com/providers/web/default.go
bsquizz: The code still removes the root-level tlsCAPath from both web/default.go and web/local.go. Any app reading cdappconfig.json at .tlsCAPath (the root-level field) will now silently get null when they upgrade to this version of Clowder. Can we fix this to not introduce a breaking change?

controllers/cloud.redhat.com/providers/web/local.go
bsquizz: The code still removes the root-level tlsCAPath from both web/default.go and web/local.go. Any app reading cdappconfig.json at .tlsCAPath (the root-level field) will now silently get null when they upgrade to this version of Clowder. Can we fix this to not introduce a breaking change?

controllers/cloud.redhat.com/providers/dependencies/impl.go
```diff
@@ -269,22 +282,30 @@ func configureAppRefDependencyEndpoints(innerDeployment *crd.ClowdAppRefDeployme
 	serviceName := depAppRef.GetDeploymentNamespacedName(innerDeployment).Name
 	apiPaths := provutils.GetAPIPaths(innerDeployment, serviceName)
 
+	// For ClowdAppRef TLS, if PrivatePort is not set but Port is, use Port as the default for PrivatePort
+	// This allows ClowdAppRef to specify only one TLS port that serves both public and private endpoints
+	tlsPrivatePort := coalesceInt32(depAppRef.Spec.RemoteEnvironment.TLS.PrivatePort, envWebConfig.TLS.PrivatePort)
+	if depAppRef.Spec.RemoteEnvironment.TLS.Enabled && tlsPrivatePort == 0 {
```
bsquizz: Found by Claude: The condition here only checks `depAppRef.Spec.RemoteEnvironment.TLS.Enabled`, but TLS can also be enabled via the env default (envWebConfig.TLS.Enabled). The fallback should trigger in that case too, so the condition should be:

```suggestion
    if (depAppRef.Spec.RemoteEnvironment.TLS.Enabled || envWebConfig.TLS.Enabled) && tlsPrivatePort == 0 {
```

bsquizz (dismissed):
See my newest replies, plus one new issue found during another Claude review.

controllers/cloud.redhat.com/providers/dependencies/impl.go
jharting: @JuanmaBM v2 dependency endpoint should either use TLSPort, or Port. Never h2c. More context in https://redhat-internal.slack.com/archives/C08QV47C53J/p1781607554558099

controllers/cloud.redhat.com/providers/dependencies/impl.go
```diff
@@ -424,30 +637,50 @@ func (dep *dependenciesProvider) makeV2DependencyEndpoints() {
...
+	envWebConfig := &dep.Env.Spec.Providers.Web
+
+	// Build maps for lookups
+	appMap := make(map[string]crd.ClowdApp)
+	for i := range apps.Items {
+		iapp := &apps.Items[i]
+		appMap[iapp.Name] = *iapp
 	}
 
-	v2Map := buildV2EndpointMap(publicFormat)
```
jharting, line 447: nitpick: with these changes, `buildV2EndpointMap()` seems no longer used, i.e. dead code that should be removed.

bsquizz (approved):

PR #1788, opened 2026-06-23
fix: update UBI base image from 9.7 to 9.8 to mitigate RPM vulnerabilities

ffeghali:
Addresses ENGPROD-10067 — updates ubi-minimal from 9.7-1778562320 to 9.8-1781496742 in both Dockerfile and build/Dockerfile-local. 

This pulls in el9_8 RPM fixes for gnutls, glibc, openssl, krb5-libs, libarchive, libcap, libnghttp2, p11-kit, systemd-libs, and glib2. Go module vulnerabilities (caddy, go-jose, pgx, stdlib) are already fixed on master.


RunItBack1127 (approved):
Dep updates look good, E2E tests pass!

PR #1794, opened 2026-06-30
Fix ignored k8s.io packages

JGiola:
The 'ignoreDeps' renovate config doesn't support regex, remove it and add all the k8s.io packages we have to exclude them.


RunItBack1127 (approved):
Matches what is in the `go.mod`, looks good! The rest of the `k8s.io` dependencies are indirect, so not needed to specify in the `renovate.json`

PR #1805, opened 2026-08-03
fix: update base images and Go deps to address CVEs

bsquizz:
## Summary

- Update `ubi9/ubi-minimal` from `9.8-1782797275` to `9.8-1785339117` to pick up RPM-level security fixes for glib2, libacl, glibc, libsolv, openssl-fips-provider, and coreutils
- Update `ubi9/go-toolset` from `1.25-1778504036` / `1.25.9-1777043046` to `1.25.9-1778675823` (latest available build) across Dockerfile and Tekton pipelines
- Bump Go version from 1.25.9 to 1.26.5 to address 12 stdlib CVEs (no k8s dependency changes)
- Update Go module dependencies:
  - `golang.org/x/text` 0.38.0 → 0.39.0 (CVE-2026-56852)
  - `google.golang.org/grpc` 1.81.1 → 1.82.1 (GHSA-hrxh-6v49-42gf)
  - `github.com/klauspost/compress` 1.18.6 → 1.18.7 (GHSA-259r-337f-4rfw)

## Test plan

- [x] `go build ./...` passes
- [x] `make test` passes
- [ ] Konflux CI builds successfully

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bsquizz:
Analysis of the test failure. Likely not related to code changes here:

```
  Root cause: The openshift-customer-monitoring namespace was deleted by a previous test before this test's step 00-install.yaml ran.

  The test correctly creates both namespaces in 00-install.yaml (line 9-15 creates openshift-customer-monitoring). But look at the very first reconciliation error (log line 37-38):

  unable to create new content in namespace openshift-customer-monitoring because it is being terminated

  A previous KUTTL test that also used the openshift-customer-monitoring namespace was still tearing it down when this test started. By the time the next retry hit, the namespace
  had fully terminated (line 53: namespaces "openshift-customer-monitoring" not found). The namespace from 00-install.yaml was never recreated because Kubernetes was still
  processing the old deletion.

  The controller then got stuck in an infinite retry loop — every reconciliation attempt failed trying to create the ServiceMonitor in that missing namespace, which prevented it
  from ever creating the Deployment, Secret, or ScaledObject. After the KUTTL 30-second timeout, the assertion on step 1-pods failed because those resources never existed.

  This is a test infrastructure issue, not a code bug and not related to your PR. It's a namespace lifecycle race between KUTTL tests that share the openshift-customer-monitoring
  namespace. The fix would be either to give this test its own unique monitoring namespace name or to add a KUTTL step that waits for the namespace to be fully gone before
  recreating it.
  ```

maknop (approved):

PR #1815, opened 2026-09-14
Update tekton pipelines go-toolset image

JGiola:
Update the tekton pipelines to use the latest available image for go 1.26.7 to keep the tests inline with the used go version in Dockerfile.

Tangentially this port the fix for CVE-2026-33818 also in the pipeline tests runs


ffeghali (approved):
/lgtm

